PR #4, opened 2020-11-25
Update to Cadence v0.10.2, SDK v0.12.2, and Emulator v0.12.2

turbolent:


go.sum
```diff
@@ -640,24 +641,28 @@ github.com/onflow/cadence v0.4.0-beta1/go.mod h1:gaPtSctdMzT5NAoJgzsRuwUkdgRswVH
...
 github.com/onflow/flow-go-sdk v0.12.1 h1:tdY1wxECl15CKSA5jJ6lBtDU/iISQRn91DsEBulYtew=
 github.com/onflow/flow-go-sdk v0.12.1/go.mod h1:7sVH3rpQ+zk2I2xLF/Ob/TCOa3Mz+N2J5DtCDtLBtv4=
+github.com/onflow/flow-go-sdk v0.12.2 h1:ukIAek19pwYoHmSSYeaJ729vCA+04VEv6tCRlxwiWRI=
+github.com/onflow/flow-go-sdk v0.12.2/go.mod h1:jWwkpCkhxzeGmCs7oV6ydx50svez/PYUuhg8Y7dY0Ko=
 github.com/onflow/flow-go/crypto v0.9.4 h1:jSMo91fNCF5Dywt0VEM/AEf/O1fcqRYvubezi9AhlaU=
 github.com/onflow/flow-go/crypto v0.9.4/go.mod h1:oXuvU0Dr4lHKgye6nHEFbBXIWNv+dBQUzoVW5Go38+o=
-github.com/onflow/flow-go/crypto v0.11.0 h1:LsyrIqYFu/THD4tSiCuKq9h/woMofdwvR6r45ZdVYrM=
-github.com/onflow/flow-go/crypto v0.11.0/go.mod h1:oXuvU0Dr4lHKgye6nHEFbBXIWNv+dBQUzoVW5Go38+o=
+github.com/onflow/flow-go/crypto v0.11.1-0.20201124074740-4553dbb0bc38 h1:GIKNc6pq8sBuhY6IxYkXptrs3UQhGvxlrmj/+xEywLA=
+github.com/onflow/flow-go/crypto v0.11.1-0.20201124074740-4553dbb0bc38/go.mod h1:oXuvU0Dr4lHKgye6nHEFbBXIWNv+dBQUzoVW5Go38+o=
```
jordanschalm, line 665: Should we point to the same `crypto` version as `flow-go` version?

jordanschalm (approved):

go.sum
turbolent: Ideally we would us the version of the crypto module that flow-go uses.

However, the version of crypto pinned in flow-go is wrong (see https://github.com/onflow/flow-go/pull/137, brought it up yesterday again: https://axiomzen.slack.com/archives/C015G65HR2P/p1606232145256900). There's a tag for crypto/v0.12.0 missing, which flow-go should use.

This bump is to crypto as of flow-go tag v0.12.0, which is what flow-go implicitly uses when its built due to the replace statements in its go.mod file. This is basically just a workaround for upstream problem.

turbolent:
I also need to update the Cadence Language Server

go.sum
jordanschalm: Right. I mean if we specify the same version/commit here (both `v0.12.0`), it will ensure the build uses both modules at the same version. Unless `flow-go` is somehow pinning a version of `flow-go/crypto` that is _after_ its own version, the version specified here will be chosen.

But I agree the proper solution is removing the `replace` in `flow-go`

Kay-Zee (approved):

go.sum
turbolent: `flow-go/crypto` is it's own module, and in Go, must be tagged separately, as `crypto/v.012.0`, but this tag is missing; the tag `v0.12.0` without the nested module name is not considered. So it's basically using the commit that should have the tag.

PR #5, opened 2020-11-27
Print contracts instead of code

turbolent:
The `Code` field is deprecated and empty now. Show all contracts instead.

### DIFF
--- a/flow/accounts/get/get.go
+++ b/flow/accounts/get/get.go
@@ -83,7 +83,7 @@ func printAccount(account *flow.Account, printCode bool) {
 	fmt.Println("  ---")
 	if printCode {
 		for name, code := range account.Contracts {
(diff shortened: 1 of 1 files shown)


flow/accounts/get/get.go
```diff
@@ -82,8 +82,10 @@ func printAccount(account *flow.Account, printCode bool) {
 	}
 	fmt.Println("  ---")
 	if printCode {
-		fmt.Println("Code:")
-		fmt.Println(string(account.Code))
+		for name, code := range account.Contracts {
+			fmt.Printf("Code '%s':", name)
```
Kay-Zee, line 86:
```suggestion
			fmt.Printf("Code '%s':\n", name)
```

Kay-Zee (approved):

PR #7, opened 2020-12-01
Add Bastian & Janez as code owners

psiemens:
🎸 🍔 

Bastian and Janez have been maintaining the CLI lately as part of the larger Go toolchain (Go SDK, Flow Go, Emulator). This will help them move faster and allow them to request reviews from each other when updating dependencies.


turbolent (approved):
🙏 🚀 🌊

PR #10, opened 2020-12-04
Add back contract deployment

### LINKED ISSUE #2: Add support for deploying contracts during account creation
### Issue To Be Solved

The CLI used to have a flag for deploying code while creating an account, but it was removed in #1 as part of the update to support multiple contracts per account.

Bring back this functionality and allow specifying zero or more contracts that should be deployed when a contract is created.

Kay-Zee:
Closes #2 

## Description

Adds a `--contract` flag to allow deploying a contract along with account creation.

How do we feel about the requirement for the flag contents to be `name:filepath`?

______

For contributor use:

### DIFF
--- a/flow/accounts/create/create.go
+++ b/flow/accounts/create/create.go
@@ -19,7 +19,9 @@
 package create
 
 import (
+	"io/ioutil"
 	"log"
+	"strings"
 
 	"github.com/onflow/flow-go-sdk"
 	"github.com/onflow/flow-go-sdk/crypto"
@@ -37,6 +39,7 @@ type Config struct {
 	HashAlgo string   `default:"SHA3_256" flag:"hash-algo" info:"Hash used for the digest"`
 	Host     string   `flag:"host" info:"Flow Access API host address"`
 	Results  bool     `default:"false" flag:"results" info:"Display the results of the transaction"`
+	Contract string   `flag:"contract,c" info:"Contract to be deployed during account creation. <name:path>"`
 }
 
 var conf Config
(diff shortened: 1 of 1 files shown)


flow/accounts/create/create.go
```diff
@@ -70,7 +73,28 @@ var Cmd = &cobra.Command{
 			}
 		}
 
-		tx := templates.CreateAccount(accountKeys, nil, signerAccount.Address)
+		var contracts []templates.Contract
+
+		if len(conf.Contract) > 0 {
+			contractFlagContent := strings.Split(conf.Contract, ":")
```
turbolent, line 79:
```suggestion
			contractFlagContent := strings.SplitN(conf.Contract, ":", 2)
```

flow/accounts/create/create.go
```diff
@@ -37,6 +39,7 @@ type Config struct {
 	HashAlgo string   `default:"SHA3_256" flag:"hash-algo" info:"Hash used for the digest"`
 	Host     string   `flag:"host" info:"Flow Access API host address"`
 	Results  bool     `default:"false" flag:"results" info:"Display the results of the transaction"`
+	Contract string   `flag:"contract,c" info:"Contract to be deployed during account creation. <name:path>"`
```
turbolent, line 42: Good idea to use `name:path`. Maybe make it a slice like `Keys` above?

turbolent (approved):
Nice! Thanks for bringing this back

flow/accounts/create/create.go
Kay-Zee: I didn't SplitN so that i could use the length as an extra check, but `:` in filename/path would probably just make it error on read later, which is probably okay.

Kay-Zee:
i was also wondering if it makes sense to go with using the command args, similar to `upgrade_contracts`.

flow/accounts/create/create.go
Kay-Zee: that's a pretty easy way to make it support multiple contracts, so sure! added.

psiemens (approved):
Looks good! 

I think it's good to make `--contracts` a flag rather than a positional argument given that it's not required for account creation

PR #19, opened 2020-12-29
Add events

### LINKED ISSUE #18: Request events for block range
### Issue To Be Solved

Currently, it's only possible to get block events for a single block at a time. The access API currently supports getting event info for multiple blocks at once. We should try and leverage this functionality.

### Suggest A Solution

We should either:
a) Add the ability to query multiple blocks at once using the current API. (i.e. put under the `blocks` sub command)
 - The advantage here is a large amount of reusable code
 - The disadvantage is that we'd actually not use the events API provided by the access API in a very optimized way

b) Add a new subcommand `events` which would simply query events.
 - The advantage would be a cleaner separation of intent, and better use of the AccessAPI,
 - The disadvantage would be that we'd now be able to query events from both the `blocks` and `events` sub commands, which could be somewhat confusing, as well as still not having the ability to query multiple blocks at once

I am likely to do a PoC of option b) and go from there.

Kay-Zee:
Closes #18 

## Description

Adds new `events` sub command that 

______

For contributor use:

### DIFF
--- a/flow/events/get/get.go
+++ b/flow/events/get/get.go
@@ -30,7 +30,8 @@ import (
 )
 
 type Config struct {
-	Host string `flag:"host" info:"Flow Access API host address"`
+	Host    string `flag:"host" info:"Flow Access API host address"`
+	Verbose bool   `flag:"verbose" info:"Verbose output"`
 }
 
 var conf Config
@@ -48,7 +49,7 @@ var Cmd = &cobra.Command{
 
 		eventName, startHeight, endHeight := validateArguments(host, args)
 
-		cli.GetBlockEvents(host, startHeight, endHeight, eventName)
+		cli.GetBlockEvents(host, startHeight, endHeight, eventName, conf.Verbose)
 	},
 }
 
--- a/flow/events/get/get.go
+++ b/flow/events/get/get.go
@@ -37,9 +37,10 @@ type Config struct {
 var conf Config
 
 var Cmd = &cobra.Command{
-	Use:   "get <event_name> <block_height_range_start> <optional:block_height_range_end|latest>",
-	Short: "Get events in a block range",
-	Args:  cobra.RangeArgs(2, 3),
+	Use:     "get <event_name> <block_height_range_start> <optional:block_height_range_end|latest>",
+	Short:   "Get events in a block range",
+	Args:    cobra.RangeArgs(2, 3),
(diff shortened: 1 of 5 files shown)


flow/result.go
```diff
@@ -107,16 +107,16 @@ func GetBlockEvents(host string, height uint64, eventType string) {
...
 	})
 
 	if err != nil {
 		Exitf(1, "Failed to query block event by height: %s", err)
 	}
 
-	for _, blockEvent := range events {
-		fmt.Printf("Events for Block %s:", blockEvent.BlockID)
+	for i, blockEvent := range events {
+		fmt.Printf("Events for Block %s (%d):", blockEvent.BlockID, startHeight+uint64(i))
```
janezpodhostnik, line 119: What does the number in parenthesis represent?

janezpodhostnik (commented):
Nice addition. I had one question.

vishalchangrani (approved):

flow/result.go
Kay-Zee: it's the height of the block. better to label?

Kay-Zee:
Example usage:
```
flow events get A.1654653399040a61.FlowToken.TokensDeposited 11559500 11559600
```

joshuahannan:
Could you add a events help indication that shows the correct format of an event definition in case people don't know to use `A.1654653399040a61.FlowToken.TokensDeposited`. Like if they get it wrong, it shows the correct format?

flow/events/get/get.go
```diff
@@ -0,0 +1,98 @@
...
+
+type Config struct {
+	Host    string `flag:"host" info:"Flow Access API host address"`
+	Verbose bool   `flag:"verbose" info:"Verbose output"`
+}
+
+var conf Config
+
+var Cmd = &cobra.Command{
+	Use:   "get <event_name> <block_height_range_start> <optional:block_height_range_end|latest>",
```
turbolent, line 40: Maybe use "type ID" instead of "name", as someone might assume it's possible to just specify e.g `AccountCreated`. 

Also, maybe make it clearer that `latest` here is a literal, not a variable like `block_height_range_start` and `block_height_range_end`

```suggestion
	Use:   "get <event_type_id> <block_height_range_start> <optional:block_height_range_end|'latest'>",
```

turbolent (approved):
Looks good!

turbolent:
@Kay-Zee Maybe add your example as `Example` to the `Command` (https://github.com/spf13/cobra/blob/master/command.go#L64)

PR #21, opened 2020-12-30
Add staking/delegating info for accounts

### LINKED ISSUE #20: Query Account Staking Info
### Issue To Be Solved

Another common query action is to get staking info. While this is mostly available through using `scripts execute`, I'm considering adding the ability to get staking/delegation info for the account as a helper, as that seems to be a very common task

### Suggest A Solution

Adding a `staking-info` sub command to accounts, that will query staking/delegation info for the account.

Kay-Zee:
Closes #20 

## Description

Adds ability to query for Staking / Delegation info

## Outstanding Questions
- This imports flow-core-contracts. That likely increases the size of the binary dramatically, compared to before, due to the large number of scripts that likely get imported along side the two that i'm using. Does it make sense to import that templates package?
- Currently just pretty printing the JSON output of the script. Would like to print it in a nicer format, similar to the event output, but i think that can come in a later PR?

______

For contributor use:


flow/result.go
```diff
@@ -123,15 +123,15 @@ func GetBlockEvents(host string, height uint64, eventType string) {
...
 	// Basic event info printing
 	for _, event := range events {
-		fmt.Printf("  Event %d: %s\n", event.EventIndex, event.String())
+		fmt.Printf(Indent+"Event %d: %s\n", event.EventIndex, event.String())
 		if txID {
-			fmt.Printf("  Tx ID: %s\n", event.TransactionID)
+			fmt.Printf(Indent+"Tx ID: %s\n", event.TransactionID)
 		}
-		fmt.Println("    Fields:")
+		fmt.Println(Indent + Indent + "Fields:")
```
janezpodhostnik, line 134: There Is inconsistent formatting in this file.

janezpodhostnik (approved):
LGTM. One file has some indenting problem I think.

flow/result.go
Kay-Zee: I'm not sure what inconsistencies you mean. If you're talking about the spacing between the `+` and the variables/strings, that's just what `gofmt` forces.

I can rejig the printing logic abit to get around this, if it looks too odd to you

psiemens (approved):
LGTM.

I think it's fine for the CLI to import `flow-core-contracts`. 👍

PR #42, opened 2021-02-01
Feature: project commands 

### LINKED ISSUE #30: Automated contract deployment
### Issue To Be Solved

Currently when using the emulator or testnet, a developer needs to manually redeploy their contracts to apply updates. This is often a tedious process, especially if multiple contracts need to be deployed, or if there is more than one account.

Can the CLI help to automate this process? Ideally a developer would be able to run a simple command to bring the emulator (or testnet) in sync with their local files.

### Suggest A Solution

This isn't a concrete solution, but I think we should look into expanding `flow.json` to include an entry for contracts. I'm not sure what the format would be, but it would allow developers to declare a mapping between accounts and contracts on their local filesystem. The CLI could then detect and redeploy any changes to these files, either automatically or at the user's request.

psiemens:
Closes #30 

## Description

This PR adds a new `flow project` command to hold commands for managing a Flow project.

The `flow project` commands use a revised `flow.json` format inspired by the [discussion in this issue](#30).

______

For contributor use:


sideninja:
If possible in the future maybe we can avoid beta command by detecting the config file format and switching to the correct parser. In case the format is old we don't have to even show the commands available with the new format.

janezpodhostnik (approved):

PR #45, opened 2021-02-01
Add deploy-contracts command

### LINKED ISSUE #30: Automated contract deployment

psiemens:
Closes #30 

## Description

This PR introduces the `flow beta deploy-contracts` command, which automates the contract deployment process for Flow developers.

This command reads contract source files defined in `flow.json` and maps them to their declared target accounts. The following features are supported:

- A contract can be mapped to a different account on each network (emulator, testnet, mainnet)
- A contract can import other contracts by filename. Filenames are replaced with their target addresses (declared in `flow.json`) before deployment
- A contract can import a contract by alias. Aliases are replaced before deployment using the `"aliases"` map in `flow.json`.
- The deployment command resolves dependencies between contracts and deploys them in deterministic topological ordering so that a contract is only deployed after its dependencies. Import cycles result in an error.

cli-contract-deployment-6

Below is the `flow.json` config used in the above screen recording:

```json
{
  "emulator": {
    "default": {
      "port": 3569,
      "serviceKey": {
        "privateKey": "40055011eda87a6c7b38943d1cc81c013eeb5bf791ad5330a375ab229a401e95",
        "signatureAlgorithm": "ECDSA_P256",
        "hashAlgorithm": "SHA3_256"
      }
    }
  },
  "networks": {
    "emulator": {
      "host": "127.0.0.1:3569",
      "chain": "flow-emulator"
    }
  },
  "aliases": {
    "FungibleToken": {
      "emulator": "0xee82856bf20e2aa6"
    }
  },
  "contracts": {
    "nft": {
      "target": {
        "emulator": "emulator-service-account"
      },
      "source": {
        "NonFungibleToken": "./cadence/kittyItems/contracts/NonFungibleToken.cdc"
      }
    },
    "kitty-items": {
      "target": {
        "emulator": "emulator-service-account"
      },
      "source": {
        "Kibble": "./cadence/kibble/contracts/Kibble.cdc",
        "KittyItems": "./cadence/kittyItems/contracts/KittyItems.cdc",
        "KittyItemsMarket": "./cadence/kittyItemsMarket/contracts/KittyItemsMarket.cdc"
      }
    }
  },
  "accounts": {
    "emulator-service-account": {
      "chain": "flow-emulator",
      "address": "service",
      "keys": [
        {
          "signatureAlgorithm": "ECDSA_P256",
          "hashAlgorithm": "SHA3_256",
          "type": "hex",
          "index": 0,
          "context": {
            "privateKey": "40055011eda87a6c7b38943d1cc81c013eeb5bf791ad5330a375ab229a401e95"
          }
        }
      ]
    }
  }
}
```

______

For contributor use:


sideninja:
**Documentation**
I think it would be great to have better documentation on the command and more verbose error handling in cases contracts are not deployed. One such case is when the flow emulator returns an error.

**Sync Configuration**
Another issue might be that developers will init configuration, run the emulator, change the configuration file, and then since the emulator will not be restarted there will be a mismatch between emulator and deployer.

**Security**
One concern I have is security with flow.json file since the nature of this file is to be included in svn the same way as package.json is. There is one important difference between those two, flow.json includes secret keys and I can imagine devs will use this for mainnet deployment and they will forget to remove the secret from that configuration. Something to keep in mind. 

**Remote Imports**
It could also be cool to have an option to include contracts from URLs and github. Maybe even built into tool some most standard contracts such as NFT so you don't even have to provide URLs but the tool knows where the contracts are (this second step might be useful once the community establish standard contracts the same way as for an example erc on ethereum)

**Configuration Flow**
Lastly I think besides having improved documentation for CLI in general and specifically for this subcommand, there could also be a subcommand that would allow adding contracts to the configuration so developers wouldn't have to write config manually. Pros of this in my opinion are avoiding typos, commands could offer natural flow with documentation so even if a developer won't read the documentation (I prefer using tools that don't come with a manual you must read), they can figure out how it works by just trying out (which is almost impossible in case of manual configuration), another benefit of this subcommand is potentially having an ability to just add contracts in a folder and not specifying one by one (some kind of discovery).

sideninja:
Feedback on the flow.json schema, I agree flow.private.json is a good idea to maybe avoid pushing secrets to github, it would also be helpful to have a warning in CLI when ran to warn you to be careful with the secret when running on mainnet? Or should CLI even discourage using that for mainnet? 

I think schema can be further simplify by not specifying default values. Such an example is:
```
"signatureAlgorithm": "ECDSA_P256",
"hashAlgorithm": "SHA3_256",
"type": "hex",
"index": 0,
```
These keys are default and I think should be avoided in config but if a developer wants to change them they can.
My thinking is to have as little as possible in config so it is easier to read and understand. 

Another thing that might be confusing a bit is that under the contracts key the first key are not contracts themselves but accounts. I'm not even sure what that first key represents and how can it be used (but that still might be because of my lack of understanding the code):
```
"contracts": {
    "nft": {
      "target": {
        "emulator": "emulator-service-account"
      },
      "source": {
        "NonFungibleToken": "./cadence/kittyItems/contracts/NonFungibleToken.cdc"
      }
    },
```
Is `nft` key meant to reference anything? If not it might be a good idea to just make an array with objects.

Are `emulator`, `testnet` and `mainet` meant to be magical values with special meaning or just something we should define in config? In one way it could be good to have this predefined values but also allow for user to define more, since I think there can also be other networks  (`testing`, `qa`... which could be different environments devs will potentially have setup locally or something).

sideninja:
Test project:

**Kitty-items**

Deployed to emulator on two accounts,

Deployed to testnet on two accounts

I managed to lower the complexity of config (line count) by **58%** for same configuration - **49 lines vs 114 lines** 

schema current:

```json
{
	"emulator": {
		"default": {
			"port": 3569,
			"serviceKey": {
				"privateKey": "dd72967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47",
				"signatureAlgorithm": "ECDSA_P256",
				"hashAlgorithm": "SHA3_256"
			}
		}
	},

	"networks": {
		"emulator": {
			"host": "127.0.0.1:3569",
			"chain": "flow-emulator"
		},
    "testnet": {
      "host": "access.testnet.nodes.onflow.org:9000",
      "chain": "testnet"
    }
	},

	"aliases": null,

	"contracts": {
		"test": {
			"target": {
				"emulator": "account-3",
				"testnet": "account-1"
			},
			"source": {
				"KittyItems": "./cadence/kittyItems/contracts/KittyItems.cdc",
        "KittyItemsMarket": "./cadence/kittyItemsMarket/contracts/KittyItemsMarket.cdc"
			}
		},
		"hungry-kitties": {
      "target": {
				"emulator": "account-2",
				"testnet": "account-4"
      },
      "source": {
				"NonFungibleToken": "../hungry-kitties/cadence/contracts/NonFungibleToken.cdc",
				"FungibleToken": "../hungry-kitties/cadence/contracts/FungibleToken.cdc",
        "Kibble": "./cadence/kibble/contracts/Kibble.cdc",
        "KittyItems": "./cadence/kittyItems/contracts/KittyItems.cdc",
        "KittyItemsMarket": "./cadence/kittyItemsMarket/contracts/KittyItemsMarket.cdc"
      }
    }
	},
	
	"accounts": {
		"account-1": {
			"address": "0xabd6e0586b0af502",
			"chain": "testnet",
			"keys": [
				{
					"type": "hex",
					"index": 0,
					"signatureAlgorithm": "ECDSA_P256",
					"hashAlgorithm": "SHA3_256",
					"context": {
						"privateKey": "1232967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"
					}
				}
			]
		},
		"account-2": {
			"address": "0x2c1162386b0a245f",
			"chain": "testnet",
			"keys": [
				{
					"type": "hex",
					"index": 0,
					"signatureAlgorithm": "ECDSA_P256",
					"hashAlgorithm": "SHA3_256",
					"context": {
						"privateKey": "222967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"
					}
				}
			]
		},		
		"account-3": {
			"address": "service",
			"chain": "flow-emulator",
			"keys": [
				{
					"type": "hex",
					"index": 0,
					"signatureAlgorithm": "ECDSA_P256",
					"hashAlgorithm": "SHA3_256",
					"context": {
						"privateKey": "dd72967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"
					}
				}
			]
		},
		"account-4": {
			"address": "0xf8d6e0586b0a20c7",
			"chain": "flow-emulator",
			"keys": [
				{
					"type": "hex",
					"index": 0,
					"signatureAlgorithm": "ECDSA_P256",
					"hashAlgorithm": "SHA3_256",
					"context": {
						"privateKey": "30dd224db9c7a70f035f8092f6c5a44c920bb6cc4c86c4d7b1b34cb32d2f936f"
					}
				}
			]
		}
	}
}
```

shorter version:

- remove default values:
    - account remove all the keys default values, make it either string for a single key with all default values or array for multiple keys as in old format but since most cases will probably be single key with default algos maybe this is better
- I think contracts section can define key value for each contract and from that point on config just reference them by name - so we don't have duplication of paths as visible in above scenario and also allows us to define array of contracts for each account (shorter more visible)
- remove `chain` value since everything can be referenced by network name although I might not understand that correctly
- emulator under network? since we define networks can that be a place for emulator config to also be defined since emulator is treated as a network in this case? Also multiple networks could have their own config in there - example gcloud-testnet could include also keys for api etc

```json
{
  "contracts": {
    "NonFungibleToken": "../hungry-kitties/cadence/contracts/NonFungibleToken.cdc",
    "FungibleToken": "../hungry-kitties/cadence/contracts/FungibleToken.cdc",
    "Kibble": "./cadence/kibble/contracts/Kibble.cdc",
    "KittyItems": "./cadence/kittyItems/contracts/KittyItems.cdc",
    "KittyItemsMarket": "./cadence/kittyItemsMarket/contracts/KittyItemsMarket.cdc"
  },

  "deploy": {
    "account-1": ["KittyItems", "KittyItemsMarket"],
    "account-2": ["NonFungibleToken", "FungibleToken", "Kibble", "KittyItems", "KittyItemsMarket"],
    "account-3": ["KittyItems", "KittyItemsMarket"],
    "account-4": ["NonFungibleToken", "FungibleToken", "Kibble", "KittyItems", "KittyItemsMarket"]
  },

  "accounts": {
    "account-1": {
      "address": "0xabd6e0586b0af502",
      "keys": "1232967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47",
      "network": "testnet"
    },
    "account-2": {
      "address": "0x2c1162386b0a245f",
      "keys": "22232967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47",
      "network": "testnet"
    },
    "account-3": {
      "address": "service",
      "keys": "service",
      "network": "emulator"
    },
    "account-4": {
      "address": "0xf8d6e0586b0a20c7",
      "keys": "4442967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47",
      "network": "emulator"
    }
  },

  "networks": {
		"emulator": {
			"host": "127.0.0.1:3569",
			"keys": "dd72967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"
		},
    "testnet": "access.testnet.nodes.onflow.org:9000"
  },

	"aliases": null
}
```

psiemens:
Some sketches from our call: 

```
flow project deploy account-2 account-3 account-4

flow project deploy --network emulator

flow project deploy nft kitty-items

flow project deploy kitty-items
```

```json
"contracts": {
  "NonFungibleToken": "../hungry-kitties/cadence/contracts/NonFungibleToken.cdc",
  "Kibble": "./cadence/kibble/contracts/Kibble.cdc",
  "KittyItems": "./cadence/kittyItems/contracts/KittyItems.cdc",
  "KittyItemsMarket": "./cadence/kittyItemsMarket/contracts/KittyItemsMarket.cdc"
},
"aliases: {
  "emulator": {
      "FungibleToken": "0xabc",
  },
  "testnet": {
    "FungibleToken": "0xabc",
    "NonFungibleToken": "0x1234",
  }
}
"deploy": {
    "testnet": {
      "account-1": ["KittyItems", "KittyItemsMarket"],
      "account-2": ["Kibble", "KittyItems", "KittyItemsMarket"],
    }, 
    "emulator": {
      "account-3": ["KittyItems", "KittyItemsMarket"],
      "account-4": ["NonFungibleToken", "Kibble", "KittyItems", "KittyItemsMarket"]
    }
  },
```

```json
  "contracts": {
    "nft": {
      "target": {
        "emulator": "emulator-service-account",
        "testnet": "0xabc"
      },
      "source": {
        "NonFungibleToken": "./cadence/kittyItems/contracts/NonFungibleToken.cdc"
      }
    },
    "kitty-items": {
      "target": {
        "emulator": "emulator-service-account",
         "testnet": "testnet-admin-account"
      },
      "source": {
        "Kibble": "./cadence/kibble/contracts/Kibble.cdc",
        "KittyItems": "./cadence/kittyItems/contracts/KittyItems.cdc",
        "KittyItemsMarket": "./cadence/kittyItemsMarket/contracts/KittyItemsMarket.cdc",
      }
    }
  }
```

bjartek:
it would be nice to have some cli features to help generate/modify that flow.json file. 

- when I add an account it can be added to the file
- if i have a folder of contracts  i want to be added to an account i can do that

10thfloor:
@sideninja 

> My thinking is to have as little as possible in config 
👍🏼

turbolent:
This looks great! 👏 🎉

10thfloor:
This is amazing. This config https://github.com/onflow/flow-cli/pull/45#issuecomment-772654030 looks really promising.  What would be the command for doing a 'reset' of the deployment? 
Something like 
```
flow project reset 
```
Followed by a confirmation step: 
```
Are you sure you want to reset project-kittyItems? (N/y)
```

10thfloor:
Some contracts could be deployed to multiple accounts no?
We could allow for: 
```
...
 "contracts": {
    "Profile": {
      "target": {
        "emulator": "emulator-service-account",
        "testnet": ["0xabc", "0xdef"] // Array of accounts
      },
      "source": {
        "NonFungibleToken": "./cadence/kittyItems/contracts/NonFungibleToken.cdc"
      }
    },
  ...

**Comment by 10thfloor** [comment]
☝🏼 But again, this seems like "state" not "configuration", which means we _could eliminate this from the config file_. Because when we're talking about "which-accounts-go-where" that's not config is it?

**Comment by joshuahannan** [comment]
Yeah this all looks really great! I'm still wondering what the plan would be for importing contracts that are in other repos. Doing it by URL would probably work, but isn't ideal, but I definitely want to avoid having to copy and paste the contracts into your repo.

**Comment by sideninja** [comment]
> it would be nice to have some cli features to help generate/modify that flow.json file.
> 
> * when I add an account it can be added to the file
> * if i have a folder of contracts  i want to be added to an account i can do that

I agree 100% and want to build that when it will be possible.



> This is amazing. This config [#45 (comment)](https://github.com/onflow/flow-cli/pull/45#issuecomment-772654030) looks really promising.
> 
> Quick thought relevent to this thread: What would be the command for doing a 'reset' of a deployment?
> 
> ```
> flow project reset 
> ```
> 
> Followed by a confirmation step?
> 
> ```
> Are you sure you want to reset "project-kittyItems" ? (N/y)
> ```

I like your proposal. There will probably be more subcommands around the project command which will allow you to manage the project.





> Yeah this all looks really great! I'm still wondering what the plan would be for importing contracts that are in other repos. Doing it by URL would probably work but isn't ideal, but I definitely want to avoid having to copy and paste the contracts into your repo.

We had a quick discussion on this topic and we envisioned the possibility in this configuration to have different source formats. One is for sure URL but as you mentioned it might not be the best way, so I think we will have a separate issue on that very soon and implement a solution from the feedback.

File: flow/beta/cli/project.go [hunk 1]
```diff
@@ -29,59 +29,77 @@ import (
...
+	return newProject(conf)
 }
 
+//REF: discuss and rethink what project init is and where configuration init should be stored
+// InitProject create project structure
 func InitProject() *Project {
-	serviceAccount, serviceAccountKey := generateEmulatorServiceAccount()
+	serviceAccount := generateEmulatorServiceAccount()
 
+	config := config.Config{
```
psiemens, line 73: > discuss and rethink what project init is and where configuration init should be stored

I wasn't sure about the best way to structure this. I ended up defining `defaultConfig` in `project.go` because I was trying to keep the `config` package "stateless" by making it only describe the config _structure_, not the default values.

In general, I tried to decouple "project" and "config" so that it'd be easier to support multiple config formats without having to change the actual project model. For example, there could be multiple backwards-compatible versions of the config schema, or even a YAML version.

However, there ended up being a lot of overlap. I'm fine with moving all of the config-related logic (including default values) to the `config` package. In fact, what if we just combined `project.go` and `config` into a single package? We can try to create a config abstraction later if it's needed.

flow/beta/cli/project.go
```diff
@@ -113,145 +132,52 @@ func generateEmulatorServiceAccount() (*Account, *keys.HexAccountKey) {
...
 	}
 
-	return flow.HexToAddress(target[2:])
-}
-
-type Account struct {
-	name    string
-	address flow.Address
-	chainID flow.ChainID
-	keys    []keys.AccountKey
```
psiemens, line 179: Have you given thought to how `keys.AccountKey` would fit into the new structure? Is it still useful to have keys defined in their own package?

I created the `keys` package to avoid polluting the config package with business logic, and also to isolate some of the more sensitive crypto code.

psiemens (commented):
@sideninja Looking good so far! I'm going to take a closer look at the config encoding/decoding logic once you have all the pieces working.

flow/beta/cli/project.go
sideninja: I agree. In general, I think default values should be part of the config but shouldn't be tied to JSON config. I strongly agree with the possibility of having different formats. I think config with value validation and default values etc. should be a layer above that. I would at that point create different config parsers (some would support different formats and others would support older versions) but then all the parsers would return a structure defined in config. Right now since we only support one format all of that is together but I would at a later point just move parsing of JSON to another file but keep the structures I've created in the config file since all further operations are based on those structures and hence provide an abstraction over the concrete format of the config file.

flow/beta/cli/project.go
sideninja: Will look into this

flow/beta/cli/project.go
psiemens: > I think config with value validation and default values etc. should be a layer above that.

+1 for separating the config structure from the actual encoding logic. However, I think you're right that we should just focus on the JSON structure for now and split it out later if needed.

>  I want to expose config at a later point as a single point of config for the whole cli and hence should be generic

I agree 👍

bjartek:
Not sure if this is appropriate for this pr, but yesterday i deployed the versus contracts to testnet. In order to do that I created a key, then added an account for that key to get an address for it. I had to manually add the key and address to flow.json.

It would be nice it we can think of this posibility when the new config file is created.

sideninja:
> Not sure if this is appropriate for this pr, but yesterday i deployed the versus contracts to testnet. In order to do that I created a key, then added an account for that key to get an address for it. I had to manually add the key and address to flow.json.
> 
> It would be nice it we can think of this posibility when the new config file is created.

I feel your pain and I agree there should be a command for that.

sideninja:
After discussing with @psiemens we think in the light of new schema it might be more appropriate to also rethink where we should define accounts for the emulator. We believe service account should be defined inline under `accounts`object with other accounts to provide more consistency. Example schema:

```
...
"accounts": {
  "account-2": {
    "address": "0x2c1162386b0a245f",
    "keys": "fbb8ad74b47"
  },
  "emulator-service": {
      "address": "service",
      "keys": "123123123"
  }
},
"networks": {
  "emulator": {
    "host": "127.0.0.1:3569",
    "serviceAccount": "emulator-service",
  },
}
...
```

This way new developers looking at config won't be confused in what this special service account is but it will seem natural. Network object can then also keep config for emulator and reference the "emulator-service" account so there is no black magic going on when we start the emulator. This way we will know emulator is started with mentioned account.

flow/beta/cli/config/config.go
```diff
@@ -83,33 +160,177 @@ func (k EmulatorServiceKey) MarshalJSON() ([]byte, error) {
...
+		return err
+	}
+
+	// go over each network
+	for network, v := range raw {
+		// for each network go through all accounts
+		for account, c := range v {
+			deploy := new(Deploy)
+			deploy.Network = network
+			deploy.Account = account
```
psiemens, line 179: It looks like you implemented these custom `UnmarshalJSON` functions to be able to assign the object keys (`network`, `account`) as values on the resulting `Deploy` struct. Is that correct?

I like the new `Config` struct you defined. The flatter structure is much easier to work with. However, I worry that the encoding/decoding logic is now a bit confusing -- it kind of parses and transforms the data at the same time.

I know you're working on refactoring the config logic already, so feel free to ignore this if you have a different approach, but I wanted to suggest that we define the JSON config structure with a separate `jsonConfig` type 👇 

(BTW, with this approach you don't need to define the `*Collection` structs.)

```go
type Config struct {
  Contracts Contracts
  // ...
}

type Contracts []Contract

type jsonConfig struct {
  Contracts jsonContracts `json:"contracts"`
  // ...
}

func (j jsonConfig) transformToConfig() Config {
  return Config{
    Contracts: j.Contracts.transformToConfig(),
    // ...
  }
}

type jsonContracts map[string]jsonContract

func (j jsonContracts) transformToContracts() Contracts {
  contracts := make(Contracts, 0)
  
  for contractName, c := range j {
    if c.Source != "" {
      contract := Contract{
        Name: contractName,
        Source: source,
      }
      
      contracts = append(contracts, contract) 
    }
    
    for networkName, source := range c.SourcesByNetwork {
      contract := Contract{
        Name: contractName,
        Source: source,
        Network: networkName,
      }
      
      contracts = append(contracts, contract) 
    }
    
    return contracts
}

type jsonContract struct {
    Source           string
    SourcesByNetwork map[string]string
}

func (j *jsonContract) UnmarshalJSON(b []byte) error {
  // ref: https://irshadhasmat.medium.com/golang-simple-json-parsing-using-empty-interface-and-without-struct-in-go-language-e56d0e69968
  var val interface{}
  
  err := json.Unmarshal(b, &val)
  if err != nil {
    return err
  }
  
  switch typedVal := val.(type) {
  case string:
    // simple: single source for all networks
    j.Source = typedVal
  case map[string]string:
    // advanced: different source for each network
    j.SourcesByNetwork = typedVal
  default:
    return errors.New("invalid contract definition")
  }

  return nil
}
```

flow/beta/cli/config/config.go
```diff
@@ -83,33 +160,177 @@ func (k EmulatorServiceKey) MarshalJSON() ([]byte, error) {
...
+
+			// try to parse contracts as array of strings
+			contracts := []string{}
+			err := json.Unmarshal(c, &contracts)
+			if err == nil { // simple format
+				deploy.Contracts = contracts
+			} else { // advanced fromat
+				//TODO: implement format with contract init values
+			}
```
psiemens, line 188: This currently misses the case when `err != nil`

flow/beta/cli/config/config.go
```diff
@@ -83,33 +160,177 @@ func (k EmulatorServiceKey) MarshalJSON() ([]byte, error) {
...
+			"signatureAlgorithm": "ECDSA_P256",
+			"hashAlgorithm": "SHA3_256",
+			"context": {
+				"privateKey": "`+keysString+`"
+			}
+		}]`), &keys)
+		a.Keys = keys
+	} else {
+		return err
+	}
```
psiemens, line 327: This case is redundant since `err == nil`, can just return `nil` below.

flow/beta/cli/config/config.go
```diff
@@ -83,33 +160,177 @@ func (k EmulatorServiceKey) MarshalJSON() ([]byte, error) {
...
+	var address string
+	json.Unmarshal(raw["address"], &address)
+
+	// TODO: address validation format
+	if address == "service" {
+		if a.ChainID == "" {
+			a.ChainID = flow.Emulator //TODO: find better way for defaults in general
+		}
+
+		a.Address = flow.ServiceAddress(a.ChainID)
```
psiemens, line 300: I think it would make sense to add this special logic to the "transform" step in order simplify this `UnmarshalJSON` function.

flow/beta/cli/config/config.go
```diff
@@ -83,33 +160,177 @@ func (k EmulatorServiceKey) MarshalJSON() ([]byte, error) {
...
+		json.Unmarshal([]byte(`[{
+			"type": "hex",
+			"index": 0,
+			"signatureAlgorithm": "ECDSA_P256",
+			"hashAlgorithm": "SHA3_256",
+			"context": {
+				"privateKey": "`+keysString+`"
+			}
+		}]`), &keys)
+		a.Keys = keys
```
psiemens, line 324: You can just define a Go slice literal for the default case:

```go
keys := []AccountKey{
 AccountKey{
    Type: KeyTypeHex,
    Index: 0,
    SigAlgo: crypto.ECDSA_P256,
    HashAlgo: crypto.SHA3_256,
    Context: map[string]string{
      "privateKey": keyString,
    },
  },
}

File: flow/beta/cli/config/config.go [hunk 8]
```diff
@@ -138,6 +360,7 @@ func (a AccountKey) MarshalJSON() ([]byte, error) {
 	})
 }
 
+// Save configuration to a path file in json format
```
psiemens, line 363: The convention in Go is to write function doc comments as a complete sentence where the first word always refers to the funtion (even when it's a bit awkward). These comments are parsed by [godoc](https://blog.golang.org/godoc)  and used to autogenerate documentation, so it's best to follow the expected format.

```suggestion
// Save saves the configuration to the specified path file in JSON format.
```

psiemens: BTW, it's also conventional in Go doc comments to always capitalize acronyms (e.g. JSON instead of json)

psiemens (commented):
For this review I focused on the JSON encoding/decoding logic.

sideninja:
Does it make sense to update `AccountKey` structure for advance schema from:
```
type AccountKey struct {
	Type     KeyType
	Index    int
	SigAlgo  crypto.SignatureAlgorithm
	HashAlgo crypto.HashAlgorithm
	Context  map[string]string
}
```
to this structure:
```
type AccountKey struct {
	Key      string
	Type     KeyType
	Index    int
	SigAlgo  crypto.SignatureAlgorithm
	HashAlgo crypto.HashAlgorithm
}
```

What is the reason that key is saved in the `Context` and has to be accessed by: `Account.AccountKey.Context["privateKey"]`? Wouldn't it be better to have it: `Account.AccountKey.Key`? If context should be dynamic data giving more context to the AccountKey it can stay but since the private key is crucial I would define it separately as a property. Maybe I'm missing something

flow/beta/cli/config/json/account.go
```diff
@@ -0,0 +1,200 @@
...
+func (j jsonAccounts) transformToJSON(accounts config.Accounts) jsonAccounts {
+	jsonAccounts := jsonAccounts{}
+
+	for _, a := range accounts {
+		// if simple
+		//REF: horible implementation improvemenet needed
+		if len(a.Keys) == 1 && a.Keys[0].Index == 0 &&
+			a.Keys[0].Type == config.KeyTypeHex &&
+			a.Keys[0].SigAlgo == crypto.ECDSA_P256 &&
+			a.Keys[0].HashAlgo == crypto.SHA3_256 {
```
psiemens, line 114: Yeah, I was thinking about how we'd reverse this case. Given just a key struct, there's technically no way of knowing if it should be reduced to the simple case.

I think this is the right approach, though. Maybe just move this if clause to a separate helper function? e.g. `isDefaultKeyFormat(a.Keys)`

flow/beta/cli/config/json/account.go
```diff
@@ -0,0 +1,200 @@
...
+	if err == nil {
+		j.Advanced = advanced
+		return nil
+	}
+
+	return err
+}
+
+func (j jsonAccount) MarshalJSON() ([]byte, error) {
+	if j.Simple.Address != "" {
```
psiemens, line 195:
```suggestion
	if j.Simple != jsonAccountSimple{} {
```

You can compare directly to the empty (zeroed) struct. I find this a bit cleaner.

flow/beta/cli/config/json/account.go
```diff
@@ -0,0 +1,200 @@
...
+		return json.Marshal(j.Simple)
+	} else {
+		return json.Marshal(j.Advanced)
+	}
```
psiemens, line 199:
```suggestion
	if j.Simple.Address != "" {
		return json.Marshal(j.Simple)
	}
	
	return json.Marshal(j.Advanced)
```

It's a bit more idiomatic in Go to just collapse the `else` case in situations like this.

flow/beta/cli/config/json/account.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// try advanced format
+	var advanced jsonAccountAdvanced
+	err = json.Unmarshal(b, &advanced)
+	if err == nil {
+		j.Advanced = advanced
+		return nil
+	}
+
+	return err
```
psiemens, line 191: Maybe add a TODO to improve the error message here? If the key is neither a valid simple nor advanced key, the advanced error will always be returned, which could be a bit misleading.

I don't think it's worth solving now, but probably worth a TODO.

flow/beta/cli/config/json/account.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		accounts = append(accounts, account)
+	}
+
+	return accounts
+}
+
+// transformToJSON transforms config structure to json structures for saving
+func (j jsonAccounts) transformToJSON(accounts config.Accounts) jsonAccounts {
```
psiemens, line 105: I don't think this function needs to be attached to `jsonAccounts`. I know it's a longer name, but I would just define it like this:

```go
func transformAccountsToJSON(accounts config.Accounts) jsonAccounts { ... }

File: flow/beta/cli/config/json/account.go [hunk 14]
```diff
@@ -0,0 +1,200 @@
...
+	jsonAccounts := jsonAccounts{}
+
+	for _, a := range accounts {
```
psiemens, line 108: Might be worthwhile to split this loop block into a couple helpers. 

```go
if isDefaultKeyFormat(a.Keys) {
  jsonAccounts[a.Name] = transformSimpleAccountToJSON(a)
} else {
  jsonAccounts[a.Name] = transformAdvancedAccountToJSON(a)
}

File: flow/beta/cli/config/json/account_test.go [hunk 15]
```diff
@@ -0,0 +1,322 @@
...
+	var jsonAccounts jsonAccounts
+	json.Unmarshal(b, &jsonAccounts)
+
+	accounts := jsonAccounts.transformToConfig()
+
+	j := jsonAccounts.transformToJSON(accounts)
+	x, _ := json.Marshal(j)
+
+	// our output format is shorted - improve test
+	assert.NotEqual(t, string(b), string(x))
```
psiemens, line 307: You can use `assert.JSONEq` here -- it will normalize the JSON strings and compare for equality.

flow/beta/cli/config/json/account_test.go
```diff
@@ -0,0 +1,322 @@
+/*
```
psiemens, line 1: Awesome tests in this file 👏  I really like how you isolated the cases

flow/beta/cli/config/json/config.go
```diff
@@ -0,0 +1,105 @@
...
+func (j jsonConfig) transformToConfig() *config.Config {
+	return &config.Config{
+		Contracts: j.Contracts.transformToConfig(),
+		Networks:  j.Networks.transformToConfig(),
+		Accounts:  j.Accounts.transformToConfig(),
+		Deploys:   j.Deploys.transformToConfig(),
+	}
+}
+
+func (j jsonConfig) transformToJSON(config *config.Config) jsonConfig {
```
psiemens, line 47: Same here, I'd make this a pure function:

```suggestion
func transformConfigToJSON(config *config.Config) jsonConfig {
```

flow/beta/cli/config/json/config.go
```diff
@@ -0,0 +1,105 @@
...
+		Contracts: j.Contracts.transformToJSON(config.Contracts),
+		Networks:  j.Networks.transformToJSON(config.Networks),
+		Accounts:  j.Accounts.transformToJSON(config.Accounts),
+		Deploys:   j.Deploys.transformToJSON(config.Deploys),
+	}
+}
+
+// Save saves the configuration to the specified path file in JSON format.
+func Save(conf *config.Config, path string) error {
+	jsonConfig := jsonConfig{}.transformToJSON(conf)
```
psiemens, line 58:
```suggestion
	jsonConf := jsonConfig{}.transformToJSON(conf)
```

Best to avoid overwriting the existing `jsonConfig` definition

flow/beta/cli/config/json/contract.go
```diff
@@ -0,0 +1,118 @@
...
+			contracts = append(contracts, contract)
+		}
+	}
+
+	return contracts
+}
+
+//REF: if we already loaded json from config no need to do this just return
+// transformToJSON transforms config structure to json structures for saving
+func (j jsonContracts) transformToJSON(contracts config.Contracts) jsonContracts {
```
psiemens, line 60:
```suggestion
func transformContractsToJSON(contracts config.Contracts) jsonContracts {
```

flow/beta/cli/config/json/deploy.go
```diff
@@ -0,0 +1,88 @@
...
+
+			deploys = append(deploys, deploy)
+		}
+	}
+
+	return deploys
+}
+
+// transformToJSON transforms config structure to json structures for saving
+func (j jsonDeploys) transformToJSON(deploys config.Deploys) jsonDeploys {
```
psiemens, line 49:
```suggestion
func transformDeploysToJSON(deploys config.Deploys) jsonDeploys {
```

flow/beta/cli/config/json/deploy.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return jsonDeploys
+}
+
+type Simple map[string][]string
+
+type jsonDeploy struct {
+	Simple
+	//TODO: advanced format will include variables
```
psiemens, line 71:
```suggestion
	// TODO: advanced format will include initializer arguments
```

flow/beta/cli/config/json/network.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		networks = append(networks, network)
+	}
+
+	return networks
+}
+
+// transformToJSON transforms config structure to json structures for saving
+func (j jsonNetworks) transformToJSON(networks config.Networks) jsonNetworks {
```
psiemens, line 57:
```suggestion
func transformNetworksToJSON(networks config.Networks) jsonNetworks {
```

psiemens (commented):
Great refactor! 

My comments are minor and mostly cosmetic. I'd say this is pretty much ready to go 💯

sideninja:
> could be deployed to multiple accounts no

The way you can deploy a contract to multiple accounts with the new configuration is simply:
```
{
...
"deploys": {
		"emulator": {
			"emulator-account": ["FungibleToken", "NonFungibleToken"],
			"admin": ["FungibleToken"]
		}
}
...
```

But this feature is currently disabled as it introduces some problems - one of most important is how to know which deployment of contract gets then imported in all the `imports` in the contract code files. I'm sure it can be solved down the line but until it's proved to be a really useful funcionality we maybe can give priority to some other things.

psiemens:
Closing this since most of the commits are from @sideninja now. Can you open another PR from the same branch?

PR #51, opened 2021-02-04
Pass arguments to scripts and transactions

### LINKED ISSUE #35: Script arguments
### Issue To Be Solved

The CLI doesn't allow the user to pass arguments into a script execution. This prevents the CLI from working with scripts that declare any parameters.

### Suggest A Solution

Update the `flow scripts execute` command like this:

```
flow scripts execute myScript.cdc --args "foo" 123.2
```

### LINKED ISSUE #36: Transaction arguments
### Issue To Be Solved

The CLI doesn't allow the user to pass arguments into a transaction. This prevents the CLI from working with transactions that declare any parameters.

### Suggest A Solution

Update the `flow transactions send` command like this:

```
flow transactions send --code myTx.cdc --args "foo" 123.2
```

MaxStalker:
## Description
This PR will bring the ability to pass arguments to scripts and transactions by specifying arguments as a string in [json-cadence]( https://docs.onflow.org/cadence/json-cadence-spec#gatsby-focus-wrapper) format. Below you can find an example, which will pass single argument of type `String` with value `Hello, Cadence`
```
flow scripts execute script.cdc --args="[{\"type\": \"String\", \"value\": \"Hello, Cadence\"}]"
```
```
flow transactions send --code=tx.cdc --args="[{\"type\": \"String\", \"value\": \"Hello, Cadence\"}]"
```
While the format of the passing arguments is different from the one mentioned in the respective issues, json-cadence format will allow to better express what type of the argument we want to pass, without guessing on CLI side.

Closes #35 
Closes #36


janezpodhostnik (approved):
Nice!

Two thoughts:
1. entering cadence arguments as a json string might be difficult unless you are copying them from somewhere... could we improve that in the future somehow?
2. would loading arguments from a file be a helpful addition?

MaxStalker:
> 1. entering cadence arguments as a json string might be difficult unless you are copying them from somewhere... could we improve that in the future somehow?

There are certain difficulties for inferring types that user want to pass - especially for complex structures 😅
But maybe, yeah 🤞
Right now we were thinking to use this option to pass arguments from codelens in VS Code, which would be formatted by extension and not manually.

MaxStalker:
> 2. would loading arguments from a file be a helpful addition?
That's an interesting idea! :)
Though we would still require user to format it in json-cadence way 🤔 
But it might be easier to do with IDE. Could be the next improvement 😉

robmyers:
> Right now we were thinking to use this option to pass arguments from codelens in VS Code, which would be formatted by extension and not manually.

People will try to use this from the command line, and in their own tool pipelines.

It may be worth adding a --dry-run or --print-args-and-exit flag that just outputs the parsed --args values and doesn't actually send the transaction, to make sure that there are no surprises for people doing that.

flow/arguments.go
```diff
@@ -0,0 +1,37 @@
+package cli
```
turbolent, line 1: This new file needs the license header, e.g. https://github.com/onflow/flow-cli/blob/master/flow/cadence/vscode/vscode.go#L1-L17.

The `check-headers.sh` script is broken, could you please copy the fixed version from https://github.com/onflow/cadence/blob/master/check-headers.sh and add it as part of this PR? It should detect the missing license header

flow/arguments.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type CadenceArgument struct {
+	Value cadence.Value
+}
+
+func (v CadenceArgument) MarshalJSON() ([]byte, error) {
+	return jsoncdc.Encode(v.Value)
+}
+func (v *CadenceArgument) UnmarshalJSON(b []byte) (err error) {
```
turbolent, line 16: Please run `go fmt` on the file

flow/arguments.go
```diff
@@ -0,0 +1,37 @@
...
+	"encoding/json"
+	"github.com/onflow/cadence"
+	jsoncdc "github.com/onflow/cadence/encoding/json"
+)
+
+type CadenceArgument struct {
+	Value cadence.Value
+}
+
+func (v CadenceArgument) MarshalJSON() ([]byte, error) {
```
turbolent, line 13: Good idea to create a wrapper value and implement the json marshalling interfaces 👍

flow/scripts/execute/execute.go
```diff
@@ -29,25 +30,37 @@ import (
 )
 
 type Config struct {
+	Args string `default:"" flag:"args" info:"arguments to pass to script"`
```
turbolent, line 33: Maybe mention in the info that the option requires input in JSON-Cadence format

flow/scripts/execute/execute.go
```diff
@@ -29,25 +30,37 @@ import (
...
+			cli.Exitf(1, "Failed to read script from %s", conf.Code)
 		}
 		projectConf := new(cli.Config)
 		if conf.Host == "" {
 			projectConf = cli.LoadConfig()
 		}
-		cli.ExecuteScript(projectConf.HostWithOverride(conf.Host), code)
+
+		// Arguments
+		scriptArguments := make([]cadence.Value, 0)
```
turbolent, line 55: The default value can just be an empty slice, no need to allocate a zero-element one

```suggestion
		var scriptArguments []cadence.Value
```

flow/transactions/send/send.go
```diff
@@ -31,17 +31,19 @@ import (
 )
 
 type Config struct {
-	Signer  string `default:"service" flag:"signer,s"`
+	Args    string `default:"" flag:"args" info:"arguments to pass to transaction"`
```
turbolent, line 34: Same as above, maybe mention in the info that the option requires input in JSON-Cadence format

flow/transactions/send/send.go
```diff
@@ -31,17 +31,19 @@ import (
...
 }
 
 var conf Config
 
 var Cmd = &cobra.Command{
-	Use:   "send",
-	Short: "Send a transaction",
+	Use:     "send",
+	Short:   "Send a transaction",
+	Example: `flow transactions send --code=tx.cdc --args="[{\"type\": \"String\", \"value\": \"Hello, Cadence\"}]"`,
```
turbolent, line 46: 👏

turbolent (commented):
Nice work! Just a couple things

flow/arguments.go
MaxStalker: All the credit goes to @psiemens 🙇‍♂️

flow/scripts/execute/execute.go
MaxStalker: Fixed 🙇‍♂️

turbolent (approved):
Nice work!

turbolent:
@psiemens could you please have a look?

turbolent:
@MaxStalker should be ready to go once the conflict is resolved

janezpodhostnik (approved):

PR #66, opened 2021-02-19
Rename deployment fix

sideninja:
Changes `deploy` to `deployments` as it is more clear
______

For contributor use:

### DIFF
--- a/flow/project/commands/deploy_contracts/deploy.go
+++ b/flow/project/commands/deploy_contracts/deploy.go
@@ -59,7 +59,7 @@ var Cmd = &cobra.Command{
 
 		// check there are not multiple accounts with same contract
 		if project.ContractConflictExists(conf.Network) {
-			fmt.Println("\n❌ Currently it is not possible to deploy same contract with multiple accounts, please check Deploys in config and make sure contract is only present in one account")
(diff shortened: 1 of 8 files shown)


psiemens (approved):
Looks good!

flow/project/commands/deploy_contracts/deploy.go
```diff
@@ -59,7 +59,7 @@ var Cmd = &cobra.Command{
 
 		// check there are not multiple accounts with same contract
 		if project.ContractConflictExists(conf.Network) {
-			fmt.Println("\n❌ Currently it is not possible to deploy same contract with multiple accounts, please check Deploys in config and make sure contract is only present in one account")
+			fmt.Println("\n❌ Currently it is not possible to deploy same contract with multiple accounts, please check Deployments in config and make sure contract is only present in one account")
```
10thfloor, line 62: Why is this the case? @psiemens @sideninja

psiemens: The tooling currently assumes that each contract is deployed to at most one account. This way we can import contracts by filename and then replace them with their on-chain address. If a contract is deployed in two places, which address do we replace it with?

I think there are ways we can support this in the future but we decided it wasn't worth solving right now.

PR #77, opened 2021-02-23
Kan/sign transaction

Kay-Zee:
## Description

Allows signing a transaction without sending it.

Saves it to a file, allowing transport to other parties to finish the signing process.

Needed for multi sign use cases

______

For contributor use:


flow/transactions/sign/sign.go
```diff
@@ -0,0 +1,139 @@
...
+		for _, authorizerString := range conf.AdditionalAuthorizers {
+			authorizerAddress := flow.HexToAddress(authorizerString)
+			tx.AddAuthorizer(authorizerAddress)
+		}
+
+		tx = cli.SignTransaction(projectConf.HostWithOverride(conf.Host), signerAccount, signerRole, tx, payer)
+
+		fmt.Printf("%s encoded transaction written to %s\n", conf.Encoding, conf.Output)
+
+		output := fmt.Sprintf("%x\n", tx.Encode())
```
Kay-Zee, line 100: This encodes the file as hex, should we bother with this step? or just write bytes out to the file?

psiemens: I don't think it hurts to write as hex 👍

flow/transactions/sign/sign.go
```diff
@@ -0,0 +1,139 @@
...
+		err = ioutil.WriteFile(conf.Output, []byte(output), os.ModePerm)
```
sideninja, line 101: I think it can be a good idea to allow users to choose to write this to stdout as well as it will allow chaining of commands UNIX style. Maybe it would even be a good idea to do that as a default and if a user explicitly specifies the filename it saves it there, my thinking here is that command is called `sign` which implies signing operation more than saving - a similar example to `generate keys` command where you have `save` as a separate option. I think this could be a small change that would add to more consistency maybe.

bjartek:
Interesting, this is getting my head spinning with ideas.

flow/transactions/sign/sign.go
Kay-Zee: addressed in #110 which is now merged

sideninja (approved):

janezpodhostnik (approved):
Nice! This covers a lot of cases that weren't covered before.

With multiple authorisers It might be needed to do something like this in the future:

Auth1 (proposer): prepare tx with all authoriser addresses and payer address and sign it -> tx.rlp
Auth2: load tx.rlp, sign it  -> tx2.rlp
Payer: load tx2.rlp, sign it, send it

Kay-Zee:
For sure, will need a bit of work to enable multi payload signs. It will basically involve moving the "partial-tx" logic from the send side. I will do that in a separate PR.

sideninja:
Maybe it's better if you wait with this just so we merge the
#refactor-sharedlib because that changes a lot of code and it will also be
easier to do that after. As you wish just a flag.

PR #81, opened 2021-02-24
Migrate config

### LINKED ISSUE #55: Migrate new configuration format to entire CLI
With implementing the functionality described in issue #48 there was a refactor of configuration files and commands. New configuration implementation is structured in two layers. 
There are abstract configuration structures and configuration structures that relate to specific configuration type (like json). 

We should migrate this configuration to be a single point of truth for CLI for configuration. In order to achieve this we would need to:
- make sure old configuration format is supported (implement a parser for old JSON format and convert it to current configuration)
- replace all config references through-out the CLI and make sure they are calling the new configuration 
- make sure all the types match with the new configuration
- account creation should be saved to the new configuration structure and also retrieved 
- emulator command should use the new configuration

This migration might also open other questions such as:
- should we have a single initialization for config or not? currently, we have a command `flow init` and `flow project init`.

sideninja:
Closes #55

## Description

Migrated a new configuration schema to the whole CLI.

______

For contributor use:


flow/keys/save/kms/kms.go
```diff
@@ -39,46 +41,57 @@ type Config struct {
...
+		account, err := cli.AccountFromConfig(
+			config.Account{
+				Name:    flags.Name,
+				Address: address,
+				ChainID: flow.Emulator, // TODO: don't hardcode this
+				Keys:    keys,
+			},
+		)
+
+		/* TODO: discuss how this changed
```
sideninja, line 85: @psiemens  would need some of your feedback on this, whole kms is a bit confusing to me, also what is difference between enums kms and kms-google

sideninja:
I can resolve merge conflicts, but before I need to clarify some issue with @psiemens

flow/accounts/add-contract/add.go
```diff
@@ -22,27 +22,29 @@ import (
...
-	Host     string   `flag:"host" info:"Flow Access API host address"`
-	Results  bool     `default:"false" flag:"results" info:"Display the results of the transaction"`
+type Flags struct {
+	Signer  string `default:"service" flag:"signer,s"`
+	Host    string `flag:"host" info:"Flow Access API host address"`
+	Results bool   `default:"false" flag:"results" info:"Display the results of the transaction"`
 }
 
-var conf Config
+var flags Flags
```
psiemens, line 37: Nice, good idea 👍

flow/accounts/add-contract/add.go
```diff
@@ -52,21 +54,24 @@ var Cmd = &cobra.Command{
 			cli.Exitf(1, "Failed to read contract from source file %s", contractFilename)
 		}
 
-		signerAccount := projectConf.Accounts[conf.Signer]
+		signerAccount := project.GetAccountByName(flags.Signer)
```
psiemens, line 57: 😍 The new project struct works like a charm!

flow/cli/project.go
```diff
@@ -153,6 +140,14 @@ func (p *Project) ContractConflictExists(network string) bool {
 	return len(all) != len(uniq)
 }
 
+func (p *Project) HostWithOverride(host string) string {
```
psiemens, line 143: Maybe makes sense to later update this to `HostWithOverride(network, hostOverride string)`

flow/initialize/init.go
```diff
@@ -22,76 +22,68 @@ import (
...
-		if !cli.ConfigExists() || conf.Reset {
-			var pconf *cli.Config
-			serviceKeySigAlgo := crypto.StringToSignatureAlgorithm(conf.ServiceKeySigAlgo)
-			serviceKeyHashAlgo := crypto.StringToHashAlgorithm(conf.ServiceKeyHashAlgo)
-			if len(conf.ServicePrivateKey) > 0 {
-				serviceKey := cli.MustDecodePrivateKeyHex(serviceKeySigAlgo, conf.ServicePrivateKey)
-				pconf = InitProjectWithServiceKey(serviceKey, serviceKeyHashAlgo)
-			} else {
-				pconf = InitProject(serviceKeySigAlgo, serviceKeyHashAlgo)
+
```
psiemens, line 46: I wonder if we should just go ahead and remove this command entirely and replace with `flow project init`.

This implementation is outdated and tightly coupled to the emulator, which I don't like.

flow/keys/save/hex/hex.go
```diff
@@ -40,81 +40,89 @@ type Config struct {
 	Overwrite bool   `flag:"overwrite" info:"bool indicating if we should overwrite an existing config with the same name in the config file"`
 }
 
-var conf Config
+var flags Flags
 
 var Cmd = &cobra.Command{
 	Use:     "hex",
 	Short:   "Save a hex key to the config file",
 	Example: "flow keys save hex --name test --address 8c5303eaa26202d6 --sigalgo ECDSA_secp256k1 --hashalgo SHA2_256 --index 0 --privatekey <HEX_PRIVATEKEY>",
```
psiemens, line 48: Hmm, I think we should rethink this command 🤔 

I'm not sure it makes sense as a `keys` command.

flow/keys/save/kms/kms.go
psiemens: This code is pretty new to me, too. I'll have to study it more with fresh eyes, but I think it would be good to try to move most of this logic into the new `keys` package.

flow/transactions/send/send.go
```diff
@@ -82,20 +89,20 @@ func initConfig() {
 }
 
 func validateKeyPreReq(account *cli.Account) {
-	if account.KeyType == cli.KeyTypeHex {
+	if account.DefaultKey().Type() == config.KeyTypeHex {
 		// Always Valid
-		return
-	} else if account.KeyType == cli.KeyTypeKMS {
+		return // TODO: check difference between googleKMS and KMS
```
psiemens, line 94: In this case, Google KMS and KMS are the same thing. I think we should always include "Google" in the name, in case we choose to implement other key management services (KMS) in the future.

flow/keys/save/kms/kms.go
```diff
@@ -39,46 +41,57 @@ type Config struct {
...
+			SigAlgo:  crypto.StringToSignatureAlgorithm(flags.SigAlgo),
+			HashAlgo: crypto.StringToHashAlgorithm(flags.HashAlgo),
+			Context:  keyContext,
+		}}
+
+		account, err := cli.AccountFromConfig(
+			config.Account{
+				Name:    flags.Name,
+				Address: address,
+				ChainID: flow.Emulator, // TODO: don't hardcode this
```
psiemens, line 80: I'm starting to question the value of having `ChainID` as part of the account config...

flow/emulator/emulator.go
```diff
@@ -37,37 +33,33 @@ var Cmd = &cobra.Command{
...
-	if init {
-		pconf := initialize.InitProject(sigAlgo, hashAlgo)
-		serviceAcct = pconf.ServiceAccount()
+	serviceKey := serviceAccount.Keys[0]
 
-		fmt.Printf("⚙️   Flow client initialized with service account:\n\n")
-		fmt.Printf("👤  Address: 0x%s\n", serviceAcct.Address)
-	} else {
-		serviceAcct = cli.LoadConfig().ServiceAccount()
+	privateKey, err := crypto.DecodePrivateKeyHex(serviceKey.SigAlgo, serviceKey.Context["privateKey"])
```
psiemens, line 50: Hmm, this is a bit brittle. 

I guess _only_ hex keys can be used as the service account key, right?

I also realized that `proj.EmulatorServiceAccount()` returns type `config.Account`, but really it should return [this Account type](https://github.com/onflow/flow-cli/blob/master/flow/project/cli/project.go#L237-L242), right?

That way you could do this:

```
serviceKeyHex, ok := serviceKey.(*keys.HexAccountKey)
if !ok {
  cli.Exit(1, "Only hexadecimal keys can be used as the emulator service account key."
}

privateKey := serviceKeyHex.PrivateKey()
```

This could require you to add this function to `keys.HexAccountKey`:

```go
func (a *HexAccountKey) PrivateKey() crypto.PrivateKey {
	return a.privateKey
}

File: flow/config/json/emulator_test.go [hunk 10]
```diff
@@ -39,6 +39,6 @@ func Test_ConfigEmulatorSimple(t *testing.T) {
 
 	emulators := jsonEmulators.transformToConfig()
 
-	assert.Equal(t, "default", emulators[0].Name)
-	assert.Equal(t, 9000, emulators[0].Port)
+	assert.Equal(t, emulators[0].Name, "default")
+	assert.Equal(t, emulators[0].Port, 9000)
```
psiemens, line 43: I think these were actually the correct order before

psiemens (commented):
Looking great!

I'll take another look in the morning with fresh eyes. This got me thinking more about how we could approach a refactor

flow/keys/save/hex/hex.go
sideninja: yes, I agree. Maybe we could that as a part of refactoring commands and writing guidelines so we leave this refactor as purely focused on moving config to the whole project. I had to really break myself to not change some commands because I think there is a lot of good improvements that can be made on the command namings.

flow/keys/save/kms/kms.go
sideninja: Let's please have a conversation on this topic because I also think it's really non-consistent currently (it's defined on networks, somewhere on accounts etc...)

10thfloor: I found this confusing as well, when trying to set up the config.

sideninja:
I merged #95 in here so it will be possible to merge to master after without conflicts

sideninja:
All changes were included in #104

PR #82, opened 2021-02-24
Private config env

### LINKED ISSUE #70: Configuration for private accounts in separate configuration
### Issue To Be Solved

We believe the CLI configuration file `flow.json` should be included in the versioning control (github...) similar to how for example `package.json` is.

Since configuration can include sensitive private keys we should allow developers to separate the configuration for their testnet and mainnet accounts from the main `flow.json` configuration so they are not mistakenly committed to a versioning control systems. 

Currently, the configuration looks like this (where secret gets constantly replaced before committing - a ticking bomb):

```
...
	"accounts": {
		"testnet-account": {
			"keys": "SECRET"
		}
	},
...
```

### Suggest A Solution

After a brief discussion with @psiemens we believe there are at least the following solutions to this problems:

#### Seperate Private Config File
We can create a separate configuration file - let's call it `flow.private.json` for the sake of the example and this file could include any number of accounts that would on the runtime of CLI command be added to the original `flow.json` configuration (but not saved to the file). Use of this would happen "in-memory". 

Example for `flow.private.json`:

```
...
	"accounts": {
		"testnet-account": {
			"keys": "ACTUAL_SECRET_WOULD_BE_HERE"
		}
	},
...
```

And then `flow.json` could look like:

```
...
	"accounts": {
                "emulator-account": {
			"keys": "11c5dfdeb0ff03a7a73ef39788563b62c89adea67bbb21ab95e5f710bd1d40b7",
			"chain": "flow-emulator"
		},
		"testnet-account": {
                       "private": true
                }
	},
...
```

Or another option:

```
...
	"accounts": {
                "emulator-account": {
			"keys": "11c5dfdeb0ff03a7a73ef39788563b62c89adea67bbb21ab95e5f710bd1d40b7",
			"chain": "flow-emulator"
		},
		"testnet-account": "flow.private.json"
	},
...
```
There is also a possibility to not change anything in flow.json and just for the CLI to figure out if the file flow.private.json exists and overwrite it on the fly.

#### Environment variable support

We could also allow passing account secret as an environment variable. This can be a standalone solution or it can work nicely with the above solution.

Example:

`flow project deploy --acount=testnet --secret=123`

Example `flow.json`:

```
...
	"accounts": {
		"testnet-account": {
			"keys": "$ENV.secret"
		}
	},
...
```

### LINKED ISSUE #84: Configuration for private accounts in environment variables
### Issue To Be Solved

@psiemens wrote:
I think the highest priority is to give users a way to securely specify a few pieces of private data, and that can be solved easily and quickly with environment variable support. Do you think we could implement support for environment variables in this sprint, and then do flow.private.json as a follow-up feature?

I was looking at other examples and it looks like this syntax is quite common:

```
...
	"accounts": {
		"testnet-account": {
			"keys": "${env:MY_PRIVATE_KEY}"
		}
	},
...
```

A few examples:

https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables
https://www.serverless.com/framework/docs/providers/aws/guide/variables#referencing-environment-variables
https://www.elastic.co/guide/en/beats/winlogbeat/current/using-environ-vars.html

sideninja:
Closes #84  

## Description

Implement support for passing env variables to the configuration.

______

For contributor use:


flow/project/cli/config/json/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+		},
+		"deploys": {}
+	}`)
+
+	err := ioutil.WriteFile("test1-flow.json", b, os.ModePerm)
+	require.NoError(t, err)
+
+	conf, loadErr := Load("test1-flow.json")
+	os.Remove("test1-flow.json")
```
psiemens, line 58: I would add a separate `json.LoadString` function and then remove IO operations from the tests. These can be unreliable in CI environments.

I don't think it's too valuable to be testing the actual IO operations in this case, but if it was, the better solution would be to mock out the filesystem entirely.

psiemens:
Nice and simple solution! 👏

flow/project/cli/config/json/config_test.go
sideninja: No, I really don't like my solution but wasn't aware of an alternative - still fresh in go. I will adapt.

psiemens:
Going to close this in favour of #91

Issue #80, opened 2021-02-24
Shared library for CLI command logic


psiemens:
### Issue To Be Solved

With the new improvements to `flow.json` and the `flow project` command, we want to integrate this functionality into the Flow VS Code extension to provide a more seamless development environment.

To do so, we'd like to explore embedding more functionality into the Cadence language server, implemented here: https://github.com/onflow/cadence

I think it'd be great if we could accomplish this by moving shared functionality into a Go module that can be imported by the language server.

@MaxStalker @sideninja We'll go into more detail in our call later, but here are a few points to start off:
- VS Code should display accounts from `flow.json`
- VS Code should be able to execute transactions with the same address replacement functionality that is present in `flow project deploy`
- If the CLI supports loading contracts from remote locations (e.g. URL, git), then the VS Code extension should support this, too.
- The CLI and the VS Code extension should both support transaction/script arguments
- The CLI and the VS Code extension should both support multiple signers per transaction
- the list probably goes on...

psiemens:
Hey @sideninja and @MaxStalker, I've thought about this more since our last conversation.

Currently the language server (LSP) has the following commands:

- [CommandSubmitTransaction](https://github.com/onflow/cadence/blob/d19d99342ead3831e08c1169ae60d20b5c2499bb/languageserver/integration/commands.go#L87)
- [CommandExecuteScript](https://github.com/onflow/cadence/blob/d19d99342ead3831e08c1169ae60d20b5c2499bb/languageserver/integration/commands.go#L133)
- [CommandDeployContract](https://github.com/onflow/cadence/blob/d19d99342ead3831e08c1169ae60d20b5c2499bb/languageserver/integration/commands.go#L316)
- [CommandCreateAccount](https://github.com/onflow/cadence/blob/d19d99342ead3831e08c1169ae60d20b5c2499bb/languageserver/integration/commands.go#L244)
- ~CommandCreateDefaultAccounts~
- ~CommandSwitchActiveAccount~

I don't think we need the last two commands if we move to this new structure.

As a first step towards unifying the CLI and the language server, I think we should re-implement the above functions to use the core logic from the CLI repo (this repo). The language server commands will **still call the emulator via RPC**, but they will do it using shared code from the CLI.

Before we can do this, I think we need two things:

### Language server can load `flow.json` config

I see two ways to do this:
- Language server reads `flow.json` directly from the filesystem
[...]

PR #83, opened 2021-02-24
Design guidelines

### LINKED ISSUE #59: CLI design guidelines
### Issue To Be Solved

We should update the [CONTRIBUTING.md](https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md) document to include design guidelines for the CLI. 

This should includes things like:
- Command structure
- Arguments and flag format
- Documentation standards

sideninja:
Closes #59

## Description

Design guidelines for CLI. Define some common and industry standard architecture guidliness.

### DIFF
--- a/docs/GUIDELINES.md
+++ b/docs/GUIDELINES.md
@@ -1,24 +1,40 @@
-# CLI Style Guide
-This is a design guideline used for development of Flow CLI. Purpose of this guidline is to achieve consistency across new features and allow composability of commands that build the fundaments of great CLI design.
+# CLI Guidelines
+This is a design guideline used for development of Flow CLI. Purpose of this guidline 
+is to achieve consistency across new features and allow composability of commands that 
+build the fundaments of great CLI design.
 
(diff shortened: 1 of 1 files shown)


bjartek:
Cobra has pretty good support for tabcompletions in all major shells these days, should that be incorporated into this guide?

bjartek:
Another part that is missing in these guidelines is how to test the CLI. I would recommend keeping the body of the cobra commands as terse as possible 

1. parse and validate input into a struct
2. send that struct to a service layer (or pkg)
3. after you return transfer the errors 

If we structure the code this way, it is way easier to use the flow-cli tool as a dependency in another CLI tool if you want to make use of its features. It just composes better.

sideninja:
> Cobra has pretty good support for tabcompletions in all major shells these days, should that be incorporated into this guide?

I had this in mind since day one. For sure should be part of the guidelines 💯 

> Another part that is missing in these guidelines is how to test the CLI. I would recommend keeping the body of the cobra commands as terse as possible

Haha already on it. Look here: #80 There will be a major refactor happening in CLI in the following days.

PR #87, opened 2021-02-25
Private config file

### LINKED ISSUE #70: Configuration for private accounts in separate configuration

sideninja:
Closes #70 

## Description
Implemented private configuration file option

______

For contributor use:


flow/project/cli/config/config.go
```diff
@@ -118,6 +118,8 @@ func (c *Contracts) GetByNameAndNetwork(name string, network string) Contract {
 	return contracts[0]
 }
 
+// REF: this filtering can cause error if not found, better to refactor to returning
```
psiemens, line 121: I would turn your `REF`s into `TODO`s, since most of these notes refer to future work/improvements

flow/project/cli/config/config.go
```diff
@@ -133,10 +135,14 @@ func (c *Contracts) GetByNetwork(network string) Contracts {
...
-		return a.Name == name
-	}).([]Account)[0]
+func (a *Accounts) GetByName(name string) *Account {
+	for _, account := range *a {
+		if account.Name == name {
+			return &account
+		}
+	}
+
+	return nil
```
psiemens, line 145: Nice. Even though it's more code, I think this is easier to read

flow/project/commands/start_emulator/emulator.go
```diff
@@ -41,7 +41,7 @@ func configuredServiceKey(
 	proj := cli.LoadProject()
 
 	serviceAccount := proj.EmulatorServiceAccount()
-
+	// REF: this shouldn't be accessed like this, make a getter
```
psiemens, line 44: 👍

flow/project/commands/deploy_contracts/deploy.go
```diff
@@ -148,6 +148,7 @@ var Cmd = &cobra.Command{
 		if len(errs) == 0 {
 			fmt.Println("\n✅ All contracts deployed successfully")
 		} else {
+			// REF: better output when errors
```
psiemens, line 151: Good idea. This could even be an issue linked in the code:

```suggestion
			// TODO: better output when errors: ISSUE_LINK
```

flow/project/cli/config/processor_test.go
```diff
@@ -0,0 +1,282 @@
...
+		},
+		"accounts": {
+			"emulator-account": {
+				"address": "f8d6e0586b0a20c7",
+				"keys": "${env:TEST}",
+				"chain": "flow-emulator"
+			}
+		},
+		"deploys": {}
+	}`
```
psiemens, line 55: In order to improve the readability of these tests, do you think it would be valid to reduce the input to be just the portion of the config you are testing, or does that make the test less effective?

```suggestion
	test := `{
		"accounts": {
			"emulator-account": {
				"address": "f8d6e0586b0a20c7",
				"keys": "${env:TEST}",
				"chain": "flow-emulator"
			}
		}
	}`
```

flow/project/cli/config/processor.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+)
+
+var (
+	envRegex  *regexp.Regexp = regexp.MustCompile(`\$\{env\:(.+)\}`)
+	fileRegex *regexp.Regexp = regexp.MustCompile(`\"\$\{file\:(.+)\}\"`)
+)
+
+// processEnv finds env variables and insert env values
+func processEnv(raw string) string {
```
psiemens, line 35: Great idea to add this processor pattern 👏 I really like it

flow/project/cli/config/processor.go
```diff
@@ -0,0 +1,82 @@
...
+			match[0],
+			getEnvVariable(match[1]),
+		)
+	}
+
+	return raw
+}
+
+// processFile finds file variables and insert content
+func processFile(raw string) string {
```
psiemens, line 50: Here's one way to implement this in order to mock out the actual filesystem in tests.

[Afero](https://github.com/spf13/afero) is an abstraction system for filesystems that I think could work well in this case.

```go
import "github.com/spf13/afero"

type Preprocessor struct {
  filesystem afero.Fs
}

func NewPreprocessor(filesystem afero.Fs) *Preprocessor {
  return &Preprocessor{filesystem: filesystem}
}

func (p *Preprocessor) Run(raw string) string {
  raw = p.processEnv(raw)
  raw = p.processFile(raw)
  return raw
}

func (p *Preprocessor) processFile(raw string) string {
  // ...
  
  configFileRaw, err := p.filesystem.Open(match[1])
  
  // ...
}
```

```go
import "github.com/spf13/afero"

// in real implementation
preprocessor := NewPreprocessor(afero.NewOsFs())
// ...
```

```go
import "github.com/spf13/afero"

// in tests
mockFS := afero.NewMemMapFs()

afero.WriteFile(mockFS, "src/a/b", []byte("file b"), 0644)

preprocessor := NewPreprocessor(mockFS)
// ...
```

psiemens (commented):
I really like the way you wrote this code!

This is different than the first idea we talk about, but I think this is great! I can see this being really useful for users who want to store pieces of their config in different files.

However, do you think it would be valuable to _also_ have the `flow.private.json` file as well?

(👇 copied from this issue: https://github.com/onflow/flow-cli/issues/70)

**`flow.private.json`**
```
...
	"accounts": {
		"testnet-account": {
			"keys": "ACTUAL_SECRET_WOULD_BE_HERE"
		}
	},
...
```

**`flow.json`**
```
...
	"accounts": {
                "emulator-account": {
			"keys": "11c5dfdeb0ff03a7a73ef39788563b62c89adea67bbb21ab95e5f710bd1d40b7",
			"chain": "flow-emulator"
		},
		"testnet-account": {
                       "private": true
                }
	},
...
```

The main reason I like this approach is because it would allow for commands like this:

```sh
flow create-account "my-private-account" --save --private
```

- `--save` flag says to save the new account to config
- `--private` flag says to save as a private account in `flow.private.json`

Alternatives:
```sh
flow create-account "my-private-account" --save-private
flow create-account "my-private-account" --save-public
```

HOWEVER, this approach is more implicit. I think I'm okay with the implicitness in this case, but curious to hear your opinion.

sideninja:
> I really like the way you wrote this code!
> 
> This is different than the first idea we talk about, but I think this is great! I can see this being really useful for users who want to store pieces of their config in different files.
> 
> However, do you think it would be valuable to _also_ have the `flow.private.json` file as well?
> 
> (👇 copied from this issue: #70)
> 
> **`flow.private.json`**
> 
> ```
> ...
> 	"accounts": {
> 		"testnet-account": {
> 			"address": "f8d6e0586b0a20c7",
> 			"keys": "ACTUAL_SECRET_WOULD_BE_HERE"
> 		}
> 	},
> ...
> ```
> 
> **`flow.json`**
> 
> ```
> ...
> 	"accounts": {
>                 "emulator-account": {
> 			"address": "f8d6e0586b0a20c7",
> 			"keys": "11c5dfdeb0ff03a7a73ef39788563b62c89adea67bbb21ab95e5f710bd1d40b7",
> 			"chain": "flow-emulator"
> 		},
> 		"testnet-account": {
>                        "private": true
>                 }
> 	},
> ...
> ```
> 
> The main reason I like this approach is because it would allow for commands like this:
> 
> ```shell
> flow create-account "my-private-account" --save --private
> ```
> 
> * `--save` flag says to save the new account to config
> * `--private` flag says to save as a private account in `flow.private.json`
> 
> Alternatives:
> 
> ```shell
> flow create-account "my-private-account" --save-private
> flow create-account "my-private-account" --save-public
> ```
> 
> HOWEVER, this approach is more implicit. I think I'm okay with the implicitness in this case, but curious to hear your opinion.

I've actually implemented it in the way it was discussed and then remove the code and reimplemented in this way. I think the big benefits of having implemented it this way are:

1. You are able to have different files based on env. I think this one is major. I'm sure this will be used with testnet, mainet and different accounts and with current implementation you can do:

`flow.json`
```
{
...
    "accounts": {
        "admin-account": "${env:ENV}"
    }
...
}
```
`testnet.admin.json`
```
{
    "address": "123",
    "keys": "123"
}
```
`mainnet.admin.json`
```
{
    "address": "222",
    "keys": "333"
}
```

And then you can run it by:
```
ENV=${file:mainnet.admin.json} flow project deploy
```
or 
```
ENV=${file:testnet.admin.json} flow project deploy
```

Whereas having `flow.private.json` file will work only for one account on one network, and even if we define multiple accounts in it it starts to be confusing how we manage that file.

2. Anoher point is that this implementation works for anything, it is not only restricted to accounts. You never know maybe some other parts of the config will be required to keep locally and with this solution you can do this, whereas by using `--private` option we are kind of implicitly using it for accounts since the `private: true` is included in the account object.   


3. I think saving to files should be unified all over the CLI as it is now confusing what is saving where is saving etc. I think saving should be explicit and never default (as this is a CLI and default should be stdout so it allows piping etc). Thinking this way my solution to your problem would be different:

```
flow create-account --save admin.flow.json
```

I believe all saving should be explicit like that. In case you want to save something be explicit where unless it is an `flow init` where we make a default action to save to `flow.json` but even there you could do `flow init --save testnet.flow.json` 

We can implement this notion of private accounts but I think it might be another concept developers will need to understand whereas your idea of having `${env:something}` and now `${file:something}` feels much more consistent and easy to understand. 

4. Lastly and this one is not as important since I put good UX first but still this solution has a less complex implementation, it doesn't require to have more options in parsing JSON and one day in other formats.

Let me know if this is convincing or else I have no problem implementing the way you stated.

flow/project/cli/config/processor.go
sideninja: awesome, I will implement this

flow/project/cli/config/processor_test.go
sideninja: I agree. Will change.

flow/project/cli/config/config.go
sideninja: Yes, you were right. I will change all of them to be this way.

psiemens:
These are all great points! 

> 1. You are able to have different files based on env. I think this one is major. I'm sure this will be used with testnet, mainet and different accounts and with current implementation you can do:

Yep. I totally overlooked this. Users will likely want to store their mainnet accounts separately from their testnet accounts, for example. However, I don't think this necessarily has to be done with an environment variable. It should be possible, though.

> I believe all saving should be explicit like that. In case you want to save something be explicit where unless it is an flow init where we make a default action to save to flow.json but even there you could do flow init --save testnet.flow.json

I agree with this. Configurable options with sensible defaults 💯 

> Anoher point is that this implementation works for anything, it is not only restricted to accounts. 

I also hadn't considered this. However, I kind of think this is a different use case.

---

I'm going to try to explain the two use cases I see.

#### Use Case 1: Configure a project to _require specific accounts_ without publishing private information

In the case of Kitty Items, we know we need the following accounts: `emulator-account` and `testnet-account`.

However, `testnet-account` is different for each user, so we can't include it in the `accounts` section of the config.

If a user clones Kitty Items (or any other project), they might do this, which will fail:

```sh
# this will fail because testnet-account is not defined
flow projects deploy --network testnet
```

It would be really nice if the CLI could actually guide the user through this error:

```
> flow projects deploy --network testnet

The account with name testnet-account has not yet been created.
Use the following command to create it:

  flow create-account --name testnet-account --save accounts.testnet.json
```

```js
{
  // ...
  "accounts": {
    "testnet-account": { "fromFile": "accounts.testnet.json" }
  }
  // ...
}
```

Also, if somebody wants to add a _new_ account, the `create-account` command could automatically link it in the main `flow.json`:

```
flow create-account --name mainnet-account --save accounts.mainnet.json
```

```js
{
  // ...
  "accounts": {
    "testnet-account": { "fromFile": "accounts.testnet.json" },
    "mainnet-account": { "fromFile": "accounts.mainnet.json" }
  }
  // ...
}
```

Lastly, this format allows us to define multiple accounts in the same file:

```js
{
  // ...
  "accounts": {
    "testnet-account-1": { "fromFile": "accounts.testnet.json" },
    "testnet-account-2": { "fromFile": "accounts.testnet.json" },
  }
  // ...
}
```

#### Use Case 2: Use a different version of config in each environment (emulator, testnet, mainnet)

```
ENV=${file:mainnet.admin.json} flow project deploy
"admin-account": "${file:${env:ENV}.admin.json}"
```

To be honest, I find usage patterns like this to be a bit confusing.

I think this use case is best solved with the following:

```
flow project deploy -f flow.json -f flow.mainnet.json
```

Configuration would be merged from left to right. In the example above, `flow.json` is the base, and then `flow.mainnet.json` can add or override specific values. This is inspired by the [Docker Compose CLI](https://docs.docker.com/compose/reference/overview/#use--f-to-specify-name-and-path-of-one-or-more-compose-files), which I quite like.

#### Summary

Overall, I think it's worth it to have _first class_ support for these separate `accounts.*.json` files because account information can be _personal_ to each developer on a project, whereas multiple config versions (testnet, mainnet) are more likely shared by an entire team.

I tried to incorporate the points you brought up -- let me know what you think!

sideninja:
> > flow projects deploy --network testnet
> 
> The account with name testnet-account has not yet been created.
> Use the following command to create it:
> 
>   flow create-account --name testnet-account --save accounts.testnet.json

I agree. I really wanted to implement this, but wanted to have a separate discussion on this (it also affects guidelines). All these scenarios where some data is missing should have a response like this. I'm hoping to include that in the PR where we will go through all commands and tidy things up. However, at that time we should decide if we go with a static response like above, or prompts (where users are asked to enter missing data). In any case, we should implement this kind of response here, I agree. 





> ```js
> {
>   // ...
>   "accounts": {
>     "testnet-account": { "fromFile": "accounts.testnet.json" }
>   }
>   // ...
> }
> ```

Ok, I understand where you are coming from. I like this solution as well. 

> flow create-account --name mainnet-account --save accounts.mainnet.json

I think this is important functionality. We should in any case provide project manipulation tools asap. By that I mean that you don't have to manually edit config but you can have commands that add things to config (one good example is adding contract for deployment).



> ENV=${file:mainnet.admin.json} flow project deploy
> "admin-account": "${file:${env:ENV}.admin.json}"

Yeah, tbh I didn't like that solution myself, I think we should avoid any weird formats on the CLI input (encoded json, raw formats etc...).



> flow project deploy -f flow.json -f flow.mainnet.json

This is much better. 

I will reimplement this in order to fits the new specs.

psiemens:
Closed in favour of #95

PR #91, opened 2021-02-26
added envsubst

### LINKED ISSUE #84: Configuration for private accounts in environment variables

bjartek:
Closes #84 

Added envsubst and applied it to loading the proejct file

______

For contributor use:

### DIFF
--- a/flow/project/cli/config/json/config.go
+++ b/flow/project/cli/config/json/config.go
@@ -26,7 +26,6 @@ import (
 	"os"
 
 	"github.com/a8m/envsubst"
-	"github.com/davecgh/go-spew/spew"
 	"github.com/onflow/flow-cli/flow/project/cli/config"
 )
 
--- a/flow/project/cli/config/json/config.go
+++ b/flow/project/cli/config/json/config.go
@@ -25,6 +25,8 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/a8m/envsubst"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/onflow/flow-cli/flow/project/cli/config"
 )
 
@@ -77,7 +79,7 @@ func Save(conf *config.Config, path string) error {
 var ErrDoesNotExist = errors.New("project config file does not exist")
 
 func Load(path string) (*config.Config, error) {
-	f, err := os.Open(path)
+	f, err := envsubst.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil, ErrDoesNotExist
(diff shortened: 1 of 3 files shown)


bjartek:
Should probably also update some documentation somewhere but I do not have the overview to see where that is.

bjartek:
I get some test failure here that I do not understand, they are realted to code that I have not touched at all.

psiemens:
Thanks for adding this! However, I think it'd a dupe of #87.

Also @sideninja and I thought we could use the `${env:VAR_NAME}` syntax. What do you think of that?

bjartek:
Ok, I was s little confused.

https://github.com/drone/envsubst is another library that could have been used, I did not review them either that should probably have been done.

From my understanding neither of those librarires support your suggested syntax. If you feel that the cost og implementing your specifix syntax is worth it then do that. However the libraries have many features and emulates envsubst from bash that is a pattern that are well known.

In cases like theese I fins that I like to spend my innvoation points elsewhere and use a standard sollution. The choice however is not mine, it is yours :)

Really happy that we can focus on these issues do that it will be easier to get new devs up to speed with good tools.

sideninja:
@bjartek  I like your implementation and I agree we should leverage the work of others by using a library. I want it to be our choice ;) 

The only comment I would have on this implementation is that I think it's a good design decision to put that functionality in this file: https://github.com/onflow/flow-cli/blob/private-config-file/flow/project/cli/config/processor.go 

I would keep that design of having a "preprocessor" of config files so the JSON config is not affected by this change, in my current refactoring I also removed loading of files from JSON as I don't think that is the correct place for loading files and hence your change won't work there. Let me know what you think.

bjartek:
That makes total sense to me.

bjartek:
That file has not been merge yet, can we just merge this and fix it later?

psiemens:
> That file has not been merge yet, can we just merge this and fix it later?

@bjartek I'm good with that. I'll approve and merge 👍

psiemens (approved):
Nice and simple 👌

Issue #88, opened 2021-02-25
Project manipulation commands


sideninja:
### Issue To Be Solved

Project subset of commands should allow easy usage without manual manipulation of flow.json. Manual manipulation is good for an advance usage but it is prone to typos, errors and more complex to understand without docs.

### Suggest A Solution

The current developer flow of using CLI project is:
1. Setup project by running `project init`
2. Add some contracts to the config
3. Add some accounts to the config
4. Deploy contracts by running `project deploy`
5. Update contracts by running `project deploy --update`

Steps 2 and 3 should have a command alternative. Adding accounts should be possible with a command maybe similar to `account create --save` and contracts should be possible to add with command like `project add-contract`. Command names are not important right now (they will be further defined with CLI guidelines issue #59 and command refactor).

sideninja:
Adding project manipulation methods. I strongly believe this is a must. Also Bjartek said mulitple times and opened a PR on add contract but this goes further, we should allow adding all resources to the config via command.

```shell
project init
project deploy
project add <account|contract|deployment|network>
project remove <account|contract|deployment|network>**  
```

There should also be a guided prompt way to add resources, let me give an example:

```shell
project add deployment
Contracts in the configuration:
[0] FungibleToken
[1] Kibble
[2] KibbleMarket
Choose the contract for deployment: 1

Networks in the configuration:
[0] Emulator
[1] Testnet
[2] Mainnet
Choose the network for deployment: 1

✅ Contract 'Kibble' will be deployed to 'Testne' when you run 'flow project deploy'
```

sideninja:
closing in favor #124

PR #92, opened 2021-02-26
added command to add contract to the flow.json

bjartek:
#88 

## Description

Command to add an contract to the project struct

______

For contributor use:

### DIFF
--- a/flow/project/cli/project.go
+++ b/flow/project/cli/project.go
@@ -1,0 +1,322 @@
+/*
+ * Flow CLI
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


bjartek:
I see here that i used args and not flags, that change is the last commit so we might want to revert that. IMHO this reads better with args, since it is just key=value, but if you want to add an alias to testnet, mainnet I can see the argument for flags.

sideninja:
@bjartek great work :) 

I think this args/flags dilemma will be solved when we finish the CLI guideline as at that time we will go through all commands and make them consistent. I think args is fine for now.

bjartek:
I tried to rebase this off newest master, but not sure if it even makes sense anymore. Should we just delete this and start anew?

sideninja:
I was thinking this to become part of #124 (init section in the comment) where we create a `config` top command that has subcommands for adding/removing/viewing config values.

PR #104, opened 2021-03-05
Refactor CLI for shared library

### LINKED ISSUE #80: Shared library for CLI command logic

sideninja:
Closes #80 

## Description

CLI shared library refactor. 

______

For contributor use:


cmd/flow/main.go
```diff
@@ -20,46 +20,126 @@
 package main
 
 import (
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flow"
+
+	"github.com/onflow/flow-cli/internal/blocks"
+	"github.com/onflow/flow-cli/internal/emulator"
+
```
psiemens, line 28: If you're using Goland, I recommend enabling `goimports` to automatically sort and organize import statements: https://www.jetbrains.com/help/go/integration-with-go-tools.html#goimports

docs/project-contracts.md
```diff
@@ -33,7 +33,7 @@ In a typical project, a contract has one deployment target per network (e.g. Emu
 
 Deployment targets are defined in the `"deployments"` section of your `flow.json` file.
 
-Targets are grouped by network, where each network is a mapping from target account to contract list. 
+Targets are grouped by the network, where each network is a mapping from target account to contract list.
```
psiemens, line 36:
```suggestion
Targets are grouped by their network, where each network is a mapping from target account to contract list. 
```

cmd/flow/main.go
```diff
@@ -20,46 +20,126 @@
...
+		&command.FormatFlag,
+		"format",
+		"",
+		command.FormatFlag,
+		"Format to show result in",
+	)
+
+	c.PersistentFlags().StringVarP(
+		&command.SaveFlag,
+		"save",
```
psiemens, line 110: Does this need to be a global flag?

psiemens: Will all commands be able to accept the `--save` flag?

cmd/flow/main.go
```diff
@@ -20,46 +20,126 @@
...
+		"Logging level",
+	)
+
+	c.PersistentFlags().BoolVarP(
+		&command.RunEmulatorFlag,
+		"emulator",
+		"",
+		command.RunEmulatorFlag,
+		"Run in-memory emulator",
+	)
```
psiemens, line 130: This flag only applies to commands that would normally use the Access API, right?

docs/project-contracts.md
psiemens, line 36:
```suggestion
Targets are grouped by their network, where each network is a mapping from target account to contract list. 
```

psiemens (commented):
This is a really, really solid refactor! I've got a better idea of the code, but would love to go through it together tomorrow in the time we have booked.

cmd/flow/main.go
sideninja: thank you I will use it

cmd/flow/main.go
sideninja: I would argue any command that has output should support saving (CLI way would even be to not have a flag but to just pipe the result, but since the trend in this CLI was to support some kind of saving I wanted to unify it). If there comes a command we specifically don't want to have --save enabled we can probably blacklist it or something.

cmd/flow/main.go
sideninja: Yes. I think with your previous point we could have a subgroup of commands that have some flags (save, emulator etc). I was anyhow thinking this global command should have a structure on its own probably so we don't have to add all the commands in main but they would just be added to that structure, but I didn't have the time to solve this, maybe we can together discuss this, however not necessary to do it now.

internal/accounts/account-add.go
```diff
@@ -0,0 +1,102 @@
...
+type AccountAddResult struct {
+	*flow.Account
+}
+
+// JSON convert result to JSON
+func (r *AccountAddResult) JSON() interface{} {
+	return r
+}
+
+// String convert result to string
```
psiemens, line 79:
```suggestion
```

internal/accounts/account-add.go
```diff
@@ -0,0 +1,102 @@
...
+	fmt.Fprintf(writer, "Hash Algo\t %s\n", r.Account.DefaultKey().HashAlgo())
+	fmt.Fprintf(writer, "Signature Algo\t %s\n", r.Account.DefaultKey().SigAlgo())
+	fmt.Fprintf(writer, "Key Index\t %d\n", r.Account.DefaultKey().Index())
+
+	writer.Flush()
+
+	return b.String()
+}
+
+// Oneliner show result as one liner grep friendly
```
psiemens, line 94:
```suggestion
```

internal/accounts/contract-add.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flow"
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/psiemens/sconfig"
+	"github.com/spf13/cobra"
+)
+
+type flagsAddContract struct {
+	Account string `default:"emulator-account" flag:"account,a"`
+	Host    string `flag:"host" info:"Flow Access API host address"`
```
psiemens, line 31:
```suggestion
	Host    string `flag:"host,h" info:"Flow Access API host address"`
```

internal/accounts/contract-add.go
```diff
@@ -0,0 +1,70 @@
...
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+
+	account, err := services.Accounts.AddContract(a.flags.Account, args[0], args[1], false)
+	return &AccountResult{
+		Account:  account,
+		showCode: false,
+	}, err
```
psiemens, line 61: I find it's better to be explicit about error paths -- I learned this from a few other Go developers.

```suggestion
	account, err := services.Accounts.AddContract(a.flags.Account, args[0], args[1], false)
	if err != nil {
	  return nil, err
	}
	
	return &AccountResult{
		Account:  account,
		showCode: false,
	}, nil
```

internal/accounts/contract-update.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (c *cmdUpdateContract) Run(
+	cmd *cobra.Command,
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+
+	account, err := services.Accounts.AddContract(c.flags.Account, args[0], args[1], true)
```
psiemens, line 57: Explicit error handling

internal/accounts/contract-remove.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (c *cmdRemoveContract) Run(
+	cmd *cobra.Command,
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+
+	account, err := services.Accounts.RemoveContract(args[0], c.flags.Account)
```
psiemens, line 56: Explicit error handling

internal/accounts/create.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (a *cmdCreate) Run(
+	cmd *cobra.Command,
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+
+	account, err := services.Accounts.Create(
```
psiemens, line 62: Explicit error handling

internal/accounts/contract-add.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flow"
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/psiemens/sconfig"
+	"github.com/spf13/cobra"
+)
+
+type flagsAddContract struct {
+	Account string `default:"emulator-account" flag:"account,a"`
```
psiemens, line 30: Hmm, I'm wondering if this should be `signer` or `account`? 

- `signer` is more consistent with other commands
- `account` makes more senses in this context

internal/accounts/create.go
```diff
@@ -0,0 +1,81 @@
...
+	flags flagsCreate
+}
+
+func NewCreateCmd() command.Command {
+	return &cmdCreate{
+		cmd: &cobra.Command{
+			Use:     "create",
+			Short:   "Create a new account",
+			Example: `flow accounts create --key d651f1931a2...8745`,
+			Aliases: []string{"create"},
```
psiemens, line 50: What's the alias for?

internal/accounts/staking-info.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+func (a *cmdStakingInfo) Run(
+	cmd *cobra.Command,
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+	staking, delegation, err := services.Accounts.StakingInfo(args[0])
```
psiemens, line 40: Explicit error handling

internal/accounts/get.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (a *cmdGet) Run(
+	cmd *cobra.Command,
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+
+	account, err := services.Accounts.Get(args[0])
```
psiemens, line 57: Explicit error handling

internal/blocks/get.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Run block command
+func (s *cmdGet) Run(
+	cmd *cobra.Command,
+	args []string,
+	project *flow.Project,
+	services *services.Services,
+) (command.Result, error) {
+	block, events, collections, err := services.Blocks.GetBlock(args[0], s.flags.Events, s.flags.Verbose)
```
psiemens, line 56: Explicit error handling

internal/accounts/create.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+	}
+}
+
+func (a *cmdCreate) Run(
```
psiemens, line 55:
```suggestion
func (c *cmdCreate) Run(
```

internal/accounts/get.go
```diff
@@ -0,0 +1,70 @@
...
+			Use:     "get <address>",
+			Short:   "Gets an account by address",
+			Aliases: []string{"fetch", "g"},
+			Long:    `Gets an account by address (address, balance, keys, code)`,
+			Args:    cobra.ExactArgs(1),
+		},
+	}
+}
+
+func (a *cmdGet) Run(
```
psiemens, line 50:
```suggestion
func (c *cmdGet) Run(
```

internal/accounts/staking-info.go
```diff
@@ -0,0 +1,85 @@
...
+	return &cmdStakingInfo{
+		cmd: &cobra.Command{
+			Use:   "staking-info <address>",
+			Short: "Get account staking info",
+			Args:  cobra.ExactArgs(1),
+		},
+	}
+}
+
+func (a *cmdStakingInfo) Run(
```
psiemens, line 34:
```suggestion
func (c *cmdStakingInfo) Run(
```

internal/accounts/staking-info.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (a *cmdStakingInfo) GetFlags() *sconfig.Config {
+	return sconfig.New(&a.flags)
+}
+
+func (a *cmdStakingInfo) GetCmd() *cobra.Command {
+	return a.cmd
+}
+
+// AccountResult represent result from all account commands
```
psiemens, line 52:
```StakingResult is the result of the "flow accounts staking-info" command.
// AccountResult represent result from all account commands
```

internal/accounts/staking-info.go
```diff
@@ -0,0 +1,85 @@
...
+type StakingResult struct {
+	staking    *cadence.Value
+	delegation *cadence.Value
```
psiemens, line 55:
```suggestion
	staking    cadence.Value
	delegation cadence.Value
```

`cadence.Value` is an interface, so you shouldn't need a pointer here

internal/accounts/staking-info.go
```diff
@@ -0,0 +1,85 @@
...
+	fmt.Fprintf(writer, "Account Delegation Info:\n")
+	fmt.Fprintf(writer, "%v\n", *r.delegation)
+
+	writer.Flush()
+	return b.String()
+}
+
+// Oneliner show result as one liner grep friendly
+func (r *StakingResult) Oneliner() string {
+	return fmt.Sprintf("")
```
psiemens, line 84:
```suggestion
	return ""
```

Is this meant to be empty?

internal/accounts/get.go
```diff
@@ -0,0 +1,70 @@
...
+	cmd   *cobra.Command
+	flags flagsGet
+}
+
+func NewGetCmd() command.Command {
+	return &cmdGet{
+		cmd: &cobra.Command{
+			Use:     "get <address>",
+			Short:   "Gets an account by address",
+			Aliases: []string{"fetch", "g"},
```
psiemens, line 43: I think it'd be best to focus on getting the new refactor out before adding additional aliases, just in case we decided we don't like them or want to change the command names altogether.

internal/blocks/blocks.go
```diff
@@ -0,0 +1,90 @@
...
+type BlockResult struct {
+	block       *flow.Block
+	events      []client.BlockEvents
+	collections []*flow.Collection
+	verbose     bool
+}
+
+// JSON convert result to JSON
+func (r *BlockResult) JSON() interface{} {
+	return r
```
psiemens, line 49: All the struct fields are private, so I think they'll be ignored by the JSON marshaler

internal/collections/collections.go
```diff
@@ -19,17 +19,41 @@
...
 
-	"github.com/onflow/flow-cli/flow/collections/get"
+	"github.com/onflow/flow-go-sdk"
+	"github.com/spf13/cobra"
 )
 
 var Cmd = &cobra.Command{
 	Use:              "collections",
-	Short:            "Utilities to read collections",
+	Short:            "Utilities to manage collections",
```
psiemens, line 32: I would keep as "read", since collections aren't mutable

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+// addCommand add new command to main cmd
+// and initializes all necessary things as well as take care of errors and output
+// here we can do all boilerplate code that is else copied in each command and make sure
+// we have one place to handle all errors and ensure commands have consistent results
+func Add(c *cobra.Command, command Command) {
+	command.GetCmd().RunE = func(cmd *cobra.Command, args []string) error {
+		// initialize project but ignore error since config can be missing
+		project, _ := flow.LoadProject(flow.ConfigPath)
+
+		gateway, err := createGateway(cmd, project)
```
psiemens, line 62: `gateway` clashes with the `gateway` package. Maybe use `projectGateway`?

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/onflow/flow-cli/pkg/flow/gateway"
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/onflow/flow-cli/pkg/flow/util"
+	"github.com/onflow/flow-go-sdk/client"
+	"github.com/psiemens/sconfig"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+type Command interface {
```
psiemens, line 37: I really like how you moved all this logic into the `command` package! 

After looking at the command implementations, I'd like to propose a slightly different way of structuring this that I think will allow you to use less code


```go
// -----
// command.go 

type RunCommand func (
  *cobra.Command, 
  []string,
  *flow.Project, 
  *services.Services,
) (Result, error)

type Command struct {
  Cmd *cobra.Command
  Config interface{}
  Run RunCommand
}

// -----
// contract-add.go

var AddContractCommand = &command.Command{
  Cmd: &cobra.Command{
    Use:     "add-contract <name> <filename>",
    Short:   "Deploy a new contract to an account",
    Example: `flow accounts add-contract FungibleToken ./FungibleToken.cdc`,
    Args:    cobra.ExactArgs(2),
  },
  Config: &flagsAddContract{},
  Run: func (
    cmd *cobra.Command,
    args []string,
    project *flow.Project,
    services *services.Services,
  ) (command.Result, error) {
    // ...
  }
}

File: internal/command/command.go [hunk 28]
```diff
@@ -0,0 +1,221 @@
...
+				return nil, fmt.Errorf("provided network with name %s doesn't exists in condiguration", NetworkFlag)
+			}
+
+			host = project.Host(NetworkFlag)
+		} else {
+			host = project.Host("emulator")
+		}
+	} else if host == "" {
+		return nil, fmt.Errorf("host must be provided using --host flag or in config by initializing project: flow project init")
+	}
```
psiemens, line 110: Pretty long and complex logic, consider refactoring into a helper function

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// create default grpc client
+	return gateway.NewGrpcGateway(host)
+}
+
+// create logger utility
+func createLogger() util.Logger {
+	// disable logging if we user want a specific format like JSON
+	//(more common they will not want also to have logs)
+	verbose := util.InfoLog
```
psiemens, line 120:
```suggestion
	loglevel := util.InfoLog
```

`verbosity` would work too

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if FormatFlag != "" {
+		verbose = util.NoneLog
+	}
+
+	return util.NewStdoutLogger(verbose)
+}
+
+// outputResult takes care of showing the result
```
psiemens, line 139:
```suggestion
// formatResult formats a result for printing.
```

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+		return "", fmt.Errorf("Missing")
+	}
+
+	if FilterFlag != "" {
+		var jsonResult map[string]interface{}
+		val, _ := json.Marshal(result.JSON())
+		err := json.Unmarshal(val, &jsonResult)
+		if err != nil {
+			return "", err
+		}
```
psiemens, line 151: Is this just to convert the JSON into a filterable form? Makes sense haha! Just wondering if there's an easier way

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+	return util.NewStdoutLogger(verbose)
+}
+
+// outputResult takes care of showing the result
+func formatResult(result Result) (string, error) {
+	if result == nil {
+		return "", fmt.Errorf("Missing")
+	}
+
+	if FilterFlag != "" {
```
psiemens, line 145: Should the filter flag only be accepted if format == json?

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+			return "", err
+		}
+
+		return fmt.Sprintf("%v", jsonResult[FilterFlag]), nil
+	}
+
+	switch FormatFlag {
+	case "json":
+		jsonRes, _ := json.Marshal(result.JSON())
+		return string(jsonRes), nil
```
psiemens, line 159: Could do something like:

`return filterOutput(FilterFlag, jsonRes)`

internal/command/command.go
```diff
@@ -0,0 +1,221 @@
...
+
+// bindFlags bind all the flags needed
+func bindFlags(command Command) {
+	err := command.GetFlags().
+		FromEnvironment(flow.EnvPrefix).
+		BindFlags(command.GetCmd().PersistentFlags()).
+		Parse()
+	if err != nil {
+		fmt.Println(err)
+	}
```
psiemens, line 220: In the version I suggested above, this just becomes:

```go
func bindFlags(command Command) {
	err := sconfig.New(command.Config).
		FromEnvironment(flow.EnvPrefix).
		BindFlags(command.Cmd.PersistentFlags()).
		Parse()
	if err != nil {
		fmt.Println(err)
	}
}

**Review by psiemens — COMMENTED** [review]
Only about halfway through the code so far, but left some initial comments!

File: internal/command/command.go (hunk 32 above)
Comment (sideninja) [reply internal/command/command.go:145]: This is actually here because if we specify filter flag then format doesn't matter anymore since filter returns just the raw value itself. This is meant to be used if we want to just get the value for piping etc... so you create a new account and just want the address you pass `--filter address` and you just get `0x123` so the format here doesn't matter anymore.

File: internal/command/command.go (hunk 31 above)
Comment (sideninja) [reply internal/command/command.go:151]: Not sure what you mean but yes there might be a better way but this seemed to me stupid and simple but works, however if you have a better way I will change :D

File: internal/command/command.go (hunk 33 above)
Comment (sideninja) [reply internal/command/command.go:159]: Here the filter flag doesn't matter anymore as commented above

File: internal/command/command.go (hunk 34 above)
Comment (sideninja) [reply internal/command/command.go:220]: Hmm I'm not sure I understand what you mean by suggested above also Command structure doesn't have a Config attr.

File: internal/accounts/account-add.go [hunk 35]
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/psiemens/sconfig"
+	"github.com/spf13/cobra"
+)
+
+type flagsAdd struct {
+	SigAlgo    string `default:"ECDSA_P256" flag:"algo,a" info:"Signature algorithm"`
+	HashAlgo   string `default:"SHA3_256" flag:"service-hash-algo" info:"Hashing algorithm"`
+	KeyIndex   int    `flag:"index" info:"Account key index"`
+	KeyHex     string `flag:"privateKey" info:"Private key in hex format"`
+	KeyContext string `flag:"context" info:"Projects/<PROJECTID>/locations/<LOCATION>/keyRings/<KEYRINGID>/cryptoKeys/<KEYID>/cryptoKeyVersions/<KEYVERSION>"`
```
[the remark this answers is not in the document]
sideninja: I agree, this sample was there from before, I would like if you can better explain the concept of Context, you have said it will be arbitrary data around the key itself, but still it can't just be completely arbitrary as we need to parse it, so how do we handle the input? With different supported formats as options or?

internal/accounts/contract-add.go
sideninja: This is conflicting with --help,h flag

internal/accounts/contract-add.go
sideninja: came to same conclusion already changed it back :D

internal/accounts/create.go
sideninja: dunno, will remove, this was probably there for a while

internal/accounts/get.go
sideninja: Yes, for sure. This was there before. I will remove

internal/command/command.go
sideninja: Awesome, thanks this is so much better :)

internal/command/command.go
sideninja: Yep already did

internal/accounts/account-add.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/spf13/cobra"
+)
+
+type flagsAdd struct {
+	KeySigAlgo  string `default:"ECDSA_P256" flag:"sig-algo" info:"Signature algorithm"`
+	KeyHashAlgo string `default:"SHA3_256" flag:"hash-algo" info:"Hashing algorithm"`
+	KeyIndex    int    `flag:"index" info:"Account key index"`
+	KeyHex      string `flag:"privateKey" info:"Private key in hex format"`
+	KeyContext  string `flag:"context" info:"Projects/<PROJECTID>/locations/<LOCATION>/keyRings/<KEYRINGID>/cryptoKeys/<KEYID>/cryptoKeyVersions/<KEYVERSION>"`
```
psiemens, line 19:
```suggestion
```

I really think we should take this out (just from the CLI command, it can stay in `pkg/flow`).

It's just that I'm pretty sure we'll want to change this in the future and it'd be nice to avoid a breaking change. Anybody blocked by this can just create the account manually in `flow.json`

internal/accounts/account-add.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/onflow/flow-cli/pkg/flow"
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/spf13/cobra"
+)
+
+type flagsAdd struct {
+	KeySigAlgo  string `default:"ECDSA_P256" flag:"sig-algo" info:"Signature algorithm"`
+	KeyHashAlgo string `default:"SHA3_256" flag:"hash-algo" info:"Hashing algorithm"`
+	KeyIndex    int    `flag:"index" info:"Account key index"`
+	KeyHex      string `flag:"privateKey" info:"Private key in hex format"`
```
psiemens, line 18:
```suggestion
	KeyHex      string `flag:"private-key" info:"Private key in hex format"`
```

Other flags are hyphenated instead of camelcase. Is that true across the whole CLI?

internal/accounts/account-add.go
```diff
@@ -0,0 +1,88 @@
...
+	Flags: addFlags,
+	Run: func(
+		cmd *cobra.Command,
+		args []string,
+		project *flow.Project,
+		services *services.Services,
+	) (command.Result, error) {
+		account, err := services.Accounts.Add(
+			args[0],
+			args[1],
```
psiemens, line 41: It could be nice to assign these args to variables, or add a comment, just so we know what they are (same goes for all `args[X]` across all commands)

internal/blocks/blocks.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// String convert result to string
+func (r *BlockResult) String() string {
+	var b bytes.Buffer
+	writer := tabwriter.NewWriter(&b, 0, 8, 1, '\t', tabwriter.AlignRight)
+
+	fmt.Fprintf(writer, "Block ID\t%s\n", r.block.ID)
+	fmt.Fprintf(writer, "Prent ID\t%s\n", r.block.ParentID)
```
psiemens, line 62:
```suggestion
	fmt.Fprintf(writer, "Parent ID\t%s\n", r.block.ParentID)
```

internal/command/command.go
```diff
@@ -0,0 +1,307 @@
...
+		"Network from configuration file",
+	)
+}
+
+// addCommand add new command to main cmd
+// and initializes all necessary things as well as take care of errors and output
+// here we can do all boilerplate code that is else copied in each command and make sure
+// we have one place to handle all errors and ensure commands have consistent results
+func (c Command) Add(parent *cobra.Command) {
+	c.Cmd.RunE = func(cmd *cobra.Command, args []string) error {
```
psiemens, line 133:
```suggestion
	c.Cmd.Run = func(cmd *cobra.Command, args []string) error {
```

I think you can just use `Run` since no error is returned

internal/command/command.go
```diff
@@ -0,0 +1,307 @@
...
+		NetworkFlag,
+		"Network from configuration file",
+	)
+}
+
+// addCommand add new command to main cmd
+// and initializes all necessary things as well as take care of errors and output
+// here we can do all boilerplate code that is else copied in each command and make sure
+// we have one place to handle all errors and ensure commands have consistent results
+func (c Command) Add(parent *cobra.Command) {
```
psiemens, line 132:
```suggestion
func (c Command) AddToParent(parent *cobra.Command) {
```

The naming confused me a bit, then I realized that it works backwards (child adds itself to parent rather than parent adds child)

internal/command/command.go
```diff
@@ -0,0 +1,307 @@
...
+	FormatFlag      = ""
+	SaveFlag        = ""
+	RunEmulatorFlag = false
+	HostFlag        = ""
+	LogFlag         = "info"
+	NetworkFlag     = ""
+)
+
+// InitFlags init all the global persistent flags
+func InitFlags(cmd cobra.Command) {
```
psiemens, line 62: I just want to make sure that all of these make sense as global flags. Making a note to talk about it

internal/command/command.go
```diff
@@ -0,0 +1,307 @@
...
+
+	bindFlags(c)
+	parent.AddCommand(c.Cmd)
+}
+
+// createGateway creates a gateway to be used, defaults to grpc but can support others
+func createGateway(cmd *cobra.Command, project *flow.Project) (gateway.Gateway, error) {
+	// create in memory emulator client
+	if RunEmulatorFlag {
+		return gateway.NewEmulatorGateway(), nil
```
psiemens, line 167: Is the emulator gateway fully working? I remember you saying that it was still a WIP.

If it's not fully functional, or we don't have a solution to the persistence problem, then I think it's best to just remove `RunEmulatorFlag` for now and not make it an option

internal/keys/generate.go
```diff
@@ -0,0 +1,55 @@
...
+		Short: "Generate a new key-pair",
+	},
+	Flags: &generateFlags,
+	Run: func(
+		cmd *cobra.Command,
+		args []string,
+		project *flow.Project,
+		services *services.Services,
+	) (command.Result, error) {
+		keys, err := services.Keys.Generate(generateFlags.Seed, generateFlags.KeySigAlgo)
```
psiemens, line 47:
```suggestion
		privateKey, err := services.Keys.Generate(generateFlags.Seed, generateFlags.KeySigAlgo)
```

internal/project/init.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/onflow/flow-cli/internal/command"
+
+	"github.com/onflow/flow-cli/pkg/flow"
+
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/spf13/cobra"
+)
+
+type flagsInit struct {
+	ServicePrivateKey  string `flag:"service-priv-key" info:"Service account private key"`
```
psiemens, line 37: I just realized that the `accounts add` command uses `private-key` instead of `priv-key`. Maybe we can standardize?

internal/scripts/execute.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flow"
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/spf13/cobra"
+)
+
+type flagsScripts struct {
+	ArgsJSON string   `default:"" flag:"argsJSON" info:"arguments in JSON-Cadence format"`
```
psiemens, line 31:
```suggestion
	ArgsJSON string   `default:"" flag:"args-json" info:"arguments in JSON-Cadence format"`
```

internal/transactions/send.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flow"
+	"github.com/onflow/flow-cli/pkg/flow/services"
+	"github.com/spf13/cobra"
+)
+
+type flagsSend struct {
+	ArgsJSON string   `default:"" flag:"argsJSON" info:"arguments in JSON-Cadence format"`
```
psiemens, line 31:
```suggestion
	ArgsJSON string   `default:"" flag:"args-json" info:"arguments in JSON-Cadence format"`
```

internal/transactions/send.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+var sendFlags = &flagsSend{}
+
+var SendCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "send <filename>",
+		Short:   "Send a transaction",
+		Args:    cobra.ExactArgs(1),
+		Example: `flow transactions send tx.cdc --args String:"Hello world"`,
```
psiemens, line 45:
```suggestion
		Example: `flow transactions send tx.cdc --arg String:"Hello world"`,
```

pkg/flow/Readme.md
```diff
@@ -0,0 +1,4 @@
+## Library
```
psiemens, line 1: File should be `README.md` to be consistent with other README files

pkg/flow/config/processor_test.go
```diff
@@ -0,0 +1,107 @@
+/*
+* Flow CLI
```
psiemens, line 2: Should this file just be deleted?

pkg/flow/events.go
```diff
@@ -0,0 +1,87 @@
...
+		values[name] = field.String()
+	}
+
+	return Event{
+		Type:   event.Type,
+		Values: values,
+	}
+}
+
+func (e *Events) GetAddress() *flow.Address {
```
psiemens, line 63: Not a blocker: I think there's probably best to not include logic specific to one event inside the `Events` type

pkg/flow/gateway/emulator.go
```diff
@@ -0,0 +1,89 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gateway
+
+import (
+	"fmt"
+
+	flow2 "github.com/onflow/flow-cli/pkg/flow"
```
psiemens, line 24: Haha these name conflicts are awkward. Maybe use `flowcli` instead of `flow2`?

pkg/flow/gateway/gateway.go
```diff
@@ -16,27 +16,24 @@
...
-	if err != nil {
-		Exitf(1, "Failed to connect to host: %s", err)
-	}
-
-	collection, err := flowClient.GetCollection(ctx, collectionID)
-	if err != nil {
-		Exitf(1, "Failed to retrieve collection by ID %s: %s", collectionID, err)
-	}
-	return collection
+type Gateway interface {
```
psiemens, line 28: This is a great package and interface 👏

pkg/flow/project.go
```diff
@@ -167,13 +167,32 @@ func (p *Project) ContractConflictExists(network string) bool {
...
-func (p *Project) EmulatorServiceAccount() *config.Account {
+func (p *Project) EmulatorServiceAccount() (*Account, error) {
 	emulator := p.conf.Emulators.GetDefault()
-	return p.conf.Accounts.GetByName(emulator.ServiceAccount)
+	acc := p.conf.Accounts.GetByName(emulator.ServiceAccount)
+	return AccountFromConfig(*acc)
+}
+
+func (p *Project) SetEmulatorServiceKey(privateKey crypto.PrivateKey) {
+	acc := p.accounts[0]
```
psiemens, line 193:
```suggestion
	acc := p.conf.Accounts.GetByName(emulator.ServiceAccount)
```

I don't think we can assume that the emulator account is always first

pkg/flow/services/accounts.go
```diff
@@ -0,0 +1,458 @@
...
+
+	return account, nil
+}
+
+// StakingInfo gets staking info for the account
+func (a *Accounts) StakingInfo(accountAddress string) (*cadence.Value, *cadence.Value, error) {
+	a.logger.StartProgress(fmt.Sprintf("Fetching info for %s...", accountAddress))
+
+	address := flowsdk.HexToAddress(
+		strings.ReplaceAll(accountAddress, "0x", ""),
```
psiemens, line 169: I don't think this is needed anymore -- the SDK should automatically handle hex prefixes

psiemens (commented):
Amazing stuff, this just keeps getting better. 🎉 

I've reviewed the following:

- `/cmd` ✅ 
- `/internal/*` ✅ 
- 50% of `/pkg/flow/*`

Do you mind moving the `docs` changes and the `tests` changes to 2 additional separate PRs? You can use this branch as a base. That would make it easier to review those files and also would cut down the size of this PR.

internal/accounts/account-add.go
sideninja: good catch, I will make sure all are hyphenated

internal/accounts/account-add.go
sideninja: I'm afraid if I assign them to variable it might happen someone else might see that as a good place to do additional logic on those args, so maybe just comments for now

pkg/flow/gateway/emulator.go
sideninja: yes for sure, I missed it here but what I do is name the sdk `flowsdk`

pkg/flow/services/accounts.go
sideninja: ohh, that got implemented? awesome :))) I will remove this from everywhere

psiemens (commented):
Finished looking through `pkg/flow`! Just need to review `tests` and `docs` next.

In addition to my code comments, what do you think about renaming `pkg/flow` to `pkg/flowcli`? It would clean up the import statements and remove the collision between `flow-go-sdk` and `pkg/flow` (both use `flow` as package name). Happy to keep as-is though, just a suggestion.

pkg/flow/util/logger.go
```diff
@@ -0,0 +1,94 @@
+/*
```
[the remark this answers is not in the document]
sideninja: haha from yesterday:
Screen Shot 2021-03-24 at 9 57 07 AM

pkg/flow/services/keys.go
```diff
@@ -0,0 +1,87 @@
...
+// Keys service handles all interactions for keys
+type Keys struct {
+	gateway gateway.Gateway
+	project *flow.Project
+	logger  util.Logger
+}
+
+// NewTransactions create new transaction service
+func NewKeys(
+	gateway gateway.Gateway,
```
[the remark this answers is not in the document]
sideninja: This is actually persistently present in all services for all the fields and it seem to me like this shadowing make sense since you shouldn't access that package directly as you need to use the instance, however I know shadowing is bad and I made sure elsewhere is not present but I would have to rename all field in all services to ___Client which it looks weird to me, coming from other languages I really don't understand in Go how the instance shouldn't have the same name as "class" :/

pkg/flow/values.go
```diff
@@ -0,0 +1,25 @@
+package flow
+
+import (
+	"github.com/onflow/cadence"
+)
+
+func NewStakingInfoFromValue(value cadence.Value) map[string]interface{} {
```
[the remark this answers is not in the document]
sideninja: I agree but I'm afraid if I start putting a lot of code in commands others will follow that and soon we will again have commands containing logic that won't be possible to reuse elsewhere, I was thinking there might be more use cases to parse cadence values in the future, some other commands might use values package - for example executing cadence scripts might use it when they will try to figure out parameters based on type etc... I really wish to keep commands as minimal as possible and only do execution of service function and showing the result (I understand this function basically is used to show the result but I'm hesitant to move this as what I have stated above)

internal/command/command.go
sideninja: They don't and a I will make them private, they were like this because before they were used globaly which was a bad decision anyway

internal/command/command.go
sideninja: I can make it work, I just wanted to make a boilerplate for implementation and implement it in another PR so it doesn't get overblown as it doesn't belong in refactoring

sideninja: I will remove flag tho

internal/project/init.go
sideninja: Already fixed that yesterday myself

internal/scripts/execute.go
sideninja: Made the same change yesterday :D

internal/transactions/send.go
sideninja: yep :D

pkg/flow/config/processor_test.go
sideninja: Nope, it should be uncommented and fixed - on it

pkg/flow/events.go
sideninja: What if there will be others in future? I think there might, but if not where should I put this?

psiemens (approved):

psiemens (approved):
Feel free to merge this 👍

PR #118, opened 2021-03-22
Display transaction for confirmation

Kay-Zee:
## Description

Adds a confirmation step for signing/sending when the payload is loaded from a file.

@psiemens and I found that we when using the partial transaction functionality, we felt a little blind on what was being signed/sent. It was then suggested that we just print out the tx that was parsed.

This PR prints out the parsed TX in a way that will hopefully assist the user in deciding if they want to sign or not. Also includes an auto confirm option for when using in a setting where we always want to confirm (i.e. in a script, maybe)

The current Transaction display format is a bit ad-hoc. Open to suggestions on how to re-arrange the info if there are any.

______

For contributor use:


sideninja:
Closing in favor of #132

PR #117, opened 2021-03-22
Kan/add payload signing

Kay-Zee:
## Description

- Improves the `sign` sub-function, by allowing it to sign payloads as well (similar to what the send function could do previously)
- Generalize some helper functions that were shared between `sign` and `send` sub commands.
- Split out some of the CLI package functions into their own files, to clean things up a bit
- Changed a few flags to better portray their purpose
  - `payer` -> `payer-address`
  - `partial-tx` -> `payload` I'm not 100% sure about this one, so open to comments here if people preferred the previous naming
- Updated to allow signing as the proposer
  - This was the main functionality that I needed.

______

For contributor use:

- [ ] Targeted PR against `master` branch
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [ ] Code follows the [standards mentioned here](https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#styleguides)
- [ ] Updated relevant documentation
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Added appropriate labels

PR #132, opened 2021-03-30
Signing and sending transactions separated

### LINKED ISSUE #133: Migrate signing transactions to sharedlib CLI
Implementation of signing is built on top of old architecture which changed significantly and needs to be implemented again.

Implementation of functionalities from #77 #117 #118 to the refactored sharedlib CLI.

sideninja:
Closes #133 

Implementation of functionalities from #77 #117 #118 to the refactored sharedlib CLI.

Allows signing a transaction without sending it. Include specifying signers, proposers and authorizers that have to then sign the transaction in a separate step as well. 

Transaction confirmation screen included.


pkg/flowcli/project/transaction.go
```diff
@@ -0,0 +1,243 @@
...
+	authorizerAddress := flow.HexToAddress(address)
+	if authorizerAddress == flow.EmptyAddress {
+		return fmt.Errorf("invalid authorizer address provided %s", address)
+	}
+
+	t.tx.AddAuthorizer(authorizerAddress)
+	return nil
+}
+
+func (t *Transaction) SignerRole(role string) error {
```
psiemens, line 153: What's the purpose of this function?

pkg/flowcli/project/transaction.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+type Transaction struct {
+	signer    *Account
+	role      signerRole
+	proposer  *Account
+	payer     flow.Address
+	tx        *flow.Transaction
+	contracts []templates.Contract
```
psiemens, line 41: Contracts feels out of place here, given that it's only used by transactions that deploy contracts. What's the overall use case for the `Transaction` struct? It feels a bit like a bloated class to me. lt takes care of various actions related to preparing and signing a transaction, but it also has implementation details of specific types of account and contract transactions.

pkg/flowcli/project/transaction.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+
+		t.AddContract(contractName, string(contractSource))
+	}
+
+	return nil
+}
+
+func (t *Transaction) AddContract(name string, source string) {
+	t.contracts = append(t.contracts,
```
psiemens, line 219: Is the `contracts` field needed? It looks like it's only used by the `SetCreateAccount` function. Why not just make that function accept a `contracts` argument?

pkg/flowcli/project/transaction.go
```diff
@@ -0,0 +1,243 @@
...
+	)
+}
+
+func (t *Transaction) SetCreateAccount(keys []*flow.AccountKey) {
+	t.tx = templates.CreateAccount(keys, t.contracts, t.signer.Address())
+}
+
+func (t *Transaction) SetUpdateContract(name string, source string) {
+	t.AddContract(name, source)
+	t.tx = templates.UpdateAccountContract(t.signer.Address(), t.contracts[0])
```
psiemens, line 233: Instead of doing `t.contracts[0]`, I would just do:

```go
contract := templates.Contract{
  Name: name,
  Source: source,
}

t.tx = templates.UpdateAccountContract(t.signer.Address(), contract)

File: pkg/flowcli/project/transaction.go [hunk 5]
```diff
@@ -0,0 +1,243 @@
...
+func (t *Transaction) AddContract(name string, source string) {
+	t.contracts = append(t.contracts,
+		templates.Contract{
+			Name:   name,
+			Source: source,
+		},
+	)
+}
+
+func (t *Transaction) SetCreateAccount(keys []*flow.AccountKey) {
```
psiemens, line 227: I don't think these `Set____` functions should be defined in the `Transaction` struct.

You could do something like this:


```
func NewUpdateAccountContractTransaction(name string, source string) *Transaction {
  // ...
}

File: pkg/flowcli/services/transactions.go [hunk 6]
```diff
@@ -147,29 +197,50 @@ func (t *Transactions) send(
...
 
-// GetStatus of transaction
-func (t *Transactions) GetStatus(
-	transactionID string,
-	waitSeal bool,
+// SendForAddressWithCode send transaction for address and private key specified with code
+func (t *Transactions) SendForAddressWithCode(
+	code []byte,
+	signerAddress string,
+	signerPrivateKey string,
```
psiemens, line 204: Why pass the private key as a string? How would this work with the KMS signer?

pkg/flowcli/project/account.go
```diff
@@ -32,6 +32,18 @@ func (a *Account) SetDefaultKey(key AccountKey) {
...
 }
 
+func (a *Account) ValidateKey() error {
+	if a.DefaultKey().Type() == config.KeyTypeGoogleKMS {
+		resourceID := a.DefaultKey().ToConfig().Context[config.KMSContextField]
+		err := util.GcloudApplicationSignin(resourceID)
+		if err != nil {
+			return err
+		}
+	}
```
psiemens, line 42: Would it make sense to move this validation into the key struct?

```suggestion
	return a.DefaultKey().Validate()
```

pkg/flowcli/gateway/grpc.go
```diff
@@ -61,37 +65,35 @@ func (g *GrpcGateway) GetAccount(address flow.Address) (*flow.Account, error) {
 	return account, nil
 }
 
-// TODO: replace with txsender - much nicer implemented
-// SendTransaction send a transaction to flow
-func (g *GrpcGateway) SendTransaction(tx *flow.Transaction, signer *project.Account) (*flow.Transaction, error) {
-	account, err := g.GetAccount(signer.Address())
+// PrepareTransactionPayload prepares the payload for the transaction from the network
+func (g *GrpcGateway) PrepareTransactionPayload(tx *project.Transaction) (*project.Transaction, error) {
```
psiemens, line 69: I'm not sure if I think it makes sense for `PrepareTransactionPayload` to be implemented in the gateway layer. It contains core logic that shouldn't need to change between gateways.

pkg/flowcli/project/transaction.go
sideninja: Its purpose is to set the signer role which is later used, it's something I just moved from the previous implementation.

pkg/flowcli/project/transaction.go
sideninja: I agree about the contracts. 
Well, the purpose or idea of the Transaction object was that I wouldn't have a Send service function that is long and has an implementation of creating transactions that could be extracted somewhere else. Do you have any other idea how to achieve this?

pkg/flowcli/project/transaction.go
sideninja: Yes I wanted to do this in the first place but then I noticed it needs a signer https://github.com/onflow/flow-cli/blob/sharedlib-signtx/pkg/flowcli/project/transaction.go#L233 which is assigned to tx beforehand, but again thinking about it I can just accept signer in that factory method.

pkg/flowcli/services/transactions.go
sideninja: This implementation is needed by the langauge server and is not supposed to be used with KMS. I will revisit this with Max if there is a better way to implement this but didn't want to block him.

pkg/flowcli/project/account.go
sideninja: 💯

pkg/flowcli/gateway/grpc.go
sideninja: That is true, so should I just fetch the data in the service layer and assign it to the transaction model? I put it here cuz I didn't want to bloat the service layer but yes it makes sense to avoid code duplication.

pkg/flowcli/services/transactions.go
```diff
@@ -53,87 +50,159 @@ func NewTransactions(
...
+	if signerAccount == nil {
+		return nil, fmt.Errorf("signer account: [%s] doesn't exists in configuration", signerName)
 	}
 
-	account := project.AccountFromAddressAndKey(address, privateKey)
+	if payloadFilename != "" && scriptFilename != "" {
+		return nil, fmt.Errorf("can not use both a transaction payload and Cadence code file")
+	} else if payloadFilename == "" && scriptFilename == "" {
+		return nil, fmt.Errorf("provide either a transaction payload or Cadence code file")
+	}
```
psiemens, line 107: Am I correct in assuming that most of the other fields are not required if `payload` is passed? 

For example, the `proposer` arg would be ignored in this case, right?

```
flow transactions sign --payload=mypayload.rlp --proposer=0xabcdef123
```

Do you think it would make sense to throw an error in this case?

pkg/flowcli/services/transactions.go
```diff
@@ -53,87 +50,159 @@ func NewTransactions(
...
 	}
 
-	t.logger.StartProgress("Sending Transaction...")
+	// we are creating a new transaction
+	tx := project.NewTransaction()
 
-	tx := flow.NewTransaction().
-		SetScript(code).
-		AddAuthorizer(signer.Address())
+	err := tx.SetSigner(signerAccount)
```
psiemens, line 130:
```suggestion
	err := tx.SetSigner(signerAccount, role)
```

Any reason not to do this? I'm not sure where else `SetSigner` is called, so apologies if this doesn't work.

pkg/flowcli/services/transactions.go
```diff
@@ -142,34 +211,55 @@ func (t *Transactions) send(
...
 	}
 
-	result, err := t.gateway.GetTransactionResult(tx, waitSeal)
+	tx, err = tx.Sign()
+	if err != nil {
+		return nil, nil, err
+	}
 
-	t.logger.StopProgress("")
+	t.logger.StartProgress("Sending Transaction...")
```
psiemens, line 252:
```suggestion
	t.logger.StartProgress("Sending transaction...")
```

psiemens (commented):
Looking good 👍 

Let me know when it's updated with master and I'll review again

pkg/flowcli/services/transactions.go
sideninja: You sir are correct and yes, fixed :)

pkg/flowcli/services/transactions.go
sideninja: Yes it makes sense, I didn't implement it like this because it progressed to this solution and I missed it

psiemens (approved):
🚢

sideninja:
> 🚢

Is that evergreen?

psiemens:
> Not sure anymore if this emoji means ship it or blocked 😄

LOL

psiemens:
I tested this feature and I think I may have found an edge case. I'm not entirely sure though.

When using the `sign` command, if `payer-address` isn't specified, then the payer defaults to the first signer. However, the first signer defaults to having the role `authorizer`. So by default, the signer is both the authorizer and payer, which is good.

The problem I noticed is that the `sign` command will generate a payload signature by default. This later fails to validate because the payer is required to sign the envelope, not the payload. 

Due to the special case below (from the docs), if the signer is both the authorizer and payer, they should be signing the envelope only:

> Special case: if an account is both the payer and either a proposer or authorizer, it is only required to sign the envelope.

```sh
# specifies account-1 as payer, but signs as authorizer
flow transactions sign --signer=account-1 myTx.cdc -x payload --save=myTx.rlp
```

PR #143, opened 2021-04-05
Feature/sharedlib

### LINKED ISSUE #80: Shared library for CLI command logic

sideninja:
## Description

Closes #80 

Refactor of CLI to introduce the concept of a shared library. Shared library is meant to be used by other software to leverage the functionality of the CLI and avoid code duplication. 

Refactor also improves code architecture so we avoid having an implementation of domain logic in CLI commands and have them separated in service layer (services related to CLI functionality), gateway layer (network access functionality).

______

For contributor use:


psiemens (approved):
🚀 🔥

PR #147, opened 2021-04-05
Sharedlib camelcase

sideninja:
## Description

Change values from JSON result to camel case.

Camel casing result values in JSON format causes a problem with the user knowing which values can now use in `--filter` flag as they are different from the normal output. 

In any case, it was bad UX to not show to the user possible filter values.

Handle invalid filter values and offer possible valid filter values:

```
❌ Result: value for filter: 'addres' doesn't exist, possible values to filter by: [address balance contracts keys]
```

______

For contributor use:

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -219,18 +219,16 @@ func createLogger(logFlag string, formatFlag string) output.Logger {
 // formatResult formats a result for printing.
 func formatResult(result Result, filterFlag string, formatFlag string) (string, error) {
 	if result == nil {
-		return "", fmt.Errorf("Missing")
+		return "", fmt.Errorf("missing result")
 	}
 
 	if filterFlag != "" {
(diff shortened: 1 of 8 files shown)


internal/command/command.go
```diff
@@ -260,6 +258,29 @@ func outputResult(result string, saveFlag string) error {
...
+
+	possibleFilters := make([]string, 0)
+	for key, _ := range jsonResult {
+		possibleFilters = append(possibleFilters, fmt.Sprintf("%s", key))
+	}
+
+	value := jsonResult[filter]
+
+	if value == nil {
+		return nil, fmt.Errorf("value for filter: '%s' doesn't exists, possible values to filter by: %s", filter, possibleFilters)
```
psiemens, line 278: Nice!

psiemens (approved):
🎉 nice!

PR #148, opened 2021-04-05
Sharedlib flow init

sideninja:
## Description

Move init command back to the top level command and remove `project init` command. Create a placeholder for config command as it will be used in the future. 

______

For contributor use:


internal/config/config.go
```diff
@@ -0,0 +1,45 @@
...
+	"text/tabwriter"
+
+	"github.com/onflow/flow-cli/pkg/flowcli/project"
+	"github.com/onflow/flow-cli/pkg/flowcli/util"
+)
+
+// todo implement top level command config
+
+// InitResult result structure
+type InitResult struct {
```
psiemens, line 15: If this struct is only used by the `init` command, maybe just move it to `init.go`

psiemens (commented):
Very nice!

internal/config/config.go
sideninja: Yes, you are right, I wanted to leave it there to start the notion of having config as top level command one day but maybe is better to not do that until there is more code.

PR #161, opened 2021-04-07
Resolve imports for transactions and scripts

### LINKED ISSUE #128: Resolving addresses in scripts and transactions
Scripts and transactions should be able to resolve imports based on the use of the project deploy command.

Transactions that `import` a contract from the address should also benefit from importing local contract and after deployment for that to be resolved to the deployed address. 

Import should be resolved directly from the configuration.

sideninja:
Closes #128 

## Description
Scripts and transactions should be able to resolve imports based on the use of the project deploy command.

______

For contributor use:


pkg/flowcli/contracts/resolver.go
```diff
@@ -0,0 +1,104 @@
...
+	return sourceTarget
+}
+
+// ImportExists checks if there is an import statement present in Cadence code
+func (r *Resolver) ImportExists() bool {
+	return len(r.parseImports()) > 0
+}
+
+// parseImports returns all imports from Cadence code as an array
+func (r *Resolver) parseImports() []string {
```
psiemens, line 93: So this function only returns non-address imports, right? Maybe rename it or adjust the comment to clarify

pkg/flowcli/contracts/resolver.go
```diff
@@ -0,0 +1,104 @@
...
+	imports := make([]string, 0)
+
+	for _, importDeclaration := range r.program.ImportDeclarations() {
+		location, ok := importDeclaration.Location.(common.StringLocation)
+		if ok && flow.HexToAddress(location.String()) == flow.EmptyAddress {
```
psiemens, line 98: Is the `flow.HexToAddress` required? I'm pretty sure address imports will always be typed as `common.AddressLocation` instead of `StringLocation`.

pkg/flowcli/contracts/resolver.go
```diff
@@ -0,0 +1,104 @@
...
+
+	for source, target := range aliases {
+		sourceTarget[path.Clean(source)] = flow.HexToAddress(target).String()
+	}
+
+	return sourceTarget
+}
+
+// ImportExists checks if there is an import statement present in Cadence code
+func (r *Resolver) ImportExists() bool {
```
psiemens, line 88:
```suggestion
func (r *Resolver) HasStringImports() bool {
```

When I first saw this function being used, I was worried it was counting all imports including addresses

internal/command/command.go
```diff
@@ -193,19 +193,26 @@ func createGateway(host string) (gateway.Gateway, error) {
...
+
+	// host flag has highest priority
+	if hostFlag != "" {
+		return hostFlag, nil
+	}
+	// network flag with project initialized is next
 	if networkFlag != "" && proj != nil {
 		check := proj.NetworkByName(networkFlag)
 		if check == nil {
 			return "", fmt.Errorf("provided network with name %s doesn't exists in condiguration", networkFlag)
```
psiemens, line 209:
```suggestion
			return "", fmt.Errorf("network with name %s does not exist in configuration", networkFlag)
```

pkg/flowcli/arguments.go
```diff
@@ -75,6 +75,11 @@ func ParseArgumentsCommaSplit(input []string) ([]cadence.Value, error) {
 			)
 		}
 
+		// if we are passing address check and handle without 0x prefix as else it wont work and error is not descriptive
+		if argInput[0] == "Address" && !strings.Contains(argInput[1], "0x") {
+			argInput[1] = fmt.Sprintf("0x%s", argInput[1])
+		}
```
psiemens, line 81: This logic is a bit hard to follow. Maybe move to a helper like `sanitizeAddressArg(argInput)`

pkg/flowcli/services/transactions.go
```diff
@@ -132,6 +137,28 @@ func (t *Transactions) send(
...
 
+	resolver, err := contracts.NewResolver(code)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if resolver.ImportExists() {
+		if network == "" {
+			return nil, nil, fmt.Errorf("missing network, specify which network to use to resolve imports in transaction code")
+		} else if codePath == "" { // when used as lib with code we don't support imports
```
psiemens, line 148: I would just make this a separate `if` statement, no need to have an `else` chain if each block has a `return` statement

pkg/flowcli/services/scripts.go
```diff
@@ -49,25 +52,49 @@ func NewScripts(
...
+	}
+
+	if resolver.ImportExists() {
+		if s.project == nil {
+			return nil, fmt.Errorf("missing configuration, initialize it: flow init")
+		} else if network == "" {
+			return nil, fmt.Errorf("missing network, specify which network to use to resolve imports in script code")
+		} else if scriptPath == "" { // when used as lib with code we don't support imports
+			return nil, fmt.Errorf("resolving imports in scripts not supported")
+		}
```
psiemens, line 87: I would just make each a separate `if` statement, no need to have an `else` chain if each block has a `return` statement. Feels a bit easier to read IMO

psiemens (approved):
Very nice!

pkg/flowcli/services/scripts.go
sideninja: I was trying to save up lines but you are right I will change

pkg/flowcli/contracts/resolver.go
sideninja: Oh yes for sure. Didn't see the type.

PR #165, opened 2021-04-08
Docs: fixed command syntax

sideninja:
## Description
Unified command declaration syntax.
______

For contributor use:

### DIFF
--- a/docs/account-add-contract.md
+++ b/docs/account-add-contract.md
@@ -5,7 +5,9 @@ sidebar_title: Deploy a Contract
 
 Deploy a new contract to a Flow account using the Flow CLI.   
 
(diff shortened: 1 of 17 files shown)


turbolent (commented):
What is the reason for switching away from code blocks with syntax highlighting? Maybe keep the current ones, and switch the inline ones to blocks?

sideninja:
> What is the reason for switching away from code blocks with syntax highlighting? Maybe keep the current ones, and switch the inline ones to blocks?

I thought it looks nicer, but the point is that it wasn't consistent, so I'm fine with either as long as they are consistent. I will fix to block if you think is better.

psiemens (approved):

PR #169, opened 2021-04-08
Extract out the parsing of the arguments one layer

bjartek:
Why:

* I want to reimplement go-with-the-flow on top of flow-cli
* go-with-the-flow has a builder to add Arguments and already has an array of cadnece.Value
* with the new SendWithArguments method it can be used directly

This change addresses the need by:

* go-with-the-flow to reimplement on top of flow-cli

Closes #???

## Description

______

For contributor use:

### DIFF
--- a/pkg/flowcli/services/transactions.go
+++ b/pkg/flowcli/services/transactions.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/onflow/cadence"
 	"github.com/onflow/flow-cli/pkg/flowcli/contracts"
 
 	"github.com/onflow/flow-go-sdk"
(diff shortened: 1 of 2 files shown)


bjartek:
Really like the work that has been done here! Hope that this small change can be accepted since it would make it easier for me to use.

sideninja:
Heads up: refactor of the service layer is planned to support "flow native" types, so instead of passing string value of address or cadence arg we will accept `flow.address` or `cadence.Value`.

bjartek:
Well we could always include the TransactionBuilder/ScriptBuilder from go-with-the-flow inside this lib. However it is not strictly a part of it. https://github.com/bjartek/go-with-the-flow/blob/master/gwtf/transaction.go i can easily refactor that on top of the current shared lib and just maintain that in go-with-the-flow. However I feel that the TransactionBuilder is a really nice pattern to use in testing, and flow-cli service layer is not a really really nice testing api it looks like with the new gateway abstraction that can be mocked, can be an in mem chain or can be backed by GRPC.

sideninja:
> Well we could always include the TransactionBuilder/ScriptBuilder from go-with-the-flow inside this lib. However it is not strictly a part of it. https://github.com/bjartek/go-with-the-flow/blob/master/gwtf/transaction.go i can easily refactor that on top of the current shared lib and just maintain that in go-with-the-flow. However I feel that the TransactionBuilder is a really nice pattern to use in testing, and flow-cli service layer is not a really really nice testing api it looks like with the new gateway abstraction that can be mocked, can be an in mem chain or can be backed by GRPC.

Yeah I agree, I like builders too. I implemented somethings similar in latest branch here: https://github.com/onflow/flow-cli/blob/sharedlib-signtx/pkg/flowcli/project/transaction.go 
I think this should be part of the bigger refactor I plan to do in the next days.

bjartek:
Awesome!!!

PR #171, opened 2021-04-09
Version check

### LINKED ISSUE #65: Notify when new version is available for download.
### Issue To Be Solved

It's not always easy to know when an upgrade has been released for the CLI. There are a lot of new features planned and it would be great if CLI users were informed about available upgrades as soon as they are available.

- CLI can notify users by printing a message in the terminal.
- Message could include platform-specific upgrade instructions 

### (Optional): Context

We may want to leave the actual upgrade process up to users and specific package managers, instead of having code in the CLI to perform the upgrade.

sideninja:
Closes #65 

## Description
Notifies when a new version is available. If the network is not available or there is a network error ignore the check. Caching does not provide many benefits.

______

For contributor use:

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -358,7 +358,7 @@ func checkVersion(logger output.Logger) {
 
 	if string(latestVersion) != build.Semver() {
 		logger.Info(fmt.Sprintf(
-			"\n⚠️  Version Warning: New version %s of Flow CLI is available.\n"+
+			"\n⚠️  Version warning: a new version of Flow CLI is available (%s).\n"+
 				"Follow the Flow CLI installation guide for instructions on how to install or upgrade the CLI: https://docs.onflow.org/flow-cli/install",
 			strings.ReplaceAll(string(latestVersion), "\n", ""),
 		))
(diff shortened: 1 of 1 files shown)


janezpodhostnik (approved):

internal/command/command.go
```diff
@@ -340,6 +346,25 @@ func handleError(description string, err error) {
...
+	}
+
+	defer resp.Body.Close()
+	latestVersion, _ := ioutil.ReadAll(resp.Body)
+
+	if string(latestVersion) != build.Semver() {
+		logger.Info(fmt.Sprintf(
+			"\n⚠️  Version Warning: New version %s of Flow CLI is available.\n"+
+				"Follow the Flow CLI installation guide for instructions on how to install or upgrade the CLI: https://docs.onflow.org/flow-cli/install",
+			strings.ReplaceAll(string(latestVersion), "\n", ""),
```
psiemens, line 363: I would sanitize this string before the `if` statement, just in case. You can also use `strings. TrimSpace` for this, I think.

```go
body, _ := ioutil.ReadAll(resp.Body)

latestVersion := strings.TrimSpace(string(body))

if latestVersion != build.Semver() {

psiemens (approved):

PR #203, opened 2021-04-20
Command autocomplete

### LINKED ISSUE #102: Command autocompletions
### Issue To Be Solved

CLI should support autocompletion of commands by using tab key.

### Suggest A Solution

Cobra has a good support for this https://github.com/spf13/cobra#generating-shell-completions

sideninja:
Closes #102 

## Description
Autocompletion support for CLI commands.

### DIFF
--- a/internal/completion/completion.go
+++ b/internal/completion/completion.go
@@ -1,0 +1,82 @@
+/*
+ * Flow CLI
+ *
+ * Copyright 2019-2021 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package completion
+
+import (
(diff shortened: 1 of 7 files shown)


sideninja:
💥  ISSUE: with autocompletes file completion (./file tab tab) doesn't work as it should. Explore

internal/completion/completion.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+var Cmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate completion script",
+	Long: `To load completions:
+
+Bash:
+
+  $ source <(yourprogram completion bash)
```
psiemens, line 34:
```suggestion
  $ source <(flow completion bash)
```

These should all be `flow`, right?

internal/completion/completion.go
```diff
@@ -0,0 +1,82 @@
...
+	Args:                  cobra.ExactValidArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		switch args[0] {
+		case "bash":
+			cmd.Root().GenBashCompletion(os.Stdout)
+		case "zsh":
+			cmd.Root().GenZshCompletion(os.Stdout)
+		case "powershell":
+			cmd.Root().GenPowerShellCompletion(os.Stdout)
+		}
```
psiemens, line 80: `fish` is a valid option but not covered in the switch case

sideninja: yes, it's still a bit wip, I moved on from this to another PR for now since I need to figure more things out.

sideninja:
💥 ISSUE: some commands should not have global flags under autocomplete as they don't require them - example `flow keys generate` doesn't need global host flag.

PR #205, opened 2021-04-20
Default network without config

### LINKED ISSUE #135: Default configuration for subset of commands
### Issue To Be Solved
Currently, all commands in CLI require configuration. All commands that currently require configuration just to get the network access point address should work without the configuration and get that address from the variable in the codebase.

### Suggest A Solution
We should define all the networks within the CLI codebase (this also relates to issue #125). If no configuration is present we use these values as a last resort solution to get the network address and thus not require configuration to be created with flow init. 

This way UX will be improved because a simple use case of using CLI tool is to just maybe get an account or query a block and with this solution implemented it won't be necessary to init a configuration first, which is also not a default thought process of a user, no one thinks "ohh I need to first run this command in order to execute what I want".

Priority of network value resolution will be:
- network flag within the configuration
- host flag
- flow configuration default
- codebase value default

The default value for the network should probably be emulator network, but then again that might not be the best decision as it again preconditions the emulator is running which is anyhow dependent on the configuration being initialized. It might make more sense to default it to testnet (as that is dev home).

sideninja:
Closes #135 

## Description
Allow resolving networks (emulator, testnet, mainnet) even if there is no configuration present, but be sure to first resolve config if present.

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -207,6 +207,12 @@ func createGateway(host string) (gateway.Gateway, error) {
 }
 
 // resolveHost from the flags provided
+//
+// Resolve the network host in the following order:
+// 1. if host flag is provided resolve to that host
+// 2. if conf is initialized return host by network flag
+// 3. if conf is not initialized and network flag is provided resolve to coded value for that network
+// 4. default to emulator network
 func resolveHost(proj *project.Project, hostFlag string, networkFlag string) (string, error) {
 	// don't allow both network and host flag as the host might be different
 	if networkFlag != config.DefaultEmulatorNetwork().Name && hostFlag != "" {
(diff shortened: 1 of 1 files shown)


internal/command/command.go
```diff
@@ -218,16 +224,21 @@ func resolveHost(proj *project.Project, hostFlag string, networkFlag string) (st
...
 	}
-	// default to emulator host
-	return config.DefaultEmulatorNetwork().Host, nil
+
+	networks := config.DefaultNetworks()
+	if networks.GetByName(networkFlag) != nil {
+		return networks.GetByName(networkFlag).Host, nil
+	} else {
+		return "", fmt.Errorf("invalid network with name %s", networkFlag)
+	}
```
psiemens, line 241:
```suggestion
	networks := config.DefaultNetworks()
	network := networks.GetByName(networkFlag)
	
	if network != nil {
		return network.Host, nil
	}
	
	return "", fmt.Errorf("invalid network with name %s", networkFlag)
```

psiemens (approved):

Issue #60, opened 2021-02-16
Better validation of configuration


sideninja:
There should be a JSON schema validation for configuration properties. We need to validate and provide the best possible feedback to wrong values/properties to users. I think JSON schema can be a good way to achieve this, but it will be different for each format (yaml...). I believe the first step should be improving validation on the abstract configuration structures and second step could be validation of configuration files since the first step provides the feedback the value is not correct but it doesn't say where the value error is located or if the value is missing where that value should be defined.

uhbif19:
+1 

Also, there are two formats of `flow.json` with no clear description of their differences.

sideninja:
> +1
> 
> Also, there are two formats of `flow.json` with no clear description of their differences.

There is an old `flow.json` format that if you try using with updated CLI it shows a warning. Is that what you have in mind?

uhbif19:
@sideninja I am not sure which one of `flow init` or `flow project init` is old one. If I recall correctly, it may sometimes require different formats for different tasks or fail without warning on using wrong format.

sideninja:
Oh yeah the `flow project init` is getting deprecated, there is warning in the new version - I suggest you update it.

PR #218, opened 2021-04-23
Configuration manipulation command

sideninja:
Closes #218 #60 

Command for adding resources to the configuration with validation of the inputs. 
Another command for removing resources from the configuration.


docs/manage-configuration.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Manage Flow Configuration
+sidebar_title: Manage Configuration
+description: How to manage Flow Configuration using CLI
```
psiemens, line 4:
```suggestion
description: How to configure the Flow CLI
```

internal/config/add.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var AddCmd = &cobra.Command{
+	Use:              "add <account|contract|deployment|network>",
+	Short:            "Add resource to configuration",
+	Example:          "flow config add account",
+	ValidArgs:        []string{"account", "contract", "deployment", "network"},
```
psiemens, line 29: Is `ValidArgs` needed if all of these are subcommands?

internal/config/add-account.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/onflow/flow-cli/pkg/flowcli/output"
+	"github.com/onflow/flow-cli/pkg/flowcli/project"
+	"github.com/onflow/flow-cli/pkg/flowcli/services"
+)
+
+type flagsAddAccount struct {
+	Name     string `flag:"name" info:"Name for the account"`
+	Address  string `flag:"address" info:"Account address"`
+	KeyIndex string `default:"0" flag:"key-index" info:"Account key index"`
+	SigAlgo  string `default:"ECDSA_P256" flag:"sig-algo" info:"Account key signature algorithm"`
```
psiemens, line 37:
```suggestion
	SigAlgo  string `default:"ECDSA_P256" flag:"sig-algo" info:"Signature algorithm of this account key"`
```

internal/config/add-account.go
```diff
@@ -0,0 +1,133 @@
...
+	HashAlgo string `default:"SHA3_256" flag:"hash-algo" info:"Account hash used for the digest"`
```
psiemens, line 38:
```suggestion
	HashAlgo string `default:"SHA3_256" flag:"hash-algo" info:"Hash algorithm to pair with this account key"`
```

internal/config/add-account.go
```diff
@@ -0,0 +1,133 @@
...
+	Key      string `flag:"key" info:"Account private key"`
```
psiemens, line 39: I think it might be better to be explicit here and say `private-key`, given that we use `key` on other commands to refer to a public key.

internal/config/add.go
sideninja: not anymore, before it was, forgot to remove

PR #239, opened 2021-05-07
Docs update to align code examples with v0.19.0

gregsantos:
## Description

- Updated code examples to match v0.19.0 specs
- Replaced --signer and --argsJSON flags with --authorizer and --args-json


psiemens (approved):
Thank you!

PR #241, opened 2021-05-10
Add gas limit flag to `transactions send` & `transactions build`

psiemens:
## Description

This PR adds a flag that can be used to override the default gas limit, which is useful for more complex transactions that use additional gas.

______

For contributor use:


internal/transactions/build.go
```diff
@@ -52,12 +53,15 @@ var BuildCommand = &command.Command{
 
 		globalFlags.Filter = "payload"
```
sideninja, line 54: Not sure when this was changed but we at least need to update the docs if this will be the new default value and I believe we should have a value that can be used to display the whole transaction without filters, the reason behind this is that when you build a transaction you might want to confirm all the values and it is hard to see that from the payload

sideninja (changes requested):
Tests need to be updated with the new `gasLimit` parameter.
Documentation needs to be updated too.

internal/transactions/build.go
psiemens: Ah was this not intentional? https://github.com/onflow/flow-cli/commit/498221dddc2d77d65dd1ab40cbea0c43a599aac0

psiemens:
@sideninja Updated docs and tests, removed filter override here: https://github.com/onflow/flow-cli/pull/241/commits/03106c2a9045e5fc5e3ff905f058d5f785deaa05

sideninja (approved):
🚀  looks great

PR #253, opened 2021-05-17
Include and exclude flags support

### LINKED ISSUE #249: Result optimization
### Issue To Be Solved
The results for some commands can be too verbose and hence drown the user with irrelevant or not needed data.

One such example is transaction result with payload value. Payload value is not commonly used but it can take more screen than the percentage of it being used (>50%). 

Introduce hidden fields that can be "expanded" on the request. The result can provide some values that are hidden (just provide the field name) and allow the user to specify the field to be expanded via flag. The flag can be `--include`.

Example:

```
Status          ✅ SEALED
ID              1d806939b326fcb9b0cc1aadc703db77476c285b25ce4dc471e0d535e97749b7
Payer           18eb4ee6b3c026d2
Authorizers     [6a707c4ad375ad50]

Proposal Key:   
    Address     6a707c4ad375ad50
    Index       6
    Sequence    11

Payload Signature 0:
    Address     6a707c4ad375ad50

Payload Signature 1:
    Address     6a707c4ad375ad50

Envelope Signature 0:
    Address     18eb4ee6b3c026d2
   (hidden, use: --include signatures)

Events: (hidden, use: --include events)

Arguments       No arguments

Code (hidden, use: --include code)

Payload (hidden, use: --include payload)
```

Another flag that could be introduced in the same nature is: `--exclude` flag which explicitly removes a field from the result.

sideninja:
Closes #249 

Results with a high level of verbosity support usage of `--include` and `--exclude` flags to further define the output content

### DIFF
--- a/internal/command/result.go
+++ b/internal/command/result.go
@@ -19,7 +19,7 @@
 package command
 
 import (
-	"github.com/onflow/flow-cli/pkg/flowcli/util"
+	"strings"
 )
 
 type Result interface {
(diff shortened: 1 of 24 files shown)


internal/command/result.go
```diff
@@ -18,8 +18,22 @@
...
+
 type Result interface {
 	String() string
 	Oneliner() string
 	JSON() interface{}
 }
+
+// FieldIncluded checks if field should be included in the result by checking if included flags contains field
+func FieldIncluded(included []string, field string) bool {
+	return util.ContainsStringInsensitive(included, field)
```
janezpodhostnik, line 33: `FieldIncluded` and `FieldExcluded` do the same thing, which is forward to `util.ContainsStringInsensitive(included, field)`. Should these two functions just be inlined?

janezpodhostnik (approved):
LGTM. Had one small comment.

internal/command/result.go
sideninja: Yes, I agree. Will push an update.

PR #256, opened 2021-05-17
Fix import detection

### LINKED ISSUE #252: Cannot import built-in contracts
### Problem

An error occurs when executing a script that imports a built-in contract (`Crypto` contract) with Flow CLI command.

### Steps to Reproduce

1. Create the following script

test.cdc
```
import Crypto

pub fun main() {
    log("test")
}
```

2. Execute the following command in the Flow CLI of v0.18.0 or later version

```sh
flow scripts execute ./test.cdc
```

The following error occurs
```
❌  Command Error: import Crypto could not be resolved from the configuration
```

### Context

It does not occur with v0.17.0, but with v0.18.0 and later. I am using macOS.
This may be due to `Resolve Contract Imports in Scripts and Transactions` feature.

sideninja:
Closes #252 

## Description
Import detection was fixed so it only detects file imports.

### DIFF
--- a/pkg/flowcli/contracts/resolver.go
+++ b/pkg/flowcli/contracts/resolver.go
@@ -112,8 +112,9 @@ func (r *Resolver) getFileImports() []string {
 	imports := make([]string, 0)
 
 	for _, importDeclaration := range r.program.ImportDeclarations() {
-		_, ok := importDeclaration.Location.(common.AddressLocation)
-		if !ok {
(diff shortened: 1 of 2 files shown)


janezpodhostnik (approved):

janezpodhostnik:
Would it be possible to write a test covering this?

turbolent (approved):

sideninja:
> Would it be possible to write a test covering this?

Great idea, totally forgot.

PR #264, opened 2021-05-21
QA remove lower case match

sideninja:
Remove to lower case when accessing map keys as some have multi-word keys that then don't work. Example `parentId` key can not be accesses as it is converted to `parentid`

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -331,7 +331,7 @@ func filterResultValue(result Result, filter string) (interface{}, error) {
 		possibleFilters = append(possibleFilters, key)
 	}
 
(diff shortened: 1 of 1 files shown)


janezpodhostnik (approved):

janezpodhostnik:
You could also try both. 
if `value := jsonResult[filter]` returns nothing
try `value := jsonResult[strings.ToLower(filter)]` as well

sideninja:
> You could also try both.
> if `value := jsonResult[filter]` returns nothing
> try `value := jsonResult[strings.ToLower(filter)]` as well

ohhh right, smart

PR #266, opened 2021-05-24
Revert "Update to Cadence v0.16.0 and Emulator v0.20.1"

sideninja:
Reverts onflow/flow-cli#265 - due to bug in flow-go fixed with PR https://github.com/onflow/flow-go/pull/748


turbolent:
Why was this reverted?

sideninja:
> Why was this reverted?

Because of the bug in the flow-go that causes issues with compilation: https://github.com/onflow/flow-go/pull/748
If this is needed to be released I can include it after that gets fixed or do you have another suggestion?

PR #269, opened 2021-05-26
Decode PEM public key

### LINKED ISSUE #262: Add ability to decode PEM files to use for keys
### Issue To Be Solved

A common format used by users is to store their keys in PEM files. For example someone can store their keys using Gcloud and to retrieve the public key (https://cloud.google.com/kms/docs/retrieve-public-key) they get back a PEM file. When this is given to us, we can easily use this to decode their public key and create their account on mainnet.

Something like below:
```
    "keys": [
      {
        "type": "pem",
        "index": 0,
        "signatureAlgorithm": "ECDSA_P256",
        "hashAlgorithm": "SHA3_256",
        "context": {
          "publicKeyPath": "/config/keys.pem"
        }
      },
]
```
And once the account is printed, it should display the decoded public key.

### (Optional): Context
A highly requested feature from partners looking to launch and our team responsible for deploying them.

sideninja:
Closes #262 

## Description
Support decoding of PEM keys.

### DIFF
--- a/docs/decode-keys.md
+++ b/docs/decode-keys.md
@@ -12,17 +12,50 @@ flow keys decode <rlp|pem> <encoded public key>
 
 ## Example Usage
 
+### Decode RLP Encoded Public Key
 ```shell
 > flow keys decode rlp f847b84084d716c14b051ad6b001624f738f5d302636e6b07cc75e4530af7776a4368a2b586dbefc0564ee28384c2696f178cbed52e62811bcc9ecb59568c996d342db2402038203e8
 
-Public Key 		 84d716c14b051ad6b001624f738f5d302636e6b07cc75e4530af7776a4368a2b586dbefc0564ee28384c2696f178cbed52e62811bcc9ecb59568c996d342db24 
+Public Key 		 84d716c1...bcc9ecb59568c996d342db24 
 Signature algorithm 	 ECDSA_P256
 Hash algorithm 		 SHA3_256
 Weight 			 1000
 Revoked 		 false
 ```
 
+### Decode PEM Encoded Public Key From File
+```shell
+> flow keys decode pem --from-file key.pem
+
+Public Key 		 d479b3c...c4615360039a6660a366a95f 
+Signature algorithm 	 ECDSA_P256
+Hash algorithm 		 UNKNOWN
+Revoked 		 false
+
+```
+
+## Arguments
+
+### Encoding
+- Valid inputs: `rlp`, `pem` 
+
+First argument specifies a valid encoding of the public key provided.
+
+### Optional: Public Key
+- Name: `encoded public key`
+- Valid inputs: valid encoded key content
+
+Optional second argument provides content of the encoded public key. 
+If this argument is omitted the `--from-file` must be used instead.  
+
 ## Flags
+
+### From File
+
+- Flag: `--from-file`
+- Valid inputs: valid filepath
+
+Provide file with the encoded public key. 
 
 ### Filter
 
--- a/docs/decode-keys.md
+++ b/docs/decode-keys.md
@@ -1,19 +1,19 @@
 ---
-title: Decode Account Key with the Flow CLI
-sidebar_title: Decode Account Keys
-description: How to decode rlp encoded key pair from the command line
+title: Decode Account Keys with the Flow CLI
+sidebar_title: Decode Public Keys
+description: How to decode encoded public keys from the command line
 ---
 
-The Flow CLI provides a command to decode RLP encoded account key.
+The Flow CLI provides a command to decode encoded public account keys.
 
 ```shell
-flow keys decode <rlp encoded account key>
+flow keys decode <rlp|pem> <encoded public key>
 ```
 
 ## Example Usage
 
 ```shell
-> flow keys decode f847b84084d716c14b051ad6b001624f738f5d302636e6b07cc75e4530af7776a4368a2b586dbefc0564ee28384c2696f178cbed52e62811bcc9ecb59568c996d342db2402038203e8
+> flow keys decode rlp f847b84084d716c14b051ad6b001624f738f5d302636e6b07cc75e4530af7776a4368a2b586dbefc0564ee28384c2696f178cbed52e62811bcc9ecb59568c996d342db2402038203e8
 
 Public Key 		 84d716c14b051ad6b001624f738f5d302636e6b07cc75e4530af7776a4368a2b586dbefc0564ee28384c2696f178cbed52e62811bcc9ecb59568c996d342db24 
 Signature algorithm 	 ECDSA_P256
--- a/docs/decode-keys.md
+++ b/docs/decode-keys.md
@@ -1,5 +1,5 @@
 ---
-title: Decode Account Keys with the Flow CLI
+title: Decode Public Keys with the Flow CLI
 sidebar_title: Decode Public Keys
 description: How to decode Flow public keys from the command line
 ---
--- a/internal/keys/decode.go
+++ b/internal/keys/decode.go
@@ -31,10 +31,11 @@ var decodeFlags = flagsDecode{}
 
 var DecodeCommand = &command.Command{
 	Cmd: &cobra.Command{
-		Use:     "decode <rlp encoded account key>",
-		Short:   "Decode a rlp encoded account key",
-		Args:    cobra.ExactArgs(1),
-		Example: "flow keys decode f847b8408...2402038203e8",
+		Use:       "decode <rlp|pem> <encoded public key>",
+		Short:     "Decode an encoded public key",
+		Args:      cobra.ExactArgs(2),
+		ValidArgs: []string{"rlp", "pem"},
+		Example:   "flow keys decode rlp f847b8408...2402038203e8",
 	},
 	Flags: &decodeFlags,
 	Run: func(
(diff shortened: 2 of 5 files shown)


Kay-Zee:
was thinking to also be able to decode from a file, as PEM public keys will often come in file form.

Also, would be good to give an example with PEM key? though not sure how possible that is due to how large they tend to be

pkg/flowcli/services/keys.go
```diff
@@ -78,16 +82,45 @@ func (k *Keys) Generate(inputSeed string, signatureAlgo string) (crypto.PrivateK
...
-func (k *Keys) Decode(publicKey string) (*flow.AccountKey, error) {
+// Decode encoded public key for supported encoding type
+func (k *Keys) Decode(encoded string, encoding string) (*flow.AccountKey, error) {
+	if strings.ToLower(encoding) == PEM {
+		return decodePEM(encoded)
+	} else if strings.ToLower(encoding) == RLP {
+		return decodeRLP(encoded)
+	}
+
+	return nil, fmt.Errorf("encoding type not supported. Valid encoding: RLP and PEM")
```
psiemens, line 93:
```suggestion
	return nil, fmt.Errorf("encoding type not supported. Valid encodings: RLP, PEM")
```

pkg/flowcli/services/keys.go
```diff
@@ -78,16 +82,45 @@ func (k *Keys) Generate(inputSeed string, signatureAlgo string) (crypto.PrivateK
 	return privateKey, nil
 }
 
-func (k *Keys) Decode(publicKey string) (*flow.AccountKey, error) {
+// Decode encoded public key for supported encoding type
```
psiemens, line 85:
```suggestion
// Decode encoded public key using supported encoding type
```

docs/decode-keys.md
```diff
@@ -1,19 +1,19 @@
 ---
-title: Decode Account Key with the Flow CLI
-sidebar_title: Decode Account Keys
-description: How to decode rlp encoded key pair from the command line
+title: Decode Account Keys with the Flow CLI
+sidebar_title: Decode Public Keys
+description: How to decode encoded public keys from the command line
```
psiemens, line 4:
```suggestion
description: How to decode Flow public keys from the command line
```

"decode encoded" is a bit redundant. This is better SEO too 😄

psiemens (commented):
I agree it'd be good to have a test case for PEM.

You can probably just adapt the test case in the Go SDK: https://github.com/onflow/flow-go-sdk/blob/master/crypto/crypto_test.go#L138

sideninja:
> was thinking to also be able to decode from a file, as PEM public keys will often come in file form.
> 
> Also, would be good to give an example with PEM key? though not sure how possible that is due to how large they tend to be

Yes it currently is that you provide a pathname to a file with a PEM key, so yes I need to make docs more clear about that.

sideninja:
> I agree it'd be good to have a test case for PEM.
> 
> You can probably just adapt the test case in the Go SDK: https://github.com/onflow/flow-go-sdk/blob/master/crypto/crypto_test.go#L138

The problems with writing that test is that I can not easily inject file loader and I plan to improve testing as part of separate PR where I will also make sure all file loaders are injected so they can be swapped during testing with mock ones

internal/keys/keys.go
```diff
@@ -77,8 +77,11 @@ func (k *KeyResult) String() string {
 	if k.accountKey != nil {
 		_, _ = fmt.Fprintf(writer, "Signature algorithm \t %s\n", k.accountKey.SigAlgo)
 		_, _ = fmt.Fprintf(writer, "Hash algorithm \t %s\n", k.accountKey.HashAlgo)
-		_, _ = fmt.Fprintf(writer, "Weight \t %d\n", k.accountKey.Weight)
 		_, _ = fmt.Fprintf(writer, "Revoked \t %t\n", k.accountKey.Revoked)
+
+		if k.accountKey.Weight >= 0 { // dont show invalid values
```
janezpodhostnik, line 82: In which case would this be < 0. Should this also include cases where the weight is more than the max 1000?

pkg/flowcli/services/keys.go
```diff
@@ -78,16 +82,45 @@ func (k *Keys) Generate(inputSeed string, signatureAlgo string) (crypto.PrivateK
 	return privateKey, nil
 }
 
-func (k *Keys) Decode(publicKey string) (*flow.AccountKey, error) {
+// Decode encoded public key using supported encoding type
+func (k *Keys) Decode(encoded string, encoding string) (*flow.AccountKey, error) {
+	if strings.ToLower(encoding) == PEM {
```
janezpodhostnik, line 87: should this be a switch case?

pkg/flowcli/services/keys.go
```diff
@@ -78,16 +82,45 @@ func (k *Keys) Generate(inputSeed string, signatureAlgo string) (crypto.PrivateK
...
+func decodeRLP(publicKey string) (*flow.AccountKey, error) {
 	publicKeyBytes, err := hex.DecodeString(publicKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode public key: %v", err)
 	}
 
 	accountKey, err := flow.DecodeAccountKey(publicKeyBytes)
 	if err != nil {
-		return nil, fmt.Errorf("failed to decode private key bytes: %v", err)
+		return nil, fmt.Errorf("failed to decode: %v", err)
```
janezpodhostnik, line 104: should `%w` be used instead of `%v` for errors? https://blog.golang.org/go1.13-errors#TOC_3.3.

pkg/flowcli/services/keys.go
```diff
@@ -78,16 +82,45 @@ func (k *Keys) Generate(inputSeed string, signatureAlgo string) (crypto.PrivateK
...
 	return accountKey, nil
 }
+
+func decodePEM(publicKeyPath string) (*flow.AccountKey, error) {
+	fileContent, err := util.LoadFile(publicKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	pk, err := crypto.DecodePublicKeyPEM(crypto.ECDSA_P256, string(fileContent))
```
janezpodhostnik, line 116: Is `crypto.ECDSA_P256` a valid assumption to make here?

pkg/flowcli/services/keys.go
psiemens: Ah yep, it's best to use `%w` when wrapping errors like this.

pkg/flowcli/services/keys.go
psiemens: Good catch. I think we need to support `secp256k1` as well. Maybe add a `--sig-algo` flag to the command?

sideninja: ohh... yeah, thanks for this, I actually planned to do this and then I changed from draft to open and forgot, I will do this during the QA.

pkg/flowcli/services/keys.go
sideninja: TIL

internal/keys/keys.go
sideninja: Just input validation so yeah I could check for max min

Kay-Zee:
> > was thinking to also be able to decode from a file, as PEM public keys will often come in file form.
> > Also, would be good to give an example with PEM key? though not sure how possible that is due to how large they tend to be
> 
> Yes it currently is that you provide a pathname to a file with a PEM key, so yes I need to make docs more clear about that.

does that mean for rlp you provide the hex, but for PEM you provide the file name? Might be somewhat confusing

internal/keys/keys.go
sideninja: this is used when decoding a PEM key and there is no weight attached to the key so negative value is used as an empty value so we don't show it

docs/decode-keys.md
```diff
@@ -1,29 +1,62 @@
 ---
-title: Decode Account Key with the Flow CLI
-sidebar_title: Decode Account Keys
-description: How to decode rlp encoded key pair from the command line
+title: Decode Account Keys with the Flow CLI
```
psiemens, line 2:
```suggestion
title: Decode Public Keys with the Flow CLI
```

internal/keys/decode.go
```diff
@@ -43,14 +47,23 @@ var DecodeCommand = &command.Command{
...
-		rlpEncoded := args[0]
+		encoding := args[0]
+
+		var encoded string
+		if len(args) > 1 {
+			encoded = args[1]
+		}
 
-		accountKey, err := services.Keys.Decode(rlpEncoded)
+		/* todo from file flag should be remove and should be replaced with $(echo file)
```
psiemens, line 57:
```suggestion
		/* TODO: --from-file flag should be removed and replaced with $(echo file)
```

internal/keys/keys.go
```diff
@@ -77,8 +77,11 @@ func (k *KeyResult) String() string {
 	if k.accountKey != nil {
 		_, _ = fmt.Fprintf(writer, "Signature algorithm \t %s\n", k.accountKey.SigAlgo)
 		_, _ = fmt.Fprintf(writer, "Hash algorithm \t %s\n", k.accountKey.HashAlgo)
-		_, _ = fmt.Fprintf(writer, "Weight \t %d\n", k.accountKey.Weight)
 		_, _ = fmt.Fprintf(writer, "Revoked \t %t\n", k.accountKey.Revoked)
+
+		if k.accountKey.Weight >= 0 { // dont show empty value
```
psiemens, line 82: Won't `k.accountKey` be `0` when empty?

sideninja: Yes, this was put in place because when you do PEM decoding the result is returned as`AccountKey` but I don't want to display weight in the result since it is not known and I set it as -1 which then gets omitted in the display. If you have a better idea please let me know as I don't really like it, but it does the job without creating new structs. I do want to display keys with weight 0 since I believe this is valid.

psiemens: Ah I see, didn't realize it was being set to -1.

psiemens (approved):
LGTM

PR #271, opened 2021-05-27
Refactor event display

sideninja:
Refactor events print out.
More testing done for different formats of events.


turbolent (commented):
Thank you for improving this! It would be nice to have a generic formatting function instead of special-casing certain fields and types 👍

psiemens (approved):

PR #286, opened 2021-06-11
Refactor Typed Service API

### LINKED ISSUE #293: Refactoring and Improving Tests
### Issue To Be Solved

### Suggest A Solution

sideninja:
Closes #293 

## Description
Addressing the issue specifically on the API refactor.

Converting all service APIs to accept "typed" arguments and not "raw" string arguments. 
This will lead to better reusability of the flowkit package and easier testing.

Refactor also removes some code duplication since the API no longer takes the raw arguments we surface the argument preparation which leads to less complicated APIs as they are less granular.
______

For contributor use:

### DIFF
--- a/internal/accounts/create.go
+++ b/internal/accounts/create.go
@@ -59,7 +59,7 @@ var CreateCommand = &command.Command{
 		args []string,
 		globalFlags command.GlobalFlags,
 		services *services.Services,
-		project *project.Project,
+		proj *project.Project,
 	) (command.Result, error) {
 		if createFlags.Results {
 			fmt.Println("⚠️ DEPRECATION WARNING: results flag is deprecated, results are by default included in all executions")
--- a/internal/accounts/create.go
+++ b/internal/accounts/create.go
@@ -20,6 +20,9 @@ package accounts
 
 import (
 	"fmt"
+	"strings"
+
+	"github.com/onflow/flow-go-sdk/crypto"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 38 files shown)


internal/accounts/create.go
```diff
@@ -52,17 +59,55 @@ var CreateCommand = &command.Command{
...
+		if hashAlgo == crypto.UnknownHashAlgorithm {
+			return nil, fmt.Errorf("invalid hash algorithm: %s", createFlags.HashAlgo)
+		}
+
+		keyWeights := createFlags.Weights
+
+		// decode public keys
+		var pubKeys []crypto.PublicKey
+		for _, k := range createFlags.Keys {
+			k = strings.ReplaceAll(k, "0x", "") // clear possible prefix
```
janezpodhostnik, line 88: https://golang.org/pkg/strings/#TrimPrefix

internal/config/init.go
```diff
@@ -53,13 +55,30 @@ var InitCommand = &command.Command{
 		args []string,
 		globalFlags command.GlobalFlags,
 		services *services.Services,
+		proj *project.Project,
 	) (command.Result, error) {
-		proj, err := services.Project.Init(
+
+		sigAlgo := crypto.StringToSignatureAlgorithm(initFlag.ServiceKeySigAlgo)
```
janezpodhostnik, line 61: this code is also here https://github.com/onflow/flow-cli/pull/286/files#diff-e012f7fa2218d10a3c80f965cd049a7cfbfb2e7c89a3c5fdf4c8b2fbdc24b0ecR73. Could be a function...

internal/quick/init.go
```diff
@@ -42,13 +46,29 @@ var InitCommand = &command.Command{
 		args []string,
 		globalFlags command.GlobalFlags,
 		services *services.Services,
+		proj *project.Project,
 	) (command.Result, error) {
-		proj, err := services.Project.Init(
+		sigAlgo := crypto.StringToSignatureAlgorithm(initFlag.ServiceKeySigAlgo)
```
janezpodhostnik, line 51: This also looks like the same code as here https://github.com/onflow/flow-cli/pull/286/files#diff-e012f7fa2218d10a3c80f965cd049a7cfbfb2e7c89a3c5fdf4c8b2fbdc24b0ecR73

janezpodhostnik (commented):
Looks good! I just had a few suggestions.

internal/transactions/get.go
```diff
@@ -59,10 +64,9 @@ var GetCommand = &command.Command{
 			fmt.Println("⚠️  DEPRECATION WARNING: use include flag instead")
 		}
 
-		tx, result, err := services.Transactions.GetStatus(
-			args[0], // transaction id
-			getFlags.Sealed,
-		)
+		id := flow.HexToID(strings.ReplaceAll(args[0], "0x", ""))
```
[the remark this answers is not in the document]
sideninja: ohh right nice

internal/config/init.go
sideninja: Not sure what do you mean?

janezpodhostnik: I thought that this part could be replaced with a library function, since its repeated a few times. But now that I look at it again, it wouldn't really make the method much shorter.

janezpodhostnik (approved):

internal/config/init.go
sideninja: Yeah, I had it extracted in lib function but then I realised I'm just putting another layer without really shortening the code

PR #287, opened 2021-06-11
Refactor Project

### LINKED ISSUE #293: Refactoring and Improving Tests

sideninja:
Closes #293 

## Description
The `project` package was renamed to `state` as it is more accurate description of the package purpose.

The `project.go` was refactored and some code was extracted into `account.go`. 
The newly created state and accounts alongside transactions were moved to the top-level package `flowkit`.

State (project) was also refactored to provide better API for accessing configuration which is an implementation detail and it is now hidden and only accessed through state APIs.

This was also done in preparation for file loader injection.
______

For contributor use:

### DIFF
--- a/internal/keys/generate.go
+++ b/internal/keys/generate.go
@@ -51,7 +51,7 @@ var GenerateCommand = &command.Command{
 		args []string,
 		globalFlags command.GlobalFlags,
 		services *services.Services,
-		proj *flowkit.State,
+		state *flowkit.State,
 	) (command.Result, error) {
 		if generateFlags.Algo != "" {
 			fmt.Println("⚠️ DEPRECATION WARNING: flag no longer supported, use '--sig-algo' instead.")
--- a/internal/keys/generate.go
+++ b/internal/keys/generate.go
@@ -51,7 +51,7 @@ var GenerateCommand = &command.Command{
 		args []string,
 		globalFlags command.GlobalFlags,
 		services *services.Services,
-		proj *flowkit.Project,
+		proj *flowkit.State,
 	) (command.Result, error) {
 		if generateFlags.Algo != "" {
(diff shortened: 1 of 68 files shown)


pkg/flowkit/account.go
```diff
@@ -121,3 +117,68 @@ func generateEmulatorServiceAccount(sigAlgo crypto.SignatureAlgorithm, hashAlgo
...
 }
+
+// Accounts is a collection of account.
+type Accounts []Account
+
+// Remove an account.
+func (a *Accounts) Remove(name string) error {
+	account := a.ByName(name)
+	if account == nil {
+		return fmt.Errorf("account named %s does not exist in configuration", name)
```
janezpodhostnik, line 128: This is just personal preference, but I feel Remove operations should be nilpotent.

janezpodhostnik (approved):
Looks good.

internal/keys/generate.go
```diff
@@ -21,12 +21,13 @@ package keys
 import (
 	"fmt"
 
+	"github.com/onflow/flow-cli/pkg/flowkit"
```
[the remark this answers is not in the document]
sideninja: ohh yeah, I need to setup auto-linting for this not sure why it doesn't work in Goland for me... will fix tnx

pkg/flowkit/account.go
sideninja: Do you mean idempotent or else please explain not sure what do you mean?

janezpodhostnik: Yes sorry, spelling mistake.

I meant that I should be able to call `Remove("x")` multiple times without any problems (errors). Like I said, just a personal preference. Feel free to ignore.

sideninja: yeah I kinda agree with you, idempotent functions are always better, will change

PR #290, opened 2021-06-11
Refactor Service Tests

### LINKED ISSUE #293: Refactoring and Improving Tests

sideninja:
Closes #293 

## Description
Refactoring current service tests to share common logic, improve mocks and plan for what codebase they are meant to cover. 

This refactor will make sure we have quality code coverage and make tests more maintainable.

______

For contributor use:


tests/mocks/Gateway.go
```diff
@@ -0,0 +1,263 @@
+// Code generated by mockery 2.7.5. DO NOT EDIT.
```
janezpodhostnik, line 1: does this generated code need to be committed?

janezpodhostnik (approved):

tests/mocks/Gateway.go
sideninja: It doesn't and I plan to remove it, but I will do that after I setup GitHub actions to auto-generate it before running tests since it is required during the testing, but good flag

PR #292, opened 2021-06-11
Refactor File Loading

### LINKED ISSUE #293: Refactoring and Improving Tests

sideninja:
Closes #293 

## Description
Addressing the issue specifically on the file loading.

This is an Important refactor. It extracts all resource loading implementation out of the service layer so we can inject an interface that describes the resource loading implementation and can be injected and mocked during testing.

Reader and writer interface was defined and the command run function added this "reader writer" interface as the argument so we can use this injected loader to do all file loading in commands and the same was done for `state` where we inject the reader writer implementation and then use it whenever we need to load or save a resource.

______

For contributor use:


janezpodhostnik (approved):
LGTM

PR #300, opened 2021-06-18
Fix for comma issue on --arg parameter value

bluesign:
## Description

--arg parameters parsing values as CSV with cobra, this PR fixes that.

For example below example fails: 

`flow transactions send my_transaction.cdc --arg "String:arg1" --arg "Array:['arrArg1','arrArg2']"
`
______
For contributor use:

### DIFF
--- a/internal/scripts/execute.go
+++ b/internal/scripts/execute.go
@@ -60,6 +60,8 @@ var ExecuteCommand = &command.Command{
 			return nil, fmt.Errorf("provide a valide filename command argument")
 		}
 
+		scriptFlags.Arg, _ = cmd.Flags().GetStringArray("arg")
+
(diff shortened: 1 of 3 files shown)


internal/scripts/execute.go
```diff
@@ -60,6 +60,8 @@ var ExecuteCommand = &command.Command{
 			return nil, fmt.Errorf("provide a valide filename command argument")
 		}
 
+		scriptFlags.Arg, _ = cmd.Flags().GetStringArray("arg")
```
sideninja, line 63: Running this with same flag syntax seems to produce error: `trying to get stringArray value of flag of type stringSlice`

sideninja (changes requested):

PR #308, opened 2021-06-24
Upgrade to flow-emulator v0.22.0

psiemens:
## Description

This PR upgrades the embedded emulator to v0.22.0. https://github.com/onflow/flow-emulator/releases/tag/v0.22.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,9 +10,9 @@ require (
 	github.com/onflow/cadence v0.18.1-0.20210621144040-64e6b6fb2337
 	github.com/onflow/cadence/languageserver v0.18.2
 	github.com/onflow/flow-core-contracts/lib/go/templates v0.6.0
-	github.com/onflow/flow-emulator v0.21.0
-	github.com/onflow/flow-go v0.18.2-canary
-	github.com/onflow/flow-go-sdk v0.20.0-alpha.1
+	github.com/onflow/flow-emulator v0.22.0
+	github.com/onflow/flow-go v0.18.4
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -10,9 +10,9 @@ require (
 	github.com/onflow/cadence v0.18.1-0.20210621144040-64e6b6fb2337
 	github.com/onflow/cadence/languageserver v0.18.2
 	github.com/onflow/flow-core-contracts/lib/go/templates v0.6.0
-	github.com/onflow/flow-emulator v0.21.0
-	github.com/onflow/flow-go v0.18.2-canary
-	github.com/onflow/flow-go-sdk v0.20.0-alpha.1
+	github.com/onflow/flow-emulator v0.22.0
+	github.com/onflow/flow-go v0.18.4
+	github.com/onflow/flow-go-sdk v0.20.1-0.20210623043139-533a95abf071
```
turbolent, line 15: Was this automatic or manual? I see it's from onflow/flow-go-sdk#197, which isn't ready yet. If we really need the update of the SDK to Cadence v0.18, then it should be the previous commit, not the patch release

psiemens: This was automatic because the emulator depends on it: https://github.com/onflow/flow-emulator/blob/master/go.mod#L14

turbolent: Ah, I had missed it there. I guess it's OK, but now building it from source will fail no?

turbolent: I think it didn't fail there, because it didn't use any `cadence.NewString` calls (the breaking change)

turbolent (approved):
Looks good!

I think if CI can build the CLI it should be fine

PR #313, opened 2021-06-28
use a newer block height and specify network in event text

### LINKED ISSUE #312: help text for flow events get does not mention that you should specify network.
## Instructions

The help text example for `flow events get` is not very helpful

### Issue To Be Solved

Use a query that works against mainnet

bjartek:
Closes #312 

## Description

______

For contributor use:


sideninja (changes requested):
I don't think this is a solution since the block will get stale at some point, I think it would be better if we give an example using the latest block -10, which is being developed.

bjartek:
agreed.

PR #319, opened 2021-06-29
Fetch many events

### LINKED ISSUE #316: Fetch multiple events for large block ranges using goroutines/channels
## Instructions

I have some code in go-with-the-flow https://github.com/bjartek/go-with-the-flow/blob/master/gwtf/event.go#L247 that I am pondering changing and submitting to this cli project. The idea is that it should be possible to specify several events and a large block size and partition the rangeQueries and run them in paralell. 

### Issue To Be Solved

Fetching two or more events for a large block size is currently not possible. 

Include the go-with-the-flow channel/go-routine code into flow cli and exapand on the `flow events get` methods or possible introduct a `flow events get-many` command

bjartek:
Closes #316 

This is not done, it is just a copy of some code and cleaning it up a little. I do not want to invest more time in this if this is something that you feel is not needed here. 

Personally I think this gives a lot of value.

______

For contributor use:


internal/events/get.go
```diff
@@ -55,7 +56,9 @@ func get(
 		end = args[2]
 	}
 
-	events, err := services.Events.Get(name, start, end)
+
+	eventNames := strings.Split(name, ",")
+	events, err := services.Events.GetMany(eventNames, start, end, 249, 10)
```
bjartek, line 61: obviously this is not going to be here for the final version, but I added it here to test it out.

bjartek: Not really sure if we want to expose the worker count either. It could be calculated from the queries slices and some properties.

bjartek:
There are some test failures here and I do not understand why they fail. Somebody who knows the internals of the emulator better then me need to take a look at it :/

internal/events/get.go
```diff
@@ -26,18 +26,29 @@ import (
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
 )
 
-type flagsGenerate struct{}
+type flagsEvents struct {
+	Start   uint64 `flag:"start" info:"Block height start"`
+	End     uint64 `flag:"end" info:"Block height end"`
+	Last    uint64 `default:"1" flag:"last" info:"Fetch last number of block relative to latestBlocks. Will be ignored if --start set"`
```
sideninja, line 32:
```suggestion
	Last    uint64 `default:"1" flag:"last" info:"Fetch number of blocks relative to the last block. Ignored if start flag is set"`
```

internal/events/get.go
```diff
@@ -26,18 +26,29 @@ import (
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
 )
 
-type flagsGenerate struct{}
+type flagsEvents struct {
+	Start   uint64 `flag:"start" info:"Block height start"`
```
sideninja, line 30:
```suggestion
	Start   uint64 `default:"" flag:"start" info:"Start block height"`
```

internal/events/get.go
```diff
@@ -26,18 +26,29 @@ import (
...
+	End     uint64 `flag:"end" info:"Block height end"`
```
sideninja, line 31:
```suggestion
	End     uint64 `default:"" flag:"end" info:"End block height"`
```

internal/events/get.go
```diff
@@ -26,18 +26,29 @@ import (
...
-var generateFlag = flagsGenerate{}
+var eventsFlags = flagsEvents{}
 
 var GetCommand = &command.Command{
 	Cmd: &cobra.Command{
-		Use:     "get <event_name> <block_height_range_start> <optional:block_height_range_end|latest>",
-		Short:   "Get events in a block range",
-		Args:    cobra.RangeArgs(2, 3),
-		Example: "flow events get A.1654653399040a61.FlowToken.TokensDeposited 11559500 11559600",
+		Use:   "get <event_name> (<event_name> <event_name>)",
```
sideninja, line 41: This is somewhat confusing, I actually from reading the help can't figure out what optionally second or third <event_name> means. I would strip that from the usage as it is an advance use case and document this in the docs more.

internal/events/get.go
```diff
@@ -26,18 +26,29 @@ import (
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
 )
 
-type flagsGenerate struct{}
+type flagsEvents struct {
+	Start   uint64 `flag:"start" info:"Block height start"`
+	End     uint64 `flag:"end" info:"Block height end"`
+	Last    uint64 `default:"1" flag:"last" info:"Fetch last number of block relative to latestBlocks. Will be ignored if --start set"`
+	Workers int    `default:"10" flag:"workers" info:"Number of workers to use when fetching events in parallel"`
+	Batch   uint64 `default:"250" flag:"batch" info:"Number of blocks to batch together when fetching events in parallel"`
```
sideninja, line 34:
```suggestion
	Batch   uint64 `default:"250" flag:"batch" info:"Number of blocks each worker will fetch"`
```

internal/events/get.go
```diff
@@ -47,15 +58,12 @@ func get(
 	_ command.GlobalFlags,
 	services *services.Services,
 ) (command.Result, error) {
-	name := args[0]
-	start := args[1] // block height range start
-	end := ""        // block height range end
-
-	if len(args) == 3 {
-		end = args[2]
+	start, end, err := services.Events.CalculateStartEnd(eventsFlags.Start, eventsFlags.End, eventsFlags.Last)
```
sideninja, line 61: Argument parsing should be done in the command itself, in this current implementation it exposes this API to anyone using the flowkit as a lib and I don't think it is necessary, because when using the lib programmatically you don't necessarily use user input.

internal/events/get.go
```diff
@@ -47,15 +58,12 @@ func get(
...
+	if err != nil {
+		return nil, err
 	}
 
-	events, err := services.Events.Get(name, start, end)
+	events, err := services.Events.Get(args, start, end, eventsFlags.Batch, eventsFlags.Workers)
```
sideninja, line 66: As in the comment above, we should have fine-grained API which defines each argument separately.

pkg/flowkit/services/events.go
```diff
@@ -50,46 +50,107 @@ func NewEvents(
...
+		return 0, 0, fmt.Errorf("cannot have end height (%d) of block range less that start height (%d)", end, start)
+	}
+	return start, end, nil
+}
+
+func makeEventQueries(events []string, startHeight uint64, endHeight uint64, blockCount uint64) []client.EventRangeQuery {
+	var queries []client.EventRangeQuery
+	for startHeight <= endHeight {
+		suggestedEndHeight := startHeight + blockCount - 1 //since we are inclusive
+		endHeight := endHeight
```
sideninja, line 81: ?

pkg/flowkit/services/events_test.go
```diff
@@ -34,74 +34,74 @@ func TestEvents(t *testing.T) {
 		t.Parallel()
 
 		_, s, gw := setup()
-		_, err := s.Events.Get("flow.CreateAccount", "0", "1")
+		_, err := s.Events.Get([]string{"flow.CreateAccount"}, 0, 0, 250, 1)
```
sideninja, line 37: Missing test for multiple workers.

sideninja (changes requested):
I look forward to this functionality, but I left some comments for you to look at.

pkg/flowkit/services/events.go
sideninja, line 81: Not sure why is this necessary?

internal/events/get.go
bjartek: When calling this code from literaly any other place except the cli you would use actual block heights here. Having the services api know about -10 as meaning the 10 latest blocks is not very logical if you ask me. I would prefer parsing it in the cli layer and having a service layer that gets startBlock/endBlocks as uints as data.

pkg/flowkit/services/events.go
bjartek: flow events get A.1654653399040a61.FlowToken.TokensDeposited 16128771 16129021
The difference here is 250, but the blockchain says 
❌ Command Error: client: rpc error: code = Unknown desc = requested block range (251) exceeded maximum (250)

Since fetching events are inclusive to the start block.

internal/events/get.go
bjartek: We could move the method into the cli layer and just use the service layers block service to fetch the latest block height.

sideninja: Yeah, I agree that having the service layer know about last X blocks is not ideal, but then the API should be changed to the proposed one `services.Events.Get(name, start, end)` and the start/end can be calculated accordingly in the command layer as you said. I think that having that function for the block calculation in the events API is not so great because of the same reasons you just mentioned - not many devs will use last X blocks when using the SDK.

pkg/flowkit/services/events.go
sideninja: No, I mean this statement specifically `endHeight := endHeight`

bjartek: Remove it. legacy from when i has the modifier on line 81 on that line instead.

internal/events/get.go
bjartek: inlined the argument parsing into the cli layer and removed the method from the service layer.

pkg/flowkit/services/events.go
bjartek: I am not completely sure why but this line needs to be here, if it is not it will not work, It will use the endHeight from the previous iteration of the loop. Try to remove it and observe the tests.

sideninja: There is a difference if that statement gets deleted because you are actually creating another variable within for block that is different from the value the for loop uses. That is why I think it would be better to either change the naming or rethink how this function is written.

bjartek:
Can  this be merged? I want to realease go with the flow v2 but need this in orrder to do so.

internal/events/get.go
```diff
@@ -19,25 +19,44 @@
...
-type flagsGenerate struct{}
+type flagsEvents struct {
+	Start   uint64 `flag:"start" info:"Start block height"`
+	End     uint64 `flag:"end" info:"End block height"`
+	Last    uint64 `default:"1" flag:"last" info:"Fetch number of blocks relative to the last block. Ignored if the start flag is set. Used as a default if no flags are provided"`
+	Workers int    `default:"10" flag:"workers" info:"Number of workers to use when fetching events in parallel"`
+	Batch   uint64 `default:"250" flag:"batch" info:"Number of blocks each worker will fetch"`
+}
 
-var generateFlag = flagsGenerate{}
```
sideninja, line 31: nice catch, thanks

internal/events/get.go
```diff
@@ -19,25 +19,44 @@
...
+		Example: `#fetch events from the latest 10 blocks is the default behavior
+flow events get A.1654653399040a61.FlowToken.TokensDeposited
+
+#specify manual start and stop blocks
+flow events get A.1654653399040a61.FlowToken.TokensDeposited --start 11559500 --end 11559600
+
+#in order to get and event from the 10 latest blocks on a network run
+flow events get A.1654653399040a61.FlowToken.TokensDeposited --last 10 --network mainnet
+
+#if you want to fetch multiple event types that is done by sending in more events. Even fetching will be done in parallel.
```
sideninja, line 55: great help section

internal/events/get.go
```diff
@@ -19,25 +19,44 @@
...
+flow events get A.1654653399040a61.FlowToken.TokensDeposited flow events get A.1654653399040a61.FlowToken.TokensWithdrawn
```
sideninja, line 56: great help section
```suggestion
flow events get A.1654653399040a61.FlowToken.TokensDeposited A.1654653399040a61.FlowToken.TokensWithdrawn
```

pkg/flowkit/services/events_test.go
```diff
@@ -34,74 +36,81 @@ func TestEvents(t *testing.T) {
...
-			{"test", "1", "-1"},
-			{"test", "10", "5"},
-		}
+	t.Run("Test create queries", func(t *testing.T) {
 
-		outputs := []string{
-			"cannot use empty string as event name",
-			"failed to parse start height of block range: -1",
-			"failed to parse end height of block range: -1",
-			"cannot have end height (5) of block range less that start height (10)",
```
sideninja, line 69: Is there a reason these failure tests were removed? I would like to leave them be.

There are no tests testing multiple workers which I think should be added since it is quite complex and it can break in the future by mistake.

sideninja (approved):
A couple of small things and then it's ready to go. Really great features.

pkg/flowkit/services/accounts_test.go
```diff
@@ -40,6 +41,17 @@ import (
 	"github.com/onflow/flow-cli/tests"
 )
 
+func setupMock() (*flowkit.State, *Services, *mocks.Gateway) {
```
bjartek, line 45: I added this method to setup since i need to mock differently then you do in setup, if there is a way to get what I want from your way of mocking then please let me know.

sideninja: I think you could just use `setup()` and then override the methods you want to define, no? but if that doesn't work for you it is fine.

sideninja (approved):
Looking good, thank you.

internal/events/get.go
```diff
@@ -47,15 +64,26 @@ func get(
...
+		}
+		start = latestBlockHeight - eventsFlags.Last
+		end = latestBlockHeight
+	} else if eventsFlags.End == 0 {
+		latestBlockHeight, err := services.Blocks.GetLatestBlockHeight()
+		if err != nil {
+			return nil, err
+		}
+		end = latestBlockHeight
 	}
```
sideninja, line 84: Suggested fix (keep in mind this is not tested)

```suggestion
	start := eventsFlags.Start
	end := eventsFlags.End

	if start == 0 {
		end, err := services.Blocks.GetLatestBlockHeight()
		if err != nil {
			return nil, err
		}
		start = end - eventsFlags.Last
	} else if end == 0 {
		end, err := services.Blocks.GetLatestBlockHeight()
		if err != nil {
			return nil, err
		}
	}
```

The suggested fix is based on your logic, but at the same time should we have this logic? If start is not provided and end is by this logic we set the start and end to be latest and latest - last value. I think we should better handle arguments and provide a warning in case some are missing. Proposing (again not tested):

```suggestion
	start := eventsFlags.Start
	end := eventsFlags.End
	last := eventsFlags.Last

	// handle if not passing start and end
	if start == 0 && end == 0 {
		end, err := services.Blocks.GetLatestBlockHeight()
		if err != nil {
			return nil, err
		}
		
		start = end - last
	} else if start == 0 || end == 0 {
		return fmt.Errorf("please provide either both start and end for range or only last flag") 
	}
```

sideninja:
This PR still includes at least two bugs.

If I run events for block range it says that start is greater than last block even though it is not:
```
./main events get A.1654653399040a61.FlowToken.TokensDeposited --start 39460719 --end 39460729 --network mainnet
Fetching events {A.1654653399040a61.FlowToken.TokensDeposited 39460719 39460729}
❌ Command Error: client: rpc error: code = OutOfRange desc = start height 39460719 is greater than the last sealed block height 16781410

➜  flow-cli git:(fetch-many-events) flow blocks get latest -n testnet

Block ID		66f8e600c5cbce941015ff78f47ded311171d229137f45a8817cec3d893d8b88
Parent ID		c098095e48972f299e1a233214807409bc44ec39b2c9af72bf3e45ae9e4f6f65
Timestamp		2021-07-22 12:20:12.657971498 +0000 UTC
Height			39460761
Total Seals		1
Total Collections	0

```

The other bug is if I run get event for last 20 blocks I get result when I specify single event but not if I specify multiple events
```
➜  flow-cli git:(fetch-many-events) ./main events get A.1654653399040a61.FlowToken.TokensDeposited --last 20 --network mainnet                      
Fetching events {A.1654653399040a61.FlowToken.TokensDeposited 16781447 16781467}

Events Block #16781454:
    Index	4
    Type	A.1654653399040a61.FlowToken.TokensDeposited
    Tx ID	2c996dc12a4aeef7d4dc26274b49ddc5c181f0af4b31fb9eb1a192494ee64de4
    Values
		- amount (UFix64): 44.99999999 
		- to (Address?): 0xab970f62ff24fd15 

... (more events)


➜  flow-cli git:(fetch-many-events) ./main events get A.1654653399040a61.FlowToken.TokensDeposited A.1654653399040a61.FlowToken.TokensWithdrawn --last 20 -n testnet
Fetching events {A.1654653399040a61.FlowToken.TokensDeposited 39460915 39460935}
Fetching events {A.1654653399040a61.FlowToken.TokensWithdrawn 39460915 39460935}

(no events)
```

sideninja:
Fetching events doesn't seem to work for me if I get events:
```
./main events get A.7e60df042a9c0868.FlowToken.TokensWithdrawn --last 30 -n testnet
Fetching events {A.7e60df042a9c0868.FlowToken.TokensWithdrawn 39462308 39462338}

(no events)
```

The transaction that should cause the events is:
```
flow transactions get 1aacded1a9883b6682dadb01535c59f83a2a08162a43e1bd7a51798704f182b1 -n testnet

Status		✅ SEALED
ID		1aacded1a9883b6682dadb01535c59f83a2a08162a43e1bd7a51798704f182b1
Payer		df9c30eb2252f1fa
Authorizers	[df9c30eb2252f1fa]

Proposal Key:	
    Address	df9c30eb2252f1fa
    Index	0
    Sequence	132

No Payload Signatures

Envelope Signature 0: df9c30eb2252f1fa
Signatures (minimized, use --include signatures)

Events:		 
    Index	0
    Type	A.7e60df042a9c0868.FlowToken.TokensWithdrawn
    Tx ID	1aacded1a9883b6682dadb01535c59f83a2a08162a43e1bd7a51798704f182b1
    Values
		- amount (UFix64): 0.10000000 
		- from (Address?): 0xdf9c30eb2252f1fa 

```

Steps to reproduce: create an account and then fetch the events caused by that account creation.

bjartek:
I really do not understand what is going on here. Both of the following examples work for me. 

```
flow events get A.7e60df042a9c0868.FlowToken.TokensWithdrawn --last 30 -n testnet
Fetching events {A.7e60df042a9c0868.FlowToken.TokensWithdrawn 39470915 39470945}

Events Block #39470924:
    Index	1
    Type	A.7e60df042a9c0868.FlowToken.TokensWithdrawn
    Tx ID	cc7b7c0bd59a0e64abd691cbf4ebb5240e737fc9aa6483213df5aa8503e9d101
    Values
		- amount (UFix64): 0.00010000
		- from (Address?): 0x8374dbb762ce2d7c

Events Block #39470940:
    Index	1
    Type	A.7e60df042a9c0868.FlowToken.TokensWithdrawn
    Tx ID	9ebc96f0ff2ac3c75b9b9ba84397b1c177a30dcb99818c5e594b3541a7b7276a
    Values
		- amount (UFix64): 0.00010000
		- from (Address?): 0x8374dbb762ce2d7c
```
and

```
flow events get A.0b2a3299cc857e29.TopShot.Withdraw A.0b2a3299cc857e29.TopShot.Deposit --last 10 --network mainnet
```
this returns both deposit and withdraw events. 

I ran `make install` in this branch and my flow version is
```
Commit: fc8841bdcefb9fffc9a7286eb946432e536c5530
```

bjartek:
I just used the testnet faucet to fund my account in tx 1da0ba974c84082b03ae6d16367162ca1d8b5a1be2f61fc2681ccd3935b783eb and fetching events from that with 
```
flow events get A.7e60df042a9c0868.FlowToken.TokensDeposited A.7e60df042a9c0868.FlowToken.TokensWithdrawn A.7e60df042a9c0868.FlowToken.TokensMinted  --last 300 -n testnet
Fetching events {A.7e60df042a9c0868.FlowToken.TokensDeposited 39471216 39471465}
Fetching events {A.7e60df042a9c0868.FlowToken.TokensWithdrawn 39471216 39471465}
Fetching events {A.7e60df042a9c0868.FlowToken.TokensMinted 39471216 39471465}
Fetching events {A.7e60df042a9c0868.FlowToken.TokensDeposited 39471466 39471516}
Fetching events {A.7e60df042a9c0868.FlowToken.TokensWithdrawn 39471466 39471516}
Fetching events {A.7e60df042a9c0868.FlowToken.TokensMinted 39471466 39471516}

Events Block #39471397:
    Index	3
    Type	A.7e60df042a9c0868.FlowToken.TokensWithdrawn
    Tx ID	1da0ba974c84082b03ae6d16367162ca1d8b5a1be2f61fc2681ccd3935b783eb
    Values
		- amount (UFix64): 0.00010000
		- from (Address?): 0x8c5303eaa26202d6

Events Block #39471397:
    Index	1
    Type	A.7e60df042a9c0868.FlowToken.TokensMinted
    Tx ID	1da0ba974c84082b03ae6d16367162ca1d8b5a1be2f61fc2681ccd3935b783eb
    Values
		- amount (UFix64): 1000.00000000

Events Block #39471397:
    Index	2
    Type	A.7e60df042a9c0868.FlowToken.TokensDeposited
    Tx ID	1da0ba974c84082b03ae6d16367162ca1d8b5a1be2f61fc2681ccd3935b783eb
    Values
		- amount (UFix64): 1000.00000000
		- to (Address?): 0x69915b410cca3c65

    Index	4
    Type	A.7e60df042a9c0868.FlowToken.TokensDeposited
    Tx ID	1da0ba974c84082b03ae6d16367162ca1d8b5a1be2f61fc2681ccd3935b783eb
    Values
		- amount (UFix64): 0.00010000
		- to (Address?): 0x912d5440f7e3769e
```

bjartek:
I found your events in block 39461954
```
flow events get A.7e60df042a9c0868.FlowToken.TokensWithdrawn A.7e60df042a9c0868.FlowToken.TokensDeposited A.912d5440f7e3769e.FlowFees.TokensDeposited flow.AccountCreated flow.AccountKeyAdded -n testnet --start 39461954 --end 39461955
```

PR #334, opened 2021-07-22
Contract add mainnet alias

bjartek:
Add the missing --mainnet-signer alias to `flow config add contract`


sideninja (changes requested):
linting fails, otherwise looks good

sideninja (approved):
LGTM

PR #340, opened 2021-07-26
CD Improvements

sideninja:
This PR adds automatic bumping of the formula for the brew to CD. It also implements code scanning for security vulnerabilities.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -5,6 +5,14 @@ on:
     types: [published]
 
 jobs:
+  security:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -30,4 +38,13 @@ jobs:
...
+          destination: flow-cli/
+  homebrew:
+      name: Bump Homebrew formula
+      runs-on: ubuntu-latest
+      steps:
+        - uses: mislav/bump-homebrew-formula-action@v1
+          with:
+            formula-name: flow-cli
+          env:
+            COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
```
Kay-Zee, line 50: I don't see `COMMITTER_TOKEN` in the secrets, are we missing it?

sideninja: Yes, I have to set it up, but I didn't yet found the time, I will do that before it is merged, should've left it as draft PR.

sideninja: @Kay-Zee  Actually looking at it, it seems that I don't have enough permissions to set repo secrets. Can you give me help with this?

janezpodhostnik (approved):

PR #348, opened 2021-08-02
make loader accept the latest valid flow json in path

### LINKED ISSUE #335: Adding a new contract to a local flow.json will copy state from flow.json in home folder
### Instructions

Have a flow.json in your home folder. 

Create a new directory and run  `flow init` in it. 
Observe that the empty flow json is created
run `touch foo && flow config add contracts --name foo --filename foo` 
observe that the flow json has the new contract and all the state from the global flow json in it.

bjartek:
Closes #335 

## Description

Rewrites the loader logic so that it will always accept the latest valid flow.json in the path sent in. So first global flow.json then local flow.json

______

For contributor use:

### DIFF
--- a/pkg/flowkit/config/loader_test.go
+++ b/pkg/flowkit/config/loader_test.go
@@ -65,6 +65,81 @@ func Test_JSONSimple(t *testing.T) {
 	assert.Equal(t, "0x21c5dfdeb0ff03a7a73ef39788563b62c89adea67bbb21ab95e5f710bd1d40b7", conf.Accounts[0].Key.PrivateKey.String())
 }
 
+func Test_ErrorWhenMissingBothDefaultJsonFiles(t *testing.T) {
+	composer := config.NewLoader(afero.Afero{Fs: mockFS})
+	composer.AddConfigParser(json.NewParser())
+
+	_, loadErr := composer.Load(config.DefaultPaths())
+
+	assert.Error(t, loadErr)
+	assert.Contains(t, loadErr.Error(), "missing configuration")
+}
+
+func Test_AllowMissingocalJson(t *testing.T) {
+	b := []byte(`{
+		"accounts": {
+			"emulator-account": {
+				"address": "f8d6e0586b0a20c7",
+				"key": "21c5dfdeb0ff03a7a73ef39788563b62c89adea67bbb21ab95e5f710bd1d40b7"
+			}
+		}
+	}`)
+
+	mockFS := afero.NewMemMapFs()
+	err := afero.WriteFile(mockFS, config.GlobalPath(), b, 0644)
+
+	assert.NoError(t, err)
+
+	composer := config.NewLoader(afero.Afero{Fs: mockFS})
+	composer.AddConfigParser(json.NewParser())
+
+	conf, loadErr := composer.Load(config.DefaultPaths())
+
+	assert.NoError(t, loadErr)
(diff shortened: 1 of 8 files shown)


bjartek:
Not sure what doc to change here really.

sideninja (changes requested):
The config composition shouldn't be removed but amended. The problem with the current change is that if you want to use composition to load configurations it only takes the last configuration, whereas the intended functionality is to load configs and merge all values but overwrite properties if they have the same key, later sources' properties overwrite earlier ones (https://docs.onflow.org/flow-cli/security/#composing-multiple-configuration-files), similar to how Object.assign works in JS. 

The change to the functionality should be that if there is a local config present load only that config, if that config is not found then load global, but if configs are specified by using `-f` flag then we should keep the composition. 

You could check here (https://github.com/onflow/flow-cli/blob/f43841f1a0306e4ebfb5b00ccf3eeff6f13846c1/pkg/flowkit/config/loader.go#L109) if the paths are equal to default (local and global flow.json) and if local flow.json is found then ignore loading global (you can break the for loop if found local and if the only other provided path is global flow.json)

bjartek:
It is better now? I revertetd all the tests and added some more logic. Not sure how to test with afero and homedir though

bjartek:
So right now if -f is not used that is the configPaths are the default ones we load first the last config (default) then the first (global) and error if neither is found. If configPaths are anything else we load everything and merge, but then all files must be present. 

The previous logic actually enabled you to specify -f flow.json -f /some/other/flow.json and if flow.json was not present it would allow it, and I find that to be illogical.

sideninja:
> So right now if -f is not used that is the configPaths are the default ones we load first the last config (default) then the first (global) and error if neither is found. If configPaths are anything else we load everything and merge, but then all files must be present.
> 
> The previous logic actually enabled you to specify -f flow.json -f /some/other/flow.json and if flow.json was not present it would allow it, and I find that to be illogical.

Yes that is true and a good fix for it, however, now that I think about this I think that it would be best if the config command would not allow specifying additional configs by using `-f` flag since it's not intuitive to which file the command saves.

bjartek:
So we should just error out if the config is not default in the config cmd layer?

sideninja:
> Or if you have a single entry to -f it could work?

Right, yeah, single entry would be OK.

bjartek:
But this is best handled in the command layer riht? Because the current logic is valid elsewhere?

sideninja:
> But this is best handled in the command layer right? Because the current logic is valid elsewhere?

Exactly.

pkg/flowkit/config/loader_test.go
```diff
@@ -65,6 +65,81 @@ func Test_JSONSimple(t *testing.T) {
...
+	composer := config.NewLoader(afero.Afero{Fs: mockFS})
+	composer.AddConfigParser(json.NewParser())
+
+	_, loadErr := composer.Load(config.DefaultPaths())
+
+	assert.Error(t, loadErr)
+	assert.Contains(t, loadErr.Error(), "missing configuration")
+}
+
+func Test_AllowMissingocalJson(t *testing.T) {
```
sideninja, line 78:
```suggestion
func Test_AllowMissingLocalJson(t *testing.T) {
```

sideninja (approved):
Left a small comment, otherwise, another great contribution, thank you 🙌

PR #350, opened 2021-08-03
Added errors to accessors methods in config

### LINKED ISSUE #349: Getter methods ByName() for contracts, accounts, networks does not return error in config package
### Instructions

pkg/flowkit/config there are getters for getting configured contracts, networks, accounts. These return an empty pointer if they are not found. It is tempting to then chain together method calls. However if you have then misconfigured things you will get not so nice error messages. 

Should we fix this so that these return error as seond argument?

bjartek:
Closes #349 

## Description

______

For contributor use:


bjartek:
../../go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.2/integration/initialization.go:47:33: multiple-value state.Networks().ByName() in single-value context

I think we might need to change something in languageserver as well?

sideninja:
> ../../go/pkg/mod/github.com/onflow/cadence/languageserver@v0.18.2/integration/initialization.go:47:33: multiple-value state.Networks().ByName() in single-value context
> 
> I think we might need to change something in languageserver as well?

Yep, this line should be updated: https://github.com/onflow/cadence/blob/master/languageserver/integration/initialization.go#L47

But probably also others, so the way you can test that is to pull language server locally, link version of the CLI/flowkit to the latest commit from this PR and then update all the changes, after that push the changes as part of PR and then link the latest commit in LS PR from the CLI so it passes tests.

sideninja (approved):
Looking great, thank you for improving the codebase 👌

bjartek:
I already made a PR in LS, it is referecend above. https://github.com/onflow/cadence/pull/1102

Issue #26, opened 2021-01-06
Fund and defund (FLOW) Emulator accounts


10thfloor:
### Issue To Be Solved

The ability to fund and defund Emulator accounts with FLOW tokens using a cli command

10thfloor:
Connect directly to the faucet via some kind of local webserver.

10thfloor:
Closing in favour of #37

PR #394, opened 2021-10-03
View functions

gauravmarwal:
## Description

This PR is for issue #26.

- Made changes to the code according to the suggestions made.
- Created CLI view commands for accounts, contracts, deployments, networks, and emulators.
- This command shows the list of all accounts, contracts, deployments, networks, or emulators in the configuration. Eg: flow config view account --> shows a list of all accounts present in the configuration.
- It also shows the properties of a particular entity if we pass it as an argument. Eg: flow config view account accountname --> shows the properties of "accountname"

## Requirements Check

- Have have you met the milestone requirements? Yes
- Have you included tests (if applicable)? Yes
- Have you met the contribution guidelines of the repos you have submitted code to (if applicable)? Yes
- If this is the last milestone:
  - Demonstrate that you've met all the acceptance criteria (link to code, demos, instructions to run etc.) 
  Link: Configuration File: [flow.json](https://drive.google.com/file/d/10OA07U0iteT6EwmtslcN7vXfQu9awQnw/view?usp=sharing) 
Commands and Results: [Commands and results](https://drive.google.com/file/d/10AhHgnlTY41vlCBeYCQlA5nzygfviAE0/view?usp=sharing)

## Other Details

- Can you please review that if our solution is up to the mark or not? Also are there any suggestions/reviews to make improvements in our solution?


sideninja (changes requested):
The view command ideally should be just one with resource as argument, furthermore results shouldn't be printed in the command function itself but the notion of Result structs should be used to pass the result from the command and the Result should implement the interface.

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+		// Declaring a string variable to print our final list.
+		var list string = ""
+
+		if len(args) == 1 {
+			for _, value := range *state.Networks() {
+				if count == 0 {
+					list = list + string(value.Name)
+					count = count + 1
+				} else if count > 0 {
+					list = list + ", " + string(value.Name)
```
sideninja, line 129: This is something that could be done with `strings.Join(val, ", ")`

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+		// Count variable for printing format.
+		var count int = 0
+
+		// IF CONDITION:
+		// If there is one argument in the command i.e. command looks like --> flow config view contract,
+		// Then we print the list of all the present accounts in the configuration.
+		// ELSE IF CONDITION:
+		// If there are arguments == 2 i.e. command looks like --> flow config view contract <contractname>,
+		// Then we print all the details of the contract "<contractname>".
+		// 	If the <contractname> doesn't exist in the configuration, then we print "Contract <contractname> does not exist".
```
sideninja, line 163: I don't think this kind of comments are standard, better to avoid.

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+					return &Result{
+						result: fmt.Sprintf("Emulator Name: %s \nPort: %d \nService Account: %s", value.Name, value.Port, value.ServiceAccount),
+					}, nil
+				}
+			}
+			return &Result{
+				result: fmt.Sprintf("Emulator %s does not exist", args[1]),
+			}, nil
+		}
+	}
```
sideninja, line 239: It seems that each of these blocks does the same thing with different values which is a good idea to extract into a function.

internal/config/config.go
```diff
@@ -32,6 +32,7 @@ func init() {
 	InitCommand.AddToParent(Cmd)
 	Cmd.AddCommand(AddCmd)
 	Cmd.AddCommand(RemoveCmd)
+	ViewCmd.AddToParent(Cmd)
```
sideninja, line 35: This should be reversed

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+type flagsViewResource struct {
+}
+
+var viewResourceFlags = flagsViewResource{}
+
+var ViewCmd = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "view",
+		Short:   "View a list of resource entities in the configuration / View the properties of a particular entity",
+		Example: "flow config view account \nflow config view account <accountname>\nflow config view contract \nflow config view contract <contractname>\nflow config view deployment \nflow config view deployment <networkname>\nflow config view emulator \nflow config view emulator <emulatorname>\nflow config view network \nflow config view network <networkname>",
```
sideninja, line 40: not necessary to give such examples for this command

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+		Args:    cobra.MaximumNArgs(2),
```
sideninja, line 41: should be exact arguments

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+				}
+			}
+			return &Result{
+				result: fmt.Sprintf("Network %s does not exist", args[1]),
+			}, nil
+		}
+	}
+
+	// If the argument passed is "contract".
+	if args[0] == "contract" {
```
sideninja, line 152: It seems that each of these blocks does the same thing with different values which is a good idea to extract into a function.

internal/config/view.go
```diff
@@ -0,0 +1,290 @@
...
+				}
+			}
+			return &Result{
+				result: fmt.Sprintf("Account %s does not exist", args[1]),
+			}, nil
+		}
+	}
+
+	// If the argument passed is "network".
+	if args[0] == "network" {
```
sideninja, line 107: It seems that each of these blocks does the same thing with different values which is a good idea to extract into a function.

sideninja:
Closing due to inactivity

PR #396, opened 2021-10-05
Update emulator

sideninja:
## Description
Update emulator to latest version v0.23.0 and also cadence to v0.19.1 with API changes implemented.
______

For contributor use:


pkg/flowkit/arguments_test.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 func TestArguments(t *testing.T) {
-
+	cadenceString, _ := cadence.NewString("42")
```
janezpodhostnik, line 32: I these cases, I think you can just use `cadence.String("42")` which doesn't return an error. Which means there is no need to pull it out.

janezpodhostnik (commented):
Nice! I left a small comment.

pkg/flowkit/arguments_test.go
sideninja: 🤦  thank you

PR #397, opened 2021-10-05
Change default events for Batch to 25 from 250

ph0ph0:
Closes #???

## Description

Changes number of blocks a worker will fetch.

The maximum block range was recently changed from 250 to 25. This fix addresses that by changing the default value for the number of blocks each worker will fetch from 250 to 25.

______

For contributor use:

- [ x] Code follows the [standards mentioned here](https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#styleguides)

### DIFF
--- a/internal/events/get.go
+++ b/internal/events/get.go
@@ -34,7 +34,7 @@ type flagsEvents struct {
 	End     uint64 `flag:"end" info:"End block height"`
 	Last    uint64 `default:"10" flag:"last" info:"Fetch number of blocks relative to the last block. Ignored if the start flag is set. Used as a default if no flags are provided"`
 	Workers int    `default:"10" flag:"workers" info:"Number of workers to use when fetching events in parallel"`
-	Batch   uint64 `default:"250" flag:"batch" info:"Number of blocks each worker will fetch"`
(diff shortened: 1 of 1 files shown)


internal/events/get.go
```diff
@@ -34,7 +34,7 @@ type flagsEvents struct {
 	End     uint64 `flag:"end" info:"End block height"`
 	Last    uint64 `default:"10" flag:"last" info:"Fetch number of blocks relative to the last block. Ignored if the start flag is set. Used as a default if no flags are provided"`
 	Workers int    `default:"10" flag:"workers" info:"Number of workers to use when fetching events in parallel"`
-	Batch   uint64 `default:"250" flag:"batch" info:"Number of blocks each worker will fetch"`
+	Batch   uint64 `default:"25" flag:"batch" info:"Number of blocks each worker will fetch"`
```
sideninja, line 37: I think the best is to use `backend.DefaultMaxHeightRange` so it directly gets the default value from the AN. Since it's not possible to do that in struct tag you should remove the default value from the tag and programmatically check if the variable has a null value and assign default.

sideninja (commented):
Suggested improvement

PR #417, opened 2021-12-07
Fix snyk token

haroldsphinx:
- fix github actions workflow

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,8 +1,8 @@
 name: CD Release
 
 on:
-  push:
-    branches: "fix-flow-cli-gha"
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -9,11 +9,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
-      - uses: actions/setup-go@v1
```
sideninja, line 12: I don't think this should be removed

haroldsphinx: oh, why is that?

sideninja (approved):

PR #431, opened 2021-12-22
Multi key account creation

Kay-Zee:
Closes #???

## Description

This PR adds the ability to specify different hashing and signature algorithms, when creating an account. This was previously not possible, and while we could add multiple keys, they all had to be uniform algorithms for both signing and hashing.

This adds the ability to specify, but also loses the defaults. With that said, do we want defaults for account creation? during the creation process, it almost feels safe to never assume, and always be explicit with the info, especially since if the account is created with the wrong algs, the account will be useless.

______

For contributor use:


Kay-Zee:
running into issues with the integration with language server.

I guess we have to determine if this is something we do want, and if it is, will have to update the language server?

cc @sideninja

internal/accounts/create.go
```diff
@@ -66,23 +66,32 @@ func create(
 		return nil, err
 	}
 
-	sigAlgo := crypto.StringToSignatureAlgorithm(createFlags.SigAlgo)
-	if sigAlgo == crypto.UnknownSignatureAlgorithm {
-		return nil, fmt.Errorf("invalid signature algorithm: %s", createFlags.SigAlgo)
+	sigAlgos := []crypto.SignatureAlgorithm{}
```
sideninja, line 69: nit
```suggestion
	sigAlgos := make([]crypto.SignatureAlgorithm, len(createFlags.SigAlgo))
```

internal/accounts/create.go
```diff
@@ -66,23 +66,32 @@ func create(
...
+	for _, hashAlgoStr := range createFlags.HashAlgo {
+
+		hashAlgo := crypto.StringToHashAlgorithm(hashAlgoStr)
+		if hashAlgo == crypto.UnknownHashAlgorithm {
+			return nil, fmt.Errorf("invalid hash algorithm: %s", createFlags.HashAlgo)
+		}
+		hashAlgos = append(hashAlgos, hashAlgo)
 	}
 
 	keyWeights := createFlags.Weights
```
sideninja, line 89: it might be a good idea to add a check for length mismatch between hash algos, sig algos and keys. Especially to avoid null pointer in line 94, but also to provide sensible error.

sideninja:
Thank you this is great :) Regarding integration and default values, we can go either way technically, we can still provide default values if none algos are provided, but I understand your concerns with this. I think that if we use default values when generating keys we can assume those same default values when creating an account. I think this is consistent enough, however, you can make the decision yourself and just let me know so I can fix the LS accordingly. In case where we still use default values, there is no change needed in case we don't I will handle the change. If we decide to not have default values I feel we should change that in the generate keys as well. 
Also, another thing to keep in mind when changing flags is backward compatibility so here I would strongly advise to keep flags and have a warning as stated in the comment.

internal/accounts/create.go
```diff
@@ -36,8 +36,8 @@ type flagsCreate struct {
 	Signer    string   `default:"emulator-account" flag:"signer" info:"Account name from configuration used to sign the transaction"`
 	Keys      []string `flag:"key" info:"Public keys to attach to account"`
 	Weights   []int    `flag:"key-weight" info:"Weight for the key"`
-	SigAlgo   string   `default:"ECDSA_P256" flag:"sig-algo" info:"Signature algorithm used to generate the keys"`
-	HashAlgo  string   `default:"SHA3_256" flag:"hash-algo" info:"Hash used for the digest"`
```
sideninja, line 40: We need to be careful with backward compatibility when we change flags. I strongly advise we have a check to see if only single flag was provided for multiple keys and still then set all those algos to that provided value but also output a warning that usage is deprecated. I would say to do that in every case where `len(sigAlgo) == 1 && len(keys) > 1`. We also need to be mindful about changing default values, since we potentially will remove default value we need to for a while still provide default values but also provide a deprecation notice. 
An example of such deprecation notices is:
https://github.com/onflow/flow-cli/blob/13897338d416f22bc8bf8f3805a9217d7d039d92/internal/keys/generate.go#L52

tests/resources.go
```diff
@@ -277,12 +277,9 @@ func newAccount(name string, address string, seed string) *flowkit.Account {
 
 func PubKeys() []crypto.PublicKey {
 	var pubKeys []crypto.PublicKey
-	for x := 0; x < 5; x++ {
-		pk, _ := crypto.GeneratePrivateKey(
-			crypto.ECDSA_P256,
-			[]byte(fmt.Sprintf("seedseedseedseedseedseedseedseedseedseedseedseed%d", x)),
-		)
-		pubKeys = append(pubKeys, pk.PublicKey())
+	privKeys := PrivKeys()
```
sideninja, line 280: 👍

Kay-Zee:
Made a few updates, added back the defaults, and now using a branch of the language server to get pass the integration error, so ready for another look @sideninja

internal/accounts/create.go
```diff
@@ -66,23 +66,50 @@ func create(
...
-	if sigAlgo == crypto.UnknownSignatureAlgorithm {
-		return nil, fmt.Errorf("invalid signature algorithm: %s", createFlags.SigAlgo)
+	if len(createFlags.SigAlgo) == 1 && len(createFlags.HashAlgo) == 1 {
+		// Fill up depending on size of key input
+		if len(createFlags.Keys) > 1 {
+			for i := 1; i < len(createFlags.Keys); i++ {
+				createFlags.SigAlgo = append(createFlags.SigAlgo, createFlags.SigAlgo[0])
+				createFlags.HashAlgo = append(createFlags.HashAlgo, createFlags.HashAlgo[0])
+			}
+			// Deprecated usage message?
```
sideninja, line 76: If we will not change the flags I don't think it's needed since this is more of a default handling logic than deprecation

internal/accounts/create.go
```diff
@@ -66,23 +66,50 @@ func create(
...
+		sigAlgos = append(sigAlgos, sigAlgo)
+	}
+
+	// read all hash algorithms
+	hashAlgos := make([]crypto.HashAlgorithm, 0, len(createFlags.HashAlgo))
+	for _, hashAlgoStr := range createFlags.HashAlgo {
+
+		hashAlgo := crypto.StringToHashAlgorithm(hashAlgoStr)
+		if hashAlgo == crypto.UnknownHashAlgorithm {
+			return nil, fmt.Errorf("invalid hash algorithm: %s", createFlags.HashAlgo)
```
sideninja, line 101: Shouldn't this be?
```suggestion
			return nil, fmt.Errorf("invalid hash algorithm: %s", hashAlgoStr)
```

sideninja (approved):
Great, thank you 🚀

PR #432, opened 2021-12-22
Add prompts for build and send-signed

### LINKED ISSUE #410: Transaction confirmation screen
Currently, transaction confirmation screens are only visible for the signing command. It would be better if the confirmation screen is visible on all transactions commands as they have a big impact and can lead to losses if the wrong transaction is sent to the network.

Add confirmation screens to `build`, `send-signed` and `send` commands.

Kay-Zee:
Closes #410

## Description

Adds prompt for the `sendsigned` and `build` commands.

Currently auto accepts for the `send` command, as this is currently used in ways that make upgrading it less convenient and would require other work in additional repos. It's also less important, as it's a single step process so it should be "easier" to know what is being sent. Still would be good to have, but i think is a bigger lift for a smaller gain, so better to just tackle to two higher value commands first.

______

For contributor use:

### DIFF
--- a/pkg/flowkit/services/transactions.go
+++ b/pkg/flowkit/services/transactions.go
@@ -195,10 +195,8 @@ func (t *Transactions) SendSigned(
 	if err != nil {
 		return nil, nil, err
 	}
-	if !approveSend {
-		if !output.ApproveTransactionForSendingPrompt(tx) {
-			return nil, nil, fmt.Errorf("transaction was not approved for sending")
-		}
+	if !approveSend && !output.ApproveTransactionForSendingPrompt(tx) {
+		return nil, nil, fmt.Errorf("transaction was not approved for sending")
 	}
 
 	t.logger.StartProgress(fmt.Sprintf("Sending transaction with ID: %s", tx.FlowTransaction().ID()))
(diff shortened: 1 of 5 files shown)


pkg/flowkit/services/transactions.go
```diff
@@ -180,11 +189,17 @@ func (t *Transactions) Sign(
...
 func (t *Transactions) SendSigned(
 	payload []byte,
+	approveSend bool,
 ) (*flow.Transaction, *flow.TransactionResult, error) {
 	tx, err := flowkit.NewTransactionFromPayload(payload)
 	if err != nil {
 		return nil, nil, err
 	}
+	if !approveSend {
+		if !output.ApproveTransactionForSendingPrompt(tx) {
```
sideninja, line 199: nit: 
```suggestion
	if !approveSend && !output.ApproveTransactionForSendingPrompt(tx) {
```

sideninja:
Thank you this is great. We just need to be careful to update LS accordingly as it uses this API. I can make a PR on top of that. 

The lining error is a nasty one, I was doing an update of dependencies and I will need to make a PR on Go SDK to stop using deprecated functionality, so as long as this is the only listing error I think we can merge until I fix that with PR, I plan to do that as soon as I get the time.

sideninja (approved):
🚀

Issue #435, opened 2022-01-03
Total Tokens Staked is no longer reported


Kay-Zee:
### Problem

`Total Tokens Staked` field in `flow accounts staking-info` returns nil ever since the contract upgrade of the staking contract during Mainnet 9.

### Steps to Reproduce

`flow --host=access.mainnet.nodes.onflow.org:9000 accounts staking-info 8a3adc67b40156bf`

### Acceptance Criteria

We need to either provide an additional function to get this info, or fix the retrieval of this data

### Context

Some partners want to know total tokens staked for a node (including delegators). This is no longer reported using `flow accounts staking-info`

<img width="545" alt="Screen Shot 2022-01-03 at 2 42 35 PM" src="https://user-images.githubusercontent.com/2775157/147988344-8138278d-0308-42cb-b005-92118e9ccb40.png">

janezpodhostnik:
I can take this since I was doing something similar today.

PR #439, opened 2022-01-18
Update staking-info command

janezpodhostnik:
Closes https://github.com/onflow/flow-cli/issues/435

## Description

The "Total Stake" field was removed from NodeInfo during a core contracts [change](https://github.com/onflow/flow-core-contracts/commit/33b80c5ab9b9168f3975b268c164f3a952ef4e1e?diff=unified#diff-65336be374bb3fc9ad7b822243e065a389d73e758c7c16223e52fc3181cea59bL288)

This PR adds it back by fetching it individually for each nodeID.

The two alternative solutions I have considered are:

- Add a script to core contracts that is a combination of three scripts that fetches all NodeInfos and DelegationInfos from the address and the locked address and for each nodeInfo also outputs its total stake.
- Have node total stake as a separate command.

~An update is also needed to the scripts that are being used. This is addressed in this core-contracts [PR](https://github.com/onflow/flow-core-contracts/pull/270). The core-contracts update can be included in this PR or can be added with a separate PR later.~ After talking to Josh, I realised we can just use different scripts to get staking/delegation info.

The output formatting was changed a bit to account for the possibility for multiple stakes and delegations:
image

______

For contributor use:


sideninja:
Thank you, this is good, but do you think we should actually make it secure by default? I was thinking to change that later. But this is a good patch anyhow.

pkg/flowkit/gateway/grpc.go
```diff
@@ -46,7 +47,7 @@ func NewGrpcGateway(host string) (*GrpcGateway, error) {
 
 	gClient, err := client.New(
 		host,
-		grpc.WithInsecure(),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
sideninja, line 50: Thank you, this is good, but do you think we should actually make it secure by default? I was thinking to change that later. But this is a good patch anyhow.

pkg/flowkit/values.go
```diff
@@ -22,21 +22,24 @@ import (
 	"github.com/onflow/cadence"
 )
 
-func NewStakingInfoFromValue(value cadence.Value) map[string]interface{} {
-	stakingInfo := make(map[string]interface{})
+func NewStakingInfoFromValue(value cadence.Value) []map[string]interface{} {
+	stakingInfo := make([]map[string]interface{}, 0)
 	arrayValue := value.(cadence.Array)
```
sideninja, line 27: Is this guaranteed to be array else should you do safe type assertion with a check `v, ok := value.(cadence.Array)` to avoid panic.

pkg/flowkit/values.go
```diff
@@ -22,21 +22,24 @@ import (
...
-	for _, field := range arrayValue.Values[0].(cadence.Struct).StructType.Fields {
-		keys = append(keys, field.Identifier)
-	}
-
-	for i, value := range arrayValue.Values[0].(cadence.Struct).Fields {
-		stakingInfo[keys[i]] = value
+	for _, v := range arrayValue.Values {
+		keys := make([]string, 0)
+		values := make(map[string]interface{})
+		for _, field := range v.(cadence.Struct).StructType.Fields {
```
sideninja, line 36: Same comment as above, type assertion safety

internal/accounts/staking-info.go
```diff
@@ -61,18 +61,61 @@ func stakingInfo(
...
+	for _, stakingInfo := range stakingInfos {
+		nodeID, ok := stakingInfo["id"]
+		if ok {
+			stakingInfo["nodeTotalStake"] = nodeStakes[nodeIDToString(nodeID)].(cadence.UFix64)
+		}
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
sideninja, line 101: I feel like this is a presentation (result building) logic, which could be better if moved to the result object, however the result doesn't have a factory method pattern established so there isn't a place that feels this could be done instead of a method on the struct itself. This is a bit nit point so I don't mind if you feel it should stay like this, but to me the command function itself should only do argument preparation (parsing, validation etc) and then the presentation should be handled in result. But then again a bit nit. Example is here: https://github.com/onflow/flow-cli/blob/janez/staking-stats/internal/transactions/transactions.go#L71 but it's a good point that it could be output format-agnostic which means it should have a common place to do this, but still maybe the common place to do this could be a factory method.

janezpodhostnik: agreed, that it is not the best place to do this. I decided to move this logic inside `accounts.StakingInfo` I feel like this is better. What do you think.

sideninja (approved):
🚀 Looks great!

PR #440, opened 2022-01-20
Khalil/1652 download protocol snapshot

kc1116:
Closes [#1652](https://github.com/dapperlabs/flow-internal/issues/1652)

## Description

This PR adds a command to flow-cli that allows users to download the latest finalized protocol state snapshot.

______

For contributor use:


jordanschalm (commented):
Nice work, very clean addition. 

I have one suggestion for how the output directory is specified.

internal/snapshot/download.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flowkit"
+	"github.com/onflow/flow-cli/pkg/flowkit/services"
+)
+
+type flagsDownload struct {
+	OutputFileName string `default:"root-protocol-state-snapshot.json" flag:"output-filename" info:"The name for the JSON output file"`
+	OutputDir      string `default:"./" flag:"output-dir" info:"The directory to output the protocol state snapshot"`
```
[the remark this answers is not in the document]
sideninja: I second this, and I would even say that you should reuse the common output flag we use in the CLI for saving `--save` at the same time this begs the question of flag vs argument, I've written CLI guidelines where I define when to use which https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#arguments 
In your case if the command can print the details to screen as well then use optional `--save` flag, if command must always save to a file then use argument, because flags by their nature indicate they are optional whereas arguments are required.

pkg/flowkit/gateway/gateway.go
```diff
@@ -38,5 +38,6 @@ type Gateway interface {
 	GetBlockByID(flow.Identifier) (*flow.Block, error)
 	GetEvents(string, uint64, uint64) ([]client.BlockEvents, error)
 	GetCollection(flow.Identifier) (*flow.Collection, error)
+	GetLatestProtocolStateSnapshot() ([]byte, error)
```
sideninja, line 41: I believe you should add this method to the emulator too https://github.com/onflow/flow-cli/blob/master/pkg/flowkit/gateway/emulator.go you can just add `return nil, ftm.Errorf("implement me")`

Makefile
```diff
@@ -93,6 +93,10 @@ install-linter:
 lint:
 	golangci-lint run -v ./...
 
+.PHONY: fix-lint
+fix-lint:
```
janezpodhostnik, line 97: ❤️

internal/snapshot/download.go
```diff
@@ -0,0 +1,72 @@
+/*
+ * Flow CLI
+ *
+ * Copyright 2019-2021 Dapper Labs, Inc.
```
janezpodhostnik, line 4:
```suggestion
 * Copyright 2019-2022 Dapper Labs, Inc.
```

internal/snapshot/download.go
```diff
@@ -0,0 +1,72 @@
...
+	services *services.Services,
+) (command.Result, error) {
+	snapshotBytes, err := services.Snapshot.GetLatestProtocolStateSnapshot()
+	if err != nil {
+		return nil, err
+	}
+
+	outputPath, err := filepath.Abs(filepath.Join(downloadFlags.OutputDir, downloadFlags.OutputFileName))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get absolute output path of protocol snapshot")
```
janezpodhostnik, line 63:
```suggestion
		return nil, fmt.Errorf("failed to get absolute output path for protocol snapshot")
```
Though I think this code might change according to Gregor's comment https://github.com/onflow/flow-cli/pull/440/files#r789478233.

internal/snapshot/snapshot.go
```diff
@@ -0,0 +1,52 @@
+/*
+ * Flow CLI
+ *
+ * Copyright 2019-2021 Dapper Labs, Inc.
```
janezpodhostnik, line 4:
```suggestion
 * Copyright 2019-2022 Dapper Labs, Inc.
```

internal/snapshot/snapshot.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:              "snapshot",
+	Short:            "Utility to download the latest finalized protocol state snapshot",
```
janezpodhostnik, line 29: I think this description should be more along the lines of 

> "Utilities for handling protocol state snapshots"

In case the `snapshot` commands get more subcommands.

internal/snapshot/snapshot.go
```diff
@@ -0,0 +1,52 @@
...
+	TraverseChildren: true,
+}
+
+func init() {
+	DownloadCommand.AddToParent(Cmd)
+}
+
+// DownloadResult represent result snapshot download command.
+type DownloadResult struct {
```
janezpodhostnik, line 38: Should this be defined in `internal/snapshot/download.go`

pkg/flowkit/services/snapshot.go
```diff
@@ -0,0 +1,61 @@
+/*
+ * Flow CLI
+ *
+ * Copyright 2019-2021 Dapper Labs, Inc.
```
janezpodhostnik, line 4:
```suggestion
 * Copyright 2019-2022 Dapper Labs, Inc.
```

janezpodhostnik (commented):
This will be very useful!

internal/snapshot/download.go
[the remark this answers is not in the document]
kc1116: Will change download command to get  @sideninja and combine the 2 flags @jordanschalm

internal/snapshot/snapshot.go
kc1116: Yes , I am following the file conventions of the other commands

pkg/flowkit/gateway/gateway.go
kc1116: I've done that see file right above this one ^

internal/snapshot/download.go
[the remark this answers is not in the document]
kc1116: Also it is not useful to display the contents of the protocol snapshot to the screen because the output is large and hard to reason about in the terminal. Strictly outputting to specified file is the cleanest UX we can provide.

pkg/flowkit/gateway/gateway.go
sideninja: Sorry you are right.

internal/snapshot/download.go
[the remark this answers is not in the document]
sideninja: That is completely understandable then, but I would advise switching the command to require this path as an argument, `flow snapshot save {location}`  or you can even leave download, if we are directly saving and that is the only behaviour then that's good. I wouldn't use a flag for save location because the command requires it.

internal/snapshot/snapshot.go
kc1116: https://github.com/onflow/flow-cli/pull/440/commits/ccb36332b51373cb03d30ad7564f670f72639391

internal/snapshot/download.go
kc1116: https://github.com/onflow/flow-cli/pull/440/commits/ccb36332b51373cb03d30ad7564f670f72639391

internal/snapshot/download.go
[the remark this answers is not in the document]
kc1116: The issue with using the builtin save flag is that before outputResult is invoked the result is [formatted](https://github.com/onflow/flow-cli/blob/fa2ee709a71fed90f17cb37799872e344484cb98/internal/command/command.go#L124) . We avoid formatting the protocol snapshot because we cannot unmarshal the byte representation which is why I am only returning the output location as the DownloadResult .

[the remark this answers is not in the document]
kc1116: Is there a way to disable the json output format for a specific command? We would need to write the bytes directly to the output file , so that we avoid failing [here](https://github.com/onflow/flow-cli/blob/fa2ee709a71fed90f17cb37799872e344484cb98/internal/command/result.go#L73)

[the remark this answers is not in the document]
kc1116: If we add relic to the build process :eyes: then we can parse the snapshot bytes in JSON use the standard save flag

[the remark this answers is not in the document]
sideninja: No no I'm saying that in case if this doesn't make sense to be displayed to the screen (which you said it doesn't) then you should stay with your idea of the command saving to the file, but I'm just saying that a command argument feels more appropriate than a flag to define the location where to save to. Also get doesn't make that much sense if the command always saves, it can stay "download" or "save". So the command should look like: `flow snapshot save ./root-snapshot.json` (note passing the location as argument instead of flag).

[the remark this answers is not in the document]
kc1116: Ahhh Okay I see , I misunderstood your previous comment .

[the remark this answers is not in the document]
kc1116: https://github.com/onflow/flow-cli/pull/440/commits/cff23dc6ca08f838b03aa54456e61bfb2a3d9b54 @sideninja

internal/snapshot/save.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/onflow/flow-cli/pkg/flowkit/services"
+)
+
+var SaveCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "save",
+		Short:   "Get the latest finalized protocol snapshot",
+		Example: "flow snapshot save /tmp/snapshot.json",
+	},
+	Flags: &struct {}{},
```
sideninja, line 39: Should the network flag be available here? deciding on which network you are connecting?

sideninja (approved):
Looks good, just a small comment on flags.

janezpodhostnik (approved):
LGTM!

sideninja (changes requested):
One more thing that is important and needs to be added is the documentation for this command. All commands in the CLI must have a doc.

internal/snapshot/save.go
kc1116: --network part of the global flags

sideninja:
@kc1116 Fixed the linting issue for you

kc1116:
@sideninja the doc for the command https://github.com/onflow/flow-cli/pull/440/commits/931d217754936138a33d3ae0cea334fa9a6c9e48

docs/snapshot-save.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### Example response
+```shell
+snapshot saved: /tmp/snapshot.json
+```
+
+## Arguments
+
+### OutputPath
```
sideninja, line 26: nit:
```suggestion
### Output Path
```

docs/snapshot-save.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Save a protocol snapshot with the FLOW CLI
+sidebar_title: Snapshot Save
+description: How to save a protocol snapshot from the command line
+---
+
+The FLOW CLI provides a command to save the latest finalized protocol state snapshot
+
+```shell
+flow snapshot save <path_to_save_snapshot>
```
sideninja, line 10: nit:
```suggestion
flow snapshot save <output path>
```

docs/snapshot-save.md
```diff
@@ -0,0 +1,103 @@
...
+snapshot saved: /tmp/snapshot.json
+```
+
+## Arguments
+
+### OutputPath
+- OutputPath: `path to save the snapshot file`
+- Valid Input: any valid string path
+
+This is where the protocol snapshot JSON file will be saved.
```
sideninja, line 30:
```suggestion
Output path where the protocol snapshot JSON file will be saved.
```

docs/snapshot-save.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Flags
+
+### Signer
+
+- Flag: `--signer`
+- Valid inputs: the name of an account defined in the configuration (`flow.json`)
+
+Specify the name of the account that will be used to sign the transaction.
```
sideninja, line 39:
```suggestion
```

docs/snapshot-save.md
```diff
@@ -0,0 +1,103 @@
...
+Specify the log level. Control how much output you want to see while command execution.
+
+### Configuration
+
+- Flag: `--config-path`
+- Short Flag: `-f`
+- Valid inputs: valid filename
+
+Specify a filename for the configuration files, you can provide multiple configuration
+files by using `-f` flag multiple times.
```
sideninja, line 98:
```suggestion

```

sideninja (approved):
Thank you, looking good 🚀

PR #449, opened 2022-01-25
Khalil/1659 secure grpc support

kc1116:
Closes https://github.com/dapperlabs/flow-internal/issues/1659

## Description

This PR adds support for secure GRPC client connections. 

______

For contributor use:


sideninja:
A couple of general pointers, backward compatibility is really important with the CLI so you need to make sure the previous `flow.json` schema works with these changes - it currently doesn't. 
Changing the default `flow.json` values for networks to advance schema to then just have empty values for those fields is not ideal, I would like keeping the same old flow.json format and just define an option to specify network key as advance schema where dev decide so. Another flag is to camel-case flow.json properties to be consistent. 
Is secure communication already supported on the backend (we should start backend first and then client)? How can you access public keys currently available for secure connection? We should have them in the docs if they are available, which also brings me to need to update all the docs - each doc should have that flag added (at least all that have host flag added) and the config docs must be updated https://docs.onflow.org/flow-cli/configuration/#networks

There are a lot of DX/UX things above but overall thank you for this, it's a great addition, just needs to be polished a bit.

sideninja (changes requested):
A couple of general pointers, backward compatibility is really important with the CLI so you need to make sure the previous `flow.json` schema works with these changes - it currently doesn't. 
Changing the default `flow.json` values for networks to then just have empty values for those fields is not ideal, I would like keeping the same old defaut `flow.json` format and just define an option to specify network key as an advanced schema where dev decide so (add property to [advance struct here instead of basic](https://github.com/onflow/flow-cli/blob/master/pkg/flowkit/config/json/network.go#L62)). Another flag is to camel-case flow.json properties to be consistent. 
Is secure communication already supported on the backend (we should start backend first and then client)? How can you access public keys currently available for secure connection? We should have them in the docs if they are available, which also brings me to need to update all the docs - each doc should have that flag added (at least all that have host flag added) and the config docs must be updated https://docs.onflow.org/flow-cli/configuration/#networks

There are a lot of DX/UX things above but overall thank you for this, it's a great addition, just needs to be polished a bit.

kc1116:
> 

I will make the adjustments to flow.json and docs. Secure GRPC connections are supported on the backend already, the network key used for the connection is the same network key that access nodes use in the protocol . You can get the access-address and network key for a secure GRPC connection from the protocol via snapshot or identity table smart contract. 

What are the other DX/UX suggestions? @sideninja

sideninja:
> > 
> 
> I will make the adjustments to flow.json and docs. Secure GRPC connections are supported on the backend already, the network key used for the connection is the same network key that access nodes use in the protocol . You can get the access-address and network key for a secure GRPC connection from the protocol via snapshot or identity table smart contract.

OK thank you.

> What are the other DX/UX suggestions? @sideninja
The things I wrote above, nothing else so far.

pkg/flowkit/config/json/network.go
```diff
@@ -48,15 +49,17 @@ func transformNetworksToJSON(networks config.Networks) jsonNetworks {
...
 		}
 	}
 
 	return jsonNetworks
 }
 
 type jsonNetwork struct {
-	Host string
+	Host       string
+	NetworkKey string
```
jordanschalm, line 62: I might be missing some complexity here, but for this case at least it seems like we can just add `json` tags to the struct, then wouldn't need the custom `Marshal`/`Unmarshal` methods. 
```suggestion
	Host       string `json:"host"`
	NetworkKey string `json:"network-key"`
```

pkg/flowkit/config/json/network_test.go
```diff
@@ -26,7 +26,10 @@ import (
 
 func Test_ConfigNetworkSimple(t *testing.T) {
 	b := []byte(`{
-    "testnet": "access.testnet.nodes.onflow.org:9000"
+    "testnet": {
+			"host": "access.testnet.nodes.onflow.org:9000",
+			"network-key": ""
```
jordanschalm, line 31: Could we add a test case for if we're using the complex format but `network-key` is fully omitted?
```
"a-network": {
     "host": "access.testnet.nodes.onflow.org:9000"
}
```

pkg/flowkit/config/json/network_test.go
```diff
@@ -83,14 +92,7 @@ func Test_TransformNetworkToJSON(t *testing.T) {
 }
 
 func Test_IngoreOldFormat(t *testing.T) {
```
jordanschalm, line 94:
```suggestion
func Test_IgnoreOldFormat(t *testing.T) {
```

pkg/flowkit/output/prompt.go
```diff
@@ -157,6 +157,21 @@ func namePrompt() string {
 	return name
 }
 
+func secureNetworkKeyPrompt() string {
+	networkKeyPrompt := promptui.Prompt{
+		Label: "Enter host network key",
+		Validate: func(s string) error {
+			return nil
```
jordanschalm, line 164: Can we make sure it is a valid network key here? (attempt to parse as a hex-encoded ECDSA public key)

jordanschalm (commented):
Nice addition. 

Added a few notes and I'll take another pass once Gregor's suggestion's are in.

jordanschalm:
Just noticed that `flow snapshot save` will panic if no args are provided. Could you change that to print an error message while you're making changes here @kc1116 ?

```
 ./flow snapshot save -n mainnet
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/onflow/flow-cli/internal/snapshot.save(0xc000266060, 0x0, 0x2, 0x5957230, 0xc000251850, 0x0, 0x0, 0x5564344, 0x4, 0x0, ...)
	github.com/onflow/flow-cli/internal/snapshot/save.go:48 +0x339
github.com/onflow/flow-cli/internal/command.Command.AddToParent.func1(0x63fb780, 0xc000266060, 0x0, 0x2)
	github.com/onflow/flow-cli/internal/command/command.go:110 +0x5e5
github.com/spf13/cobra.(*Command).execute(0x63fb780, 0xc00003c0d0, 0x2, 0x2, 0x63fb780, 0xc00003c0d0)
	github.com/spf13/cobra@v1.3.0/command.go:860 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003a3180, 0xc00080fed0, 0x1, 0x1)
	github.com/spf13/cobra@v1.3.0/command.go:974 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.3.0/command.go:902
main.main()
	github.com/onflow/flow-cli/cmd/flow/main.go:78 +0x5c5
```

pkg/flowkit/config/json/network.go
sideninja: It's also a bug here since unmarshaling to `map[string]string` will always succeed which means advanced schema will be ignored. @jordanschalm is correct, tags could just be added (this code is pretty old), but at the same time like I mentioned you should add this key to the [advanced schema](https://github.com/onflow/flow-cli/blob/8ac4778e6653594cae8d94ba2c15ee01987b42e6/pkg/flowkit/config/json/network.go#L66), so it should just add another key and try [parse it here](https://github.com/onflow/flow-cli/blob/8ac4778e6653594cae8d94ba2c15ee01987b42e6/pkg/flowkit/config/json/network.go#L80).

pkg/flowkit/output/prompt.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/0e29e14541a6f2d948b2357a9b7ff9df72b5330b

pkg/flowkit/config/json/network_test.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/0e29e14541a6f2d948b2357a9b7ff9df72b5330b

pkg/flowkit/config/json/network_test.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/0e29e14541a6f2d948b2357a9b7ff9df72b5330b

pkg/flowkit/config/json/network.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/0e29e14541a6f2d948b2357a9b7ff9df72b5330b

pkg/flowkit/config/json/network_test.go
```diff
@@ -47,13 +47,10 @@ func Test_ConfigNetworkSimple(t *testing.T) {
...
-    		"emulator": {
-				"host": "127.0.0.1:3569",
-				"network-key": ""
-			},
+    		"emulator": "127.0.0.1:3569",
     		"testnet": {
-				"host": "access.testnet.nodes.onflow.org:9000",
-				"network-key": ""
+				"Host": "access.testnet.nodes.onflow.org:9000",
+				"NetworkKey": "5000676131ad3e22d853a3f75a5b5d0db4236d08dd6612e2baad771014b5266a242bccecc3522ff7207ac357dbe4f225c709d9b273ac484fed5d13976a39bdcd"
```
jordanschalm, line 53: We should still use lowercase field names to be consistent with the rest of the config (can use eg. `json:"host"` struct tags for this).

jordanschalm (commented):
One last comment on config field name consistency

pkg/flowkit/config/json/network_test.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/2e35af800a3cefd8b912866f96951c5da2e4a326

docs/configuration.md
```diff
@@ -94,7 +94,11 @@ We'll walk through each property one by one.
   "networks": {
     "emulator": "127.0.0.1:3569",
     "mainnet": "access.mainnet.nodes.onflow.org:9000",
-    "testnet": "access.devnet.nodes.onflow.org:9000"
+    "testnet": "access.devnet.nodes.onflow.org:9000",
+    "testnetSecure": {
+      "Host": "access-001.devnet30.nodes.onflow.org:9001",
+      "NetworkKey": "ba69f7d2e82b9edf25b103c195cd371cf0cc047ef8884a9bbe331e62982d46daeebf836f7445a2ac16741013b192959d8ad26998aff12f2adc67a99e1eb2988d"
```
janezpodhostnik, line 100:
```suggestion
      "host": "access-001.devnet30.nodes.onflow.org:9001",
      "network-key": "ba69f7d2e82b9edf25b103c195cd371cf0cc047ef8884a9bbe331e62982d46daeebf836f7445a2ac16741013b192959d8ad26998aff12f2adc67a99e1eb2988d"
```

Should this be lowercased now?

docs/configuration.md
```diff
@@ -291,13 +295,28 @@ Format for networks is:
 ...
 ```
 
+```json
+...
+"networks": {
+  "NETWORK NAME": {
+    "Host": "ADDRESS",
+    "NetworkKey": "ACCESS NODE NETWORK KEY"
```
janezpodhostnik, line 303: Should this be lowercased now?

docs/configuration.md
```diff
@@ -291,13 +295,28 @@ Format for networks is:
...
 ```json
 ...
 
 "networks": {
     "emulator": "127.0.0.1:3569",
     "mainnet": "access.mainnet.nodes.onflow.org:9000",
-    "testnet": "access.devnet.nodes.onflow.org:9000"
+    "testnet": "access.devnet.nodes.onflow.org:9000",
+    "testnetSecure": {
+        "Host": "access-001.devnet30.nodes.onflow.org:9001",
```
janezpodhostnik, line 317: Should this be lowercased now?

pkg/flowkit/config/json/network.go
```diff
@@ -28,60 +29,102 @@ type jsonNetworks map[string]jsonNetwork
...
+
+type simpleNetwork struct {
+	Host string `json:"host"`
 }
 
 type advancedNetwork struct {
-	Host  string
-	Chain string
+	Host       string `json:"host"`
+	NetworkKey string `json:"key"`
```
janezpodhostnik, line 102: In some location I'm seeing `host,key` but in some locations I'm seeing `host,network-key`. Is this ok?

jordanschalm (approved):

docs/configuration.md
kc1116: They all should be "host", "key" . I am updating the docs.

pkg/flowkit/config/json/network.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/05b28f3a8666def59064eebb64f01d7102dc1e6c

docs/configuration.md
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/05b28f3a8666def59064eebb64f01d7102dc1e6c

docs/configuration.md
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/05b28f3a8666def59064eebb64f01d7102dc1e6c

janezpodhostnik (approved):

docs/configuration.md
```diff
@@ -291,13 +295,28 @@ Format for networks is:
 ...
 ```
 
+```json
+...
+"networks": {
+  "NETWORK NAME": {
+    "host": "ADDRESS",
+    "key": "ACCESS NODE NETWORK KEY"
```
sideninja, line 303: wait so this is key or network-key? if it's key then that's good

internal/command/command.go
```diff
@@ -146,34 +149,35 @@ func createGateway(host string) (gateway.Gateway, error) {
 // 2. if conf is initialized return host by network flag
 // 3. if conf is not initialized and network flag is provided resolve to coded value for that network
 // 4. default to emulator network
-func resolveHost(state *flowkit.State, hostFlag string, networkFlag string) (string, error) {
+func resolveHost(state *flowkit.State, hostFlag, hostNetworkKeyFlag, networkFlag string) (string, string, error) {
```
sideninja, line 152: Is the `hostNetworkKeyFlag` even needed here? judging by method impl it doesn't do anything but return it, so maybe consider removing that.

internal/snapshot/save.go
```diff
@@ -34,6 +34,7 @@ var SaveCommand = &command.Command{
 		Use:     "save",
```
sideninja, line 34:
```suggestion
		Use:     "save <path>",
```

docs/configuration.md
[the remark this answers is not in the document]
kc1116: I missed this one, but yes in all configuration it will be "host" and "key"

pkg/flowkit/config/json/network.go
```diff
@@ -28,60 +29,102 @@ type jsonNetworks map[string]jsonNetwork
...
+		Advanced: advancedNetwork{
+			Host:       n.Host,
+			NetworkKey: n.NetworkKey,
+		},
+	}
+}
+
 type jsonNetwork struct {
-	Host string
+	Simple   simpleNetwork
```
sideninja, line 92: nice 👍

pkg/flowkit/config/json/network_test.go
sideninja: this format btw should be converted to `"a-netowrk": "access.testnet.nodes.onflow.org:9000"` since if you are only providing data needed by simple schema you should actually provide simple schema.

pkg/flowkit/config/loader_test.go
```diff
@@ -41,7 +41,10 @@ func Test_JSONSimple(t *testing.T) {
 		},
 		"contracts": {},
 		"networks": {
-			"emulator": "127.0.0.1:3569"
```
sideninja, line 44: what is the reason for changing this tests?

pkg/flowkit/config/json/network.go
```diff
@@ -28,60 +29,102 @@ type jsonNetworks map[string]jsonNetwork
...
 // transformToConfig transforms json structures to config structure.
 func (j jsonNetworks) transformToConfig() (config.Networks, error) {
-	var networks config.Networks
+	networks := make(config.Networks, 0)
 
 	for networkName, n := range j {
-		network := config.Network{
-			Name: networkName,
-			Host: n.Host,
+		if n.Advanced.NetworkKey != "" && n.Advanced.Host != "" {
```
sideninja, line 35: if an advanced schema is passed it should contain both values else it should be a simple format.

pkg/flowkit/config/network.go
```diff
@@ -26,8 +26,9 @@ type Networks []Network
 
 // Network defines the configuration for a Flow network.
 type Network struct {
-	Name string
-	Host string
+	Name       string
+	Host       string
+	NetworkKey string
```
sideninja, line 31: nit: `Network.NetowrkKey` feels redundant in naming, it should just be: `Network.Key`

pkg/flowkit/gateway/grpc.go
```diff
@@ -57,8 +61,34 @@ func NewGrpcGateway(host string) (*GrpcGateway, error) {
...
-		client: gClient,
-		ctx:    ctx,
+		client:       gClient,
+		ctx:          ctx,
+		secureClient: false,
+	}, nil
+}
+
+// NewSecureGrpcGateway returns a new gRPC gateway with a secure client connection.
+func NewSecureGrpcGateway(host, hostNetworkKey string) (*GrpcGateway, error) {
```
sideninja, line 71: Awesome! :)

pkg/flowkit/services/snapshot.go
```diff
@@ -50,6 +50,10 @@ func NewSnapshot(
 func (s *Snapshot) GetLatestProtocolStateSnapshot() ([]byte, error) {
 	s.logger.StartProgress("Downloading protocol snapshot...")
 
+	if !s.gateway.SecureConnection() {
+		s.logger.Info("warning: using insecure client connection to download snapshot, you should use a secure network configuration...")
```
sideninja, line 54: Nice!

pkg/flowkit/config/json/network.go
kc1116: User can possibly set advanced.network key but not advanced.host.

sideninja (changes requested):
Left some technical comments, the biggest issue is that default network schema should stay `host: {string}` and it shouldn't be changed, I see you changed all the tests to have default network as `host: {string}, key: {string}` which is the advance schema. Also make sure to manually test how this change works with the previous flow.json format, does it load it ok and does it change it on saving, at the same time newly generated flow.json should look the same as the older version. Because you were changing tests I'm confused whether that is currently true.

pkg/flowkit/config/json/network.go
sideninja: but what would then the host value be? that is the case where we need to show error right? since they are passing only the key but we don't know where we should connect

pkg/flowkit/config/loader_test.go
kc1116: Reverted the changes for simple config testing since we have test that do multiple network config

pkg/flowkit/config/json/network.go
kc1116: Sounds good I will require advanced config to be complete

pkg/flowkit/config/network.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/b94433142876497e2697c8cb40ecc8b135bfaa70

pkg/flowkit/config/json/network.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/b94433142876497e2697c8cb40ecc8b135bfaa70

pkg/flowkit/config/loader_test.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/b94433142876497e2697c8cb40ecc8b135bfaa70

internal/command/command.go
kc1116: I added validation of the --network-key flag here

https://github.com/onflow/flow-cli/pull/449/commits/51177595cb61d92a32420cd746544b6dfe485cc6

pkg/flowkit/services/snapshot.go
kc1116: https://github.com/onflow/flow-cli/pull/449/commits/6c4aa75a88a8648e4527b3d4b378c395586af695

kc1116:
> Left some technical comments, the biggest issue is that default network schema should stay `host: {string}` and it shouldn't be changed, I see you changed all the tests to have default network as `host: {string}, key: {string}` which is the advance schema. Also make sure to manually test how this change works with the previous flow.json format, does it load it ok and does it change it on saving, at the same time newly generated flow.json should look the same as the older version. Because you were changing tests I'm confused whether that is currently true.

You can confirm this by looking at the [tests](https://github.com/onflow/flow-cli/pull/449/files?file-filters%5B%5D=.go&show-viewed-files=true#diff-bfa5750ab78cfcbd108e25c7cfb1aa9cfae455b32d81c1dd55cc331491726c89) 

Also you can pull down this branch and do the following 
1. init configuration
2. add network to config by prompt
3. add simple network to config by prompt (when prompted for network key leave blank)

Is there something I am missing ?

sideninja (approved):
LGTM 🚀

PR #453, opened 2022-01-29
Show more user-friendly error message if proposer key is missing

pavel-main:
Closes #???

## Description

Hi guys,

Was playing around with Flow tutorial from [NFT School](https://nftschool.dev/tutorial/flow-nft-marketplace) and encountered an error when trying to override a signer using `--signer` flag:

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/onflow/flow-cli/pkg/flowkit.(*Transaction).SetProposer(0xc000174c00, 0xc0000e8428, 0xc0005d1928)
        github.com/onflow/flow-cli/pkg/flowkit/transaction.go:208 +0xb5
github.com/onflow/flow-cli/pkg/flowkit/services.(*Transactions).Build(0xc00032d7a0, {0x17, 0x9b, 0x6b, 0x1c, 0xb6, 0x75, 0x5e, 0x31}, {0xc0005d1a98, ...}, ...)
        github.com/onflow/flow-cli/pkg/flowkit/services/transactions.go:104 +0x1a5
github.com/onflow/flow-cli/pkg/flowkit/services.(*Transactions).Send(0xc00032d7a0, 0xc0000a2b68, {0xc0000d3800, 0x4ea, 0x6ea}, {0x7ffeefbff53f, 0x20}, 0xc000046218, {0x5b21430, 0x0, ...}, ...)
        github.com/onflow/flow-cli/pkg/flowkit/services/transactions.go:216 +0x1c6
github.com/onflow/flow-cli/internal/transactions.send({0xc00032cd20, 0x1, 0x3}, {0x51a4a88, 0xc0003346a0}, {{0x0, 0x0}, {0x4ecf009, 0x4}, {0x0, ...}, ...}, ...)
        github.com/onflow/flow-cli/internal/transactions/send.go:89 +0x312
github.com/onflow/flow-cli/internal/command.Command.AddToParent.func1(0x5add9c0, {0xc00032cd20, 0x1, 0x3})
        github.com/onflow/flow-cli/internal/command/command.go:116 +0x315
github.com/spf13/cobra.(*Command).execute(0x5add9c0, {0xc00003c210, 0x3, 0x3})
        github.com/spf13/cobra@v1.1.3/command.go:856 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0xc0001fd180)
        github.com/spf13/cobra@v1.1.3/command.go:960 +0x3ad
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.1.3/command.go:897
main.main()
        github.com/onflow/flow-cli/cmd/flow/main.go:69 +0x319
```

Turns out I've created an account with public key which simply didn't exist in emulator keychain (a copy-paste miss). This minor change returns an error message in that case instead of generic panic dump:

```
➜  petstore git:(master) ✗ flow transactions send src/flow/transaction/MintToken.cdc '{"name": "Max", "breed": "Bulldog"}' --signer test-account

❌ Command Error: failed to retrieve proposer key at index 0
```

______

For contributor use:

### DIFF
--- a/pkg/flowkit/services/accounts.go
+++ b/pkg/flowkit/services/accounts.go
@@ -428,8 +428,7 @@ func (a *Accounts) prepareTransaction(
 
 	tx.SetBlockReference(block)
 
-	tx, err = tx.SetProposer(proposer, account.Key().Index())
-	if err != nil {
+	if err = tx.SetProposer(proposer, account.Key().Index()); err != nil {
 		return nil, err
 	}
 
--- a/pkg/flowkit/services/accounts.go
+++ b/pkg/flowkit/services/accounts.go
@@ -426,8 +426,12 @@ func (a *Accounts) prepareTransaction(
 		return nil, err
 	}
 
-	tx.SetBlockReference(block).
-		SetProposer(proposer, account.Key().Index())
+	tx.SetBlockReference(block)
+
+	tx, err = tx.SetProposer(proposer, account.Key().Index())
+	if err != nil {
+		return nil, err
+	}
 
 	tx, err = tx.Sign()
 	if err != nil {
--- a/pkg/flowkit/services/project.go
+++ b/pkg/flowkit/services/project.go
@@ -207,8 +207,7 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
 
 		tx.SetBlockReference(block)
 
-		tx, err = tx.SetProposer(targetAccountInfo, targetAccount.Key().Index())
-		if err != nil {
+		if err = tx.SetProposer(targetAccountInfo, targetAccount.Key().Index()); err != nil {
 			return nil, err
 		}
 
--- a/pkg/flowkit/services/project.go
+++ b/pkg/flowkit/services/project.go
@@ -205,8 +205,12 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
 			}
 		}
 
-		tx.SetBlockReference(block).
-			SetProposer(targetAccountInfo, targetAccount.Key().Index())
+		tx.SetBlockReference(block)
+
+		tx, err = tx.SetProposer(targetAccountInfo, targetAccount.Key().Index())
+		if err != nil {
+			return nil, err
+		}
 
 		tx, err = tx.Sign()
 		if err != nil {
--- a/pkg/flowkit/services/transactions.go
+++ b/pkg/flowkit/services/transactions.go
@@ -99,6 +99,10 @@ func (t *Transactions) Build(
 	if err != nil {
 		return nil, err
 	}
+	
+	if len(proposerAccount.Keys) <= proposerKeyIndex {
+		return nil, fmt.Errorf("failed to retrieve proposer key at index %d", proposerKeyIndex)
+	}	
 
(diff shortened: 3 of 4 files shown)


sideninja (changes requested):
Thank you for this fix, definitely good addition but I believe it would be better if that check would be added on the transaction method itself https://github.com/onflow/flow-cli/blob/8cc82c2f4a68aa39409611c0b7926412996f38ce/pkg/flowkit/transaction.go#L236
so the method should return an error as well.

pavel-main:
@sideninja updated 👌🏻

janezpodhostnik (approved):
Nice addition!

I have a small comment/question:

Originally the SetProposer could be used in a method chain, where it made sense that it returns a  `*Transaction`.

Now that it also returns an error it cannot be used in a method chain, so returning that `*Transaction` does not seem to make much sense anymore. Should it be removed, so that the method only returns an error? What do you think @sideninja? 

This might be a  more general question, as I see this on other methods, not just the refactored SetProposer method, so maybe this is a separate issue?

pavel-main:
@janezpodhostnik In theory, still could be chained by wrapping it into another type, e.g.:

```go
type TransactionResult struct {
    Tx *Transaction
    Error error
}
```

^ However it's gonna be easy to forget checking if `Error` is not nil

sideninja:
> Nice addition!
> 
> I have a small comment/question:
> 
> Originally the SetProposer could be used in a method chain, where it made sense that it returns a `*Transaction`.
> 
> Now that it also returns an error it cannot be used in a method chain, so returning that `*Transaction` does not seem to make much sense anymore. Should it be removed, so that the method only returns an error? What do you think @sideninja?
> 
> This might be a more general question, as I see this on other methods, not just the refactored SetProposer method, so maybe this is a separate issue?

Yes, good point. It should be like that for sure and the fact it isn't is another issue. I plan to do a small refactor PR and that would be one of the points to improve.

sideninja:
> @janezpodhostnik In theory, still could be chained by wrapping it into another type, e.g.:
> 
> ```go
> type TransactionResult struct {
>     Tx *Transaction
>     Error error
> }
> ```
> 
> ^ However it's gonna be easy to forget checking if `Error` is not nil

That's true, but like you said maybe it's easier to forget then, so if you make it just returns an error it will be more consistent with other changes. Sorry if we are a bit nit about it.

pavel-main:
@sideninja @janezpodhostnik No worries, in fact this whole discussion was the reason of the original proposal - to keep it simple, short, user-friendly and without a need to refactor most of the methods. Would you guys reconsider accepting the original commit?

sideninja:
> @sideninja @janezpodhostnik No worries, in fact this whole discussion was the reason of the original proposal - to keep it simple, short, user-friendly and without a need to refactor most of the methods. Would you guys reconsider accepting the original commit?

I would keep the @janezpodhostnik comment and just make that change, other APIs will be changed later, but if know what is best we should do it even if other APIs are not yet updated, so we are going in the right direction.

pavel-main:
@sideninja updated

sideninja (approved):
LGTM thank you

PR #454, opened 2022-01-31
Crash Reports

sideninja:
## Description
This PR implements sentry as a crash reporting metric. This way we will see all the crashes that are happening on the Flow CLI if the users agree with the report as seen in the video.

https://asciinema.org/a/RJwzOqrrsFUROWCasgMQKr6tW

______

For contributor use:


sideninja:
API keys need to be replaced after I get them generated from the SRE.

sideninja:
> Nice!
> 
> A nice improvement would be to have an `always send` and `never send` option that gets set to the config (or somewhere persistent). But that can be added later.

This case is a bit paradoxical :D because we are improving the UX of the crash reports which greatly hurts UX in the first place maybe better to make sure this screen is not displayed enough times that this setting will make it better :D But I didn't want to create this file/setting for just the sake of this, it feels a bit invasive, but I guess if later we will expand this to other metrics it will be unavoidable, but even then I would maybe first consider saving that preference in the cloud as an anonymous setting, sentry and similar allow saving preferences.

PR #465, opened 2022-02-21
Arguments parsing fix for Transactions with functions

bluesign:


pkg/flowkit/arguments.go
```diff
@@ -132,20 +132,21 @@ func ParseArgumentsWithoutType(fileName string, code []byte, args []string) (scr
...
-
 	functionDeclaration := sema.FunctionEntryPointDeclaration(program)
 	if functionDeclaration != nil {
 		if functionDeclaration.ParameterList != nil {
 			parameterList = functionDeclaration.ParameterList.Parameters
 		}
 	}
 
+	transactionDeclaration := program.TransactionDeclarations()
+	fmt.Println(transactionDeclaration)
```
sideninja, line 143:
```suggestion
```

sideninja:
This is so weird, I remember fixing this, but don't understand where it is.

bluesign:
@sideninja I felt deja vu too, I also somehow remember this was fixed.

sideninja:
Might bo some regression at some point, will take a look later, but thank you for fixing this (again).

pkg/flowkit/arguments.go
```diff
@@ -132,20 +132,20 @@ func ParseArgumentsWithoutType(fileName string, code []byte, args []string) (scr
...
-
 	functionDeclaration := sema.FunctionEntryPointDeclaration(program)
 	if functionDeclaration != nil {
 		if functionDeclaration.ParameterList != nil {
 			parameterList = functionDeclaration.ParameterList.Parameters
 		}
 	}
 
+	transactionDeclaration := program.TransactionDeclarations()
+	if len(transactionDeclaration) == 1 {
```
sideninja, line 143: If we are restricting transaction declaration per file (which I agree might be good idea) we should return an error in case there are more which says something like "only one transaction declaration in a source file allowed"

bluesign: Actually is it possible to declare 2 ?

sideninja: Actually an interesting question, but no matter the answer (which I intend to now find out, although I would say no) you could argue to have multiple of them defined in a file for whatever reason to then only have one of them sent to network, in any case we would need to return an error IMHO whatever the case is right, so the user can now why it didn't work

sideninja: I mean the signed transaction can't really contain more than one right, since the signature of payload would then be invalid and AN would also not pass the validation.

bluesign: Yeah but if somehow we can define somehow multiple transactions in a file ( like function overloading ) or init overloading, it can be nice to detect with arguments and match the transaction.

bluesign: They can contain I guess. Payload will contain all transactions ( if it is allowed ) Arguments etc will not be related. So technically you can sign ( in case of overloading )

sideninja: I see what you mean. That could be a nice feature, but, I'm not sure we want to promote this kind of patterns, right now we promote having a transactions folder containing transactions in files, and although I think we at the same time shouldn't enforce this, making alternatives possible, it could be too early to allow alternatives before some base patterns are established. This is just my opinion tho. And yes you are right about anyhow AN returning an error but it will lack more context around the error which is available in the CLI, and I like the philosophy of failing early when you can, but also not making requests that will with certainty fail just to make the client a bit lighter.

sideninja: > Payload will contain all transactions ( if it is allowed )
I don't believe it is

sideninja: I plan to release a new version today or tomorrow and it would be nice to have this included, so if you find the time to just handle an else case for this that would be great :) else let me know I can help with this.

bluesign: Hey, I am on vacation right now till the end of the week, would be great if you can add that.

sideninja: For sure, enjoy your vacation :)

sideninja (approved):

PR #491, opened 2022-05-02
Fix associated account with same address bug

sukantoraymond:
Closes #???

## Description

______

For contributor use:

### DIFF
--- a/pkg/flowkit/account.go
+++ b/pkg/flowkit/account.go
@@ -163,6 +163,17 @@ func (a Accounts) ByAddress(address flow.Address) (*Account, error) {
 	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
 }
 
+// ByAddress get an account by address.
+func (a Accounts) ByAddressAndAccountName(address flow.Address, accountName string) (*Account, error) {
+	for i := range a {
+		if a[i].address == address && a[i].name == accountName {
+			return &a[i], nil
+		}
+	}
+
+	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
+}
+
 // ByName get an account by name or returns and error if no account found
 func (a Accounts) ByName(name string) (*Account, error) {
 	for i := range a {
--- a/pkg/flowkit/services/project.go
+++ b/pkg/flowkit/services/project.go
@@ -127,6 +127,7 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
 			contract.Name,
 			contract.Source,
 			contract.Target,
+			contract.AccountName,
 			contract.Args,
 		)
 		if err != nil {
@@ -160,7 +161,7 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
 			return nil, err
 		}
 
-		targetAccount, err := p.state.Accounts().ByAddress(contract.Target())
+		targetAccount, err := p.state.Accounts().ByAddressAndAccountName(contract.Target(), contract.AccountName())
 
 		if err != nil {
 			return nil, fmt.Errorf("target account for deploying contract not found in configuration")
--- a/pkg/flowkit/state.go
+++ b/pkg/flowkit/state.go
@@ -42,10 +42,11 @@ type ReaderWriter interface {
 
 // Contract is a Cadence contract definition for a project.
 type Contract struct {
-	Name   string
-	Source string
-	Target flow.Address
-	Args   []cadence.Value
+	Name        string
+	Source      string
+	Target      flow.Address
+	AccountName string
+	Args        []cadence.Value
 }
 
 // State manages the state for a Flow project.
@@ -192,10 +193,11 @@ func (p *State) DeploymentContractsByNetwork(network string) ([]Contract, error)
 			}
 
 			contract := Contract{
-				Name:   c.Name,
-				Source: path.Clean(c.Source),
-				Target: account.address,
-				Args:   deploymentContract.Args,
+				Name:        c.Name,
+				Source:      path.Clean(c.Source),
+				Target:      account.address,
+				AccountName: account.name,
+				Args:        deploymentContract.Args,
 			}
 
 			contracts = append(contracts, contract)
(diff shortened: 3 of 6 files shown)


pkg/flowkit/services/project.go
```diff
@@ -160,7 +161,7 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
 			return nil, err
 		}
 
-		targetAccount, err := p.state.Accounts().ByAddress(contract.Target())
+		targetAccount, err := p.state.Accounts().ByAddressAndAccountName(contract.Target(), contract.AccountName())
```
sukantoraymond, line 164: now that i think about it, we can just use existing function of ByName instead of filtering by both address and account name

pkg/flowkit/state_test.go
```diff
@@ -151,7 +151,109 @@ func generateComplexProject() State {
...
+		Networks: config.Networks{{
+			Name: "emulator",
+			Host: "127.0.0.1.3569",
+		}},
+	}
+
+	p, err := newProject(&config, composer, af)
+	if err != nil {
+		fmt.Println(err)
+	}
```
sukantoraymond, line 253: This tests the new code against the sample input that caused the issue.

pkg/flowkit/state_test.go
```diff
@@ -526,7 +643,7 @@ func Test_ChangingState(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, (*pkey).String(), pk.String())
 
-	bar, err := p.Accounts().ByAddress(flow.HexToAddress("0x1"))
+	bar, err := p.Accounts().ByAddressAndAccountName(flow.HexToAddress("0x1"), "foo")
```
sukantoraymond, line 646: will need to delete existing byaddress tests since it is no longer used, if this PR is merged.

pkg/flowkit/account.go
```diff
@@ -163,6 +163,17 @@ func (a Accounts) ByAddress(address flow.Address) (*Account, error) {
 	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
 }
 
+// ByAddress get an account by address.
```
sideninja, line 166:
```suggestion
// ByAddressAndAccountName get an account by address and its name.
```

pkg/flowkit/services/project.go
sideninja: I'm now wondering the same. I think that accounts must be unique by their names (restriction should be added to the config module) and hence adding the address is unnecessary. I believe the code would be nicer if it just gets the accounts by name.

sideninja (approved):
LGTM, good work

pkg/flowkit/state_test.go
```diff
@@ -351,7 +453,12 @@ func Test_AccountByAddressSimple(t *testing.T) {
 
 	assert.Equal(t, acc.name, "emulator-account")
 }
+func Test_AccountByAddressAndNameSimple(t *testing.T) {
+	p := generateSimpleProject()
+	acc, _ := p.Accounts().ByName("emulator-account")
 
+	assert.Equal(t, acc.name, "emulator-account")
+}
```
psiemens, line 461: This looks like it's almost a duplicate of the test case below `Test_AccountByNameSimple`. Can they be combined?

`Test_AccountByAddressAndNameSimple` doesn't quite fit anymore after you renamed the method to `ByName`

pkg/flowkit/state_test.go
```diff
@@ -436,7 +543,17 @@ func Test_AccountByAddressComplex(t *testing.T) {
 	assert.Equal(t, acc1.name, "account-4")
 	assert.Equal(t, acc2.name, "account-2")
 }
+func Test_AccountByAddressAndAccountNameComplex(t *testing.T) {
```
psiemens, line 546:
```suggestion
func Test_AccountByNameWithDuplicateAddress(t *testing.T) {
```

psiemens (commented):
Nice stuff! 💯

pkg/flowkit/services/project.go
sukantoraymond: resolved

pkg/flowkit/state_test.go
sukantoraymond: addressed

pkg/flowkit/state_test.go
```diff
@@ -151,7 +151,109 @@ func generateComplexProject() State {
...
+		}, {
+			Network: "emulator",
+			Account: "account-4",
+			Contracts: []config.ContractDeployment{
+				{Name: "FungibleToken", Args: nil},
+				{Name: "NonFungibleToken", Args: nil},
+				{Name: "Kibble", Args: nil},
+				{Name: "KittyItems", Args: nil},
+				{Name: "KittyItemsMarket", Args: nil},
+			},
```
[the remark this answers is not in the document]
sukantoraymond: good point. removing

pkg/flowkit/state_test.go
sukantoraymond: addressed

pkg/flowkit/state.go
```diff
@@ -42,10 +42,11 @@ type ReaderWriter interface {
 
 // Contract is a Cadence contract definition for a project.
 type Contract struct {
-	Name   string
-	Source string
-	Target flow.Address
-	Args   []cadence.Value
+	Name        string
+	Source      string
+	Target      flow.Address
```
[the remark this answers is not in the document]
sukantoraymond: Good point. addressed

sideninja (commented):
There is a test case missing that tests the change. You should add a test to the `flowkit/services/project_test` that has two accounts with same addresses and different names and try to call deploy (follow other tests in that suite).

pkg/flowkit/state_test.go
```diff
@@ -151,7 +151,99 @@ func generateComplexProject() State {
 
 	return *p
 }
+func generateComplexProjectSameAddress() State {
```
sideninja, line 154: You don't need to duplicate this necessarily, you can just put another account with same address and different name under `generateComplexProject`.

sideninja:
@sukantoraymond how come this was closed?

pkg/flowkit/state_test.go
sukantoraymond: addressed

PR #494, opened 2022-05-05
expose emulator log on flowkit gateway

### LINKED ISSUE #325: Make it possible to output emulator log when using Emulator gateway
## Instructions

When using the emulator gateway sometimes it is desired to actually see the log. Currently i find no way of doing this, it would be nice if there was an option for this.

### Issue To Be Solved

Go-with-the-flow is beeing rewritten untop of the improvements in flowkit and it would be very nice to be able to see the log of the emulator weaved into the ouput log of the tool.

### (Optional): Context

This is the branch of go-with-the-flow that has the new draft code https://github.com/bjartek/go-with-the-flow/tree/main

`cd examples && go run demo/main.go` will print out the log messages from go-with-the-flow but not from the emulator

bluesign:
closes #325

only issue currently is error messages from emulator are wrapped via Status ( from google grpc ) like below

"rpc error: code = NotFound desc = could not find account with address 0000000000000001"

instead

"could not find account with address 0000000000000001"

### DIFF
--- a/pkg/flowkit/gateway/emulator.go
+++ b/pkg/flowkit/gateway/emulator.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/onflow/cadence"
 	emulator "github.com/onflow/flow-emulator"
+	"github.com/onflow/flow-emulator/convert/sdk"
+
 	"github.com/onflow/flow-emulator/server/backend"
 	"github.com/onflow/flow-go-sdk"
 	"github.com/onflow/flow-go-sdk/client"
@@ -148,20 +150,6 @@ func convertBlock(block *flowGo.Block) *flow.Block {
 	}
 }
 
-func convertEvents(events []flowGo.Event) []flow.Event {
-	convertedEvents := make([]flow.Event, 0)
-	for _, event := range events {
-		convertedEvents = append(convertedEvents, flow.Event{
-			Type:             string(event.Type),
-			TransactionID:    flow.Identifier(event.TransactionID),
-			TransactionIndex: int(event.TransactionIndex),
-			EventIndex:       int(event.EventIndex),
-			Payload:          event.Payload,
-		})
-	}
-	return convertedEvents
-}
-
 func (g *EmulatorGateway) GetEvents(
 	eventType string,
 	startHeight uint64,
@@ -189,7 +177,7 @@ func (g *EmulatorGateway) getBlockEvent(height uint64, eventType string) client.
 
 	for _, e := range events {
 		if e.BlockID == block.ID() {
-			result.Events = convertEvents(e.Events)
+			result.Events, _ = sdk.FlowEventsToSDK(e.Events) //convertEvents(e.Events)
 			return result
 		}
 	}
--- a/pkg/flowkit/gateway/emulator.go
+++ b/pkg/flowkit/gateway/emulator.go
@@ -20,10 +20,12 @@ package gateway
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/onflow/cadence"
 	emulator "github.com/onflow/flow-emulator"
 	"github.com/onflow/flow-emulator/convert/sdk"
+	"google.golang.org/grpc/status"
 
 	"github.com/onflow/flow-emulator/server/backend"
 	"github.com/onflow/flow-go-sdk"
@@ -41,6 +43,10 @@ type EmulatorGateway struct {
 	backend  *backend.Backend
 	ctx      context.Context
 	logger   *logrus.Logger
+}
+
+func UnwrapStatusError(err error) error {
+	return fmt.Errorf(status.Convert(err).Message())
 }
 
 func NewEmulatorGateway(serviceAccount *flowkit.Account) *EmulatorGateway {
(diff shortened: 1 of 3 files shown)


bjartek:
I used this version of flow-cli in a overflow branch and I do indeed get nested logs from this! Awesome!

pkg/flowkit/gateway/emulator.go
```diff
@@ -19,31 +19,47 @@
...
 	emulator *emulator.Blockchain
+	backend  *backend.Backend
+	ctx      context.Context
 }
 
 func NewEmulatorGateway(serviceAccount *flowkit.Account) *EmulatorGateway {
+	return NewEmulatorGatewayWithLogger(logrus.New(), serviceAccount)
+}
+
+func NewEmulatorGatewayWithLogger(logger *logrus.Logger, serviceAccount *flowkit.Account) *EmulatorGateway {
```
sideninja, line 47: Should we maybe have functional options as the last argument and just have a factory with options, so for example:
```go
NewEmulatorWithOpts(serviceAccount *flowkit.Account, ...func(*EmulatorGateway)) { ... }

NewEmulatorWithOpts(service, WithLogger(lorgus.new())
```
Just so if we in the future need to have more options.

sideninja: Also we can then have `WithContext` and no need for bellow method, although I don't think this is a good pattern in Go, context should be passed as first argument, and I know I did that mistake myself when starting (since I didn't know enough Go), I would like to refactor this before making flowkit more official, but this is a task for another PR

pkg/flowkit/gateway/emulator.go
```diff
@@ -19,31 +19,47 @@
...
+
 	return &EmulatorGateway{
-		emulator: newEmulator(serviceAccount),
+		emulator: emulator,
+		backend:  backend,
+		ctx:      context.Background(),
 	}
 }
 
+func (g *EmulatorGateway) SetContext(ctx context.Context) {
```
sideninja, line 59: Maybe not needed if with options is used

pkg/flowkit/gateway/emulator.go
```diff
@@ -65,71 +81,47 @@ func newEmulator(serviceAccount *flowkit.Account) *emulator.Blockchain {
...
-		return nil, fmt.Errorf("failed to submit transaction: %w", err)
-	}
-
-	_, err = g.emulator.CommitBlock()
-	if err != nil {
-		return nil, fmt.Errorf("failed to submit transaction: %w", err)
-	}
-
-	return t, nil
+	return tx.FlowTransaction(), g.backend.SendTransaction(context.Background(), *tx.FlowTransaction())
```
sideninja, line 88: I think it's good to be explicit in returning values:
```go
err := g.backend.SendTransaction(context.Background(), *tx.FlowTransaction())
if err != nil {
   return nil, err
}

return t, nil
```
Just because if error exists the developer might not check for it and just use the returned value (in this case it's returned regardless) which can lead to errors. I remember reading that somewhere as a good practice but not sure where. This is a bit nit tho.

pkg/flowkit/gateway/emulator.go
```diff
@@ -65,71 +81,47 @@ func newEmulator(serviceAccount *flowkit.Account) *emulator.Blockchain {
...
 }
 
 func (g *EmulatorGateway) GetTransaction(id flow.Identifier) (*flow.Transaction, error) {
-	return g.emulator.GetTransaction(id)
+	return g.backend.GetTransaction(g.ctx, id)
 }
 
 func (g *EmulatorGateway) Ping() error {
-	return nil
+	return g.backend.Ping(g.ctx)
```
sideninja, line 100: 🧡

pkg/flowkit/gateway/emulator.go
```diff
@@ -152,6 +144,20 @@ func convertBlock(block *flowGo.Block) *flow.Block {
 	}
 }
 
+func convertEvents(events []flowGo.Event) []flow.Event {
```
sideninja, line 147: I think you could use this method no? https://github.com/onflow/flow-emulator/blob/master/convert/sdk/flow.go#L217

sideninja:
Thank you a lot @bluesign this unblocks @bjartek for sure, I'm sorry I couldn't get to it sooner. 

I left some comments. On the error description I think it should be possible to unwrap the error to the grpc status error and then just return an error message, something like (not tested code):
```go
var s *status.Status
ok := errors.As(err, s)
return s.Error()
```

bjartek:
When i saw this i just _had_ to start testing. 
https://github.com/bjartek/overflow/blob/emulator-logs/overflow/setup.go#L176

I now get back computation used and debug log that I can add to Transaction result in the new API.

pkg/flowkit/gateway/emulator.go
bluesign: I see this `NewEmulator(service).WithLogger(logrus.new())` pattern used commonly but not sure which one is better tbh, but I can refactor to whichever we choose. 

I agree with context comment, actually I started with `WithContext`, then refactored it to `SetContext`, but to confess I am not too familiar with context there. I mean does contexts are has to be per call vs per lifecycle of the gateway.

pkg/flowkit/gateway/emulator.go
bluesign: I put this actually in line with my other comment, I mean I presumed most of the calls will not individually need context, but for example if you want to have timeout on let's say on script execution you can set a context before the call.

pkg/flowkit/gateway/emulator.go
bluesign: This is really good advice actually.

pkg/flowkit/gateway/emulator.go
bluesign: oh I totally missed this one, thanks

bluesign:
@sideninja 
> On the error description I think it should be possible to unwrap the error to the grpc status error and then just return an error message, something like (not tested code)

I think this is not possible `Status` is not the `error` here but actually it is generating some `Error` object, which is in internal package.

pkg/flowkit/gateway/emulator.go
sideninja: > I see this NewEmulator(service).WithLogger(logrus.new()) pattern used commonly but not sure which one is better tbh, but I can refactor to whichever we choose.

We can make it `.WithLogger()` as well as a method on the emulator, the only difference I would say is that if the option is needed during the initialization then I think it's better to have an option being passed to the factory method. I would say that for the logger it makes sense to be passed during the init but I can also see the case where you would just set it for the function you care about and then in a way disable it, but then also we could argue that's something that should be handled on the logger instance itself and verbosity settings etc not by setting it null. 

> I agree with the context comment, actually, I started with WithContext, then refactored it to SetContext, but to confess I am not too familiar with context there. I mean does contexts are has to be per call vs per lifecycle of the gateway.

In go we should pass context as the first argument, that's idiomatic go and the only reason this isn't done like that here is that when I was defining the interface for the gateway I didn't know better. I would want at some later point (maybe flowkit update we discuss) to make it correct.

sideninja:
> @sideninja
> 
> > On the error description I think it should be possible to unwrap the error to the grpc status error and then just return an error message, something like (not tested code)
> 
> I think this is not possible `Status` is not the `error` here but actually it is generating some `Error` object, which is in internal package.

You can convert it back. In Go error can be "upcasted/downcasted" to the more specific errors, I've checked and the status package even has a convenience error for it, here's an example that I tested works on the account test case:

```go
statusErr, ok := status.FromError(err)
assert.True(t, ok)
assert.Equal(t, statusErr.Message(), "could not find account with address 0000000000000001")
```

I think it might be a good idea to convert the error to just include the message instead of rpc specific message. We could do that by running:

```go
statusErr, ok := status.FromError(err)
if ok {
    return nil, fmt.Errorf(statusErr.Message())
}
```

sideninja (approved):
LGTM

PR #499, opened 2022-05-09
Bugfix/deploy account with name

### LINKED ISSUE #487: Flow CLI does not takes in the intended account on projects deploy
### Problem

When using Flow CLI command for project deployment. The account is not taken as intended, Instead it takes the first account with the same address for deployment for signing the transaction.

### Steps to Reproduce
Here is my flow.init set up:
```
	"accounts": {
		"emulator-account": {
			"key": "ae1b44c0f5e8f6992ef2348898a35e50a8b0b9684000da8b1dade1b3bcd6ebee"
		},
		"emulator-account-2": {
			"address": "4ada5fbd7073699b",
			"key": "Not shown as it is not important"
		},
		"testnet-account" : {
			"address": "4ada5fbd7073699b",
			"key": {
				"type" : "hex",
				"index" : 0,
				"signatureAlgorithm" : "ECDSA_secp256k1",
				"hashAlgorithm" : "SHA3_256",
				"privateKey" : "Not shown as it is not important",
				"publicKey" : "Not shown as it is not important"
			}
		},
```

And here are the contracts that I want to deploy : 
```
	"deployments": {
		"testnet" : {
			"testnet-account" : [
				"Voter", "ArleeScene", "Arlequin"
			]		
		}
	}
```

As you can see, I have the same account address for "emulator-account-2" and "testnet-account" 
So when I run flow project deploy -n=testnet

Errors returned : 
```
 Error deploying Voter: ([Error Code: 1006] invalid proposal key: public key 0 on account 4ada5fbd7073699b does not have a valid signature: [Error Code: 1009] invalid envelope key: public key 0 on account 4ada5fbd7073699b does not have a valid signature: signature is not valid)
 ```
 
 It might be the issue that, flow CLI is checking for the first address that come into sight which has the same address as the deploying contracts target address. Therefore the account info in "emulator-account-2" is taken instead of "testnet-account".
 
 After deletion of the "emulator-account-2" object, the transaction can then run and successfully deploy the contracts.
 
 Flow CLI should be programmed so that even I have a few accounts with the same address, it still uses the correct target account that I intended while running flow project deploy.

### PR DESCRIPTION
**Pull request by sukantoraymond**
Closes #487 

## Description
This PR changes how the account is found in flow configuration through account name instead of existing method of filtering by account address.

______

For contributor use:

### DIFF
--- a/pkg/flowkit/account.go
+++ b/pkg/flowkit/account.go
@@ -152,17 +152,6 @@ func (a *Accounts) Remove(name string) error {
 	return nil
 }
 
-// ByAddress get an account by address.
-func (a Accounts) ByAddress(address flow.Address) (*Account, error) {
-	for i := range a {
-		if a[i].address == address {
-			return &a[i], nil
-		}
-	}
-
-	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
-}
-
 // ByName get an account by name or returns and error if no account found
 func (a Accounts) ByName(name string) (*Account, error) {
 	for i := range a {
--- a/pkg/flowkit/account.go
+++ b/pkg/flowkit/account.go
@@ -163,17 +163,6 @@ func (a Accounts) ByAddress(address flow.Address) (*Account, error) {
 	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
 }
 
-// ByAddress get an account by address.
-func (a Accounts) ByAddressAndAccountName(address flow.Address, accountName string) (*Account, error) {
-	for i := range a {
-		if a[i].address == address && a[i].name == accountName {
-			return &a[i], nil
-		}
-	}
-
-	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
-}
-
 // ByName get an account by name or returns and error if no account found
 func (a Accounts) ByName(name string) (*Account, error) {
 	for i := range a {
--- a/pkg/flowkit/account.go
+++ b/pkg/flowkit/account.go
@@ -152,6 +152,17 @@ func (a *Accounts) Remove(name string) error {
 	return nil
 }
 
+// ByAddress get an account by address.
+func (a Accounts) ByAddress(address flow.Address) (*Account, error) {
+	for i := range a {
+		if a[i].address == address {
+			return &a[i], nil
+		}
+	}
+
+	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
+}
+
 // ByName get an account by name or returns and error if no account found
 func (a Accounts) ByName(name string) (*Account, error) {
 	for i := range a {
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: pkg/flowkit/account.go [hunk 1]
```diff
@@ -152,17 +152,6 @@ func (a *Accounts) Remove(name string) error {
...
-	for i := range a {
-		if a[i].address == address {
-			return &a[i], nil
-		}
-	}
-
-	return nil, fmt.Errorf("could not find account with address %s in the configuration", address)
-}
-
 // ByName get an account by name or returns and error if no account found
```
sukantoraymond, line 155: Since not used anymore, deleting ByAddress function and its tests

pkg/flowkit/services/project_test.go
```diff
@@ -111,6 +111,55 @@ func TestProject(t *testing.T) {
 		gw.Mock.AssertNumberOfCalls(t, tests.GetTransactionResultFunc, 1)
 	})
 
+	t.Run("Deploy Project Duplicate Address", func(t *testing.T) {
```
sukantoraymond, line 114: Added test for duplicate address based on previous PR comments

tests/resources.go
```diff
@@ -274,7 +274,9 @@ func Bob() *flowkit.Account {
 func Charlie() *flowkit.Account {
 	return newAccount("Charlie", "0x3", "seedseedseedseedseedseedseedseedseedseedseedseedCharlie")
 }
-
+func Donald() *flowkit.Account {
+	return newAccount("Donald", "0x3", "seedseedseedseedseedseedseedseedseedseedseedseedDonald")
+}
```
sukantoraymond, line 279: added a test account with same address to test get account with same address

pkg/flowkit/state_test.go
```diff
@@ -137,6 +137,16 @@ func generateComplexProject() State {
...
+		}, {
+			Name:    "emulator-account-2",
+			Address: flow.HexToAddress("2c1162386b0a245f"),
+			Key: config.AccountKey{
+				Type:       config.KeyTypeHex,
+				Index:      0,
+				SigAlgo:    crypto.ECDSA_P256,
+				HashAlgo:   crypto.SHA3_256,
+				PrivateKey: keys()[1],
+			},
```
sukantoraymond, line 149: Implemented additional account with same address, instead of creating another generatecomplexproject function

pkg/flowkit/account.go
sideninja: I would leave this function because the flowkit package is used as a library by others (one example https://github.com/bjartek/overflow). So it's a good utility function.

pkg/flowkit/contracts/contracts.go
```diff
@@ -39,6 +39,7 @@ type Contract struct {
 	name         string
 	source       string
 	target       flow.Address
```
sideninja, line 41: change this to `accountAddress` or change `accountName` to `targetName` either way as long as they are consistent.

sukantoraymond: addressed

pkg/flowkit/account.go
sukantoraymond: addressed

sideninja (approved):
Good job. Accepted

PR #512, opened 2022-06-01
Handle panic from event parsing argument types that causes cli to crash

### LINKED ISSUE #486: Event parsing error
CLI crashes when parsing certain events, this crash was reported using Sentry and needs to be further investigated and reproduced.

Error:
runtime.errorString github.com/onflow/cadence in OptionalType.ID

Crash reported for event parsing:
```
runtime.errorString: runtime error: invalid memory address or nil pointer dereference
  File "github.com/onflow/cadence@v0.23.0/types.go", line 71, in OptionalType.ID
  File "github.com/onflow/flow-cli/internal/events/events.go", line 131, in printField
  File "github.com/onflow/flow-cli/internal/events/events.go", line 124, in eventString
  File "github.com/onflow/flow-cli/internal/events/events.go", line 112, in eventsString
  File "github.com/onflow/flow-cli/internal/events/events.go", line 80, in (*EventResult).String
...
(6 additional frame(s) were not displayed)
```

sukantoraymond:
Closes #486

## Description

This PR is related to bug in cadence https://github.com/onflow/cadence/issues/1672, where events with argument of type of arrays will cause flow-cli to crash. This PR prevents crash in flow cli by handling the panic caused array type arguments in events.

______

For contributor use:


sukantoraymond:
Sample terminal result when trying to retrieve event with argument of array type and panic is handled
````
Raymonds-MBP-3:flow raymondsukanto$ ./flow events get A.f8d6e0586b0a20c7.HelloWorld.FooFooFooEvent
2022/06/01 17:56:55 panic occurred: runtime error: invalid memory address or nil pointer dereference

Events Block #8:
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooFooFooEvent
    Tx ID       53dc0d62152f1380c6619ae0eed9afa10ff06f105a028714d58bcb04c15d6706
    Values
                - x (): ["Event"]
                hex: 5b224576656e74225d 

````

File: internal/events/events.go [hunk 1]
```diff
@@ -126,12 +127,23 @@ func eventString(writer io.Writer, event flow.Event) {
...
+			_, _ = fmt.Fprintf(writer, "\t\t- %s (%s): %s \n", field.Identifier, typeId, v)
+			log.Println("panic occurred:", err)
+		}
+	}()
+
 	if field.Type != nil {
+		//TODO: onflow/cadence issue #1672
+		//currently getting ID for cadence array will cause panic
 		typeId = field.Type.ID()
 	}
```
sideninja, line 145:
```suggestion
		if field.Type != nil {
		// TODO: onflow/cadence issue #1672
		// currently getting ID for cadence array will cause panic
		if _, ok := field.Type.(cadence.OptionalType); ok {
			typeId = "?"
		} else {
			typeId = field.Type.ID()
		}
	}
```

internal/events/events.go
```diff
@@ -126,12 +127,23 @@ func eventString(writer io.Writer, event flow.Event) {
...
+	v := value.String()
 	var typeId string
+
+	defer func() {
+		if err := recover(); err != nil {
+			v = fmt.Sprintf("%s\n\t\thex: %x", v, v)
+			_, _ = fmt.Fprintf(writer, "\t\t- %s (%s): %s \n", field.Identifier, typeId, v)
+			log.Println("panic occurred:", err)
+		}
+	}()
```
sideninja, line 139: We don't need this recover if we handle optional type.
```suggestion
```

sukantoraymond: addressed. Thanks!

internal/events/events.go
sukantoraymond: addressed.

internal/events/events.go
```diff
@@ -128,7 +128,13 @@ func eventString(writer io.Writer, event flow.Event) {
 func printField(writer io.Writer, field cadence.Field, value cadence.Value) {
 	var typeId string
 	if field.Type != nil {
-		typeId = field.Type.ID()
```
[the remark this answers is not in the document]
sideninja: also test it with another optional type that previously worked, like optional string, although I'm not sure if optional string worked before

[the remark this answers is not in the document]
sukantoraymond: Yes, this is the result when tested with my example

````
Raymonds-MBP-3:flow raymondsukanto$ ./flow transactions send ./tx.cdc
Transaction ID: 5930133c65ee572bf4fd18ea0b2c0024680566d8a5a969c7db151810fa909af5

Status          ✅ SEALED
ID              5930133c65ee572bf4fd18ea0b2c0024680566d8a5a969c7db151810fa909af5
Payer           f8d6e0586b0a20c7
Authorizers     [f8d6e0586b0a20c7]

Proposal Key:   
    Address     f8d6e0586b0a20c7
    Index       0
    Sequence    11

No Payload Signatures

Envelope Signature 0: f8d6e0586b0a20c7
Signatures (minimized, use --include signatures)

Events:          
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooEvent
    Tx ID       5930133c65ee572bf4fd18ea0b2c0024680566d8a5a969c7db151810fa909af5
    Values
                - x (?): ["Event"] 

````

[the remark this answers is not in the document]
Comment (sukantoraymond) [reply internal/events/events.go:131]: Before change for string optional types

````
Raymonds-MBP-3:flow raymondsukanto$ ./flow events get A.f8d6e0586b0a20c7.HelloWorld.FooFooEvent

Events Block #6:
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooFooEvent
    Tx ID       7edc540ca7580d88a86ff10d6c47ecba58644105ab2ce3f9f538d056751315df
    Values
                - x (String?): "Hello, World!" 
````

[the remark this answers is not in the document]
Comment (sukantoraymond) [reply internal/events/events.go:131]: after change 

````
Envelope Signature 0: f8d6e0586b0a20c7
Signatures (minimized, use --include signatures)

Events:          
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooFooFooFooEvent
    Tx ID       c2af9dbd59ffd3390c32fb871ea52b4e045059aeb1e028a6d719d8352a4f3970
    Values
                - x (?): "Hello, World!" 

````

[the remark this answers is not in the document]
Comment (sukantoraymond) [reply internal/events/events.go:131]: Ok seems that we need to fix for non array types

File: internal/events/events.go [hunk 4]
```diff
@@ -126,14 +126,25 @@ func eventString(writer io.Writer, event flow.Event) {
...
 func printField(writer io.Writer, field cadence.Field, value cadence.Value) {
+	v := value.String()
 	var typeId string
+
+	defer func() {
+		if err := recover(); err != nil {
+			v = fmt.Sprintf("%s\n\t\thex: %x", v, v)
+			_, _ = fmt.Fprintf(writer, "\t\t- %s (%s): %s \n", field.Identifier, "?", v)
+		}
+	}()
```
sukantoraymond, line 137: I think using recover is more appropriate here because checking if the type is optional doesn't distinguish between [String]? and String?, and we can't check if it is array of string or just string too. In addition, this handles other cases like dictionary

sideninja: If you can't further define the type then recover is fine, but I wouldn't output anything in the recover since it doesn't provide much information to the user.

internal/events/events.go
```diff
@@ -126,14 +126,25 @@ func eventString(writer io.Writer, event flow.Event) {
...
+
 	if field.Type != nil {
+		//TODO: onflow/cadence issue #1672
+		//currently getting ID for cadence array will cause panic
 		typeId = field.Type.ID()
 	}
 
-	v := value.String()
 	if typeId == "" { // exception for not known typeId workaround for cadence arrays
 		v = fmt.Sprintf("%s\n\t\thex: %x", v, v)
```
sideninja, line 146: why do we output value as hex here?

internal/events/events.go
```diff
@@ -126,14 +126,25 @@ func eventString(writer io.Writer, event flow.Event) {
 }
 
 func printField(writer io.Writer, field cadence.Field, value cadence.Value) {
+	v := value.String()
 	var typeId string
+
+	defer func() {
+		if err := recover(); err != nil {
+			v = fmt.Sprintf("%s\n\t\thex: %x", v, v)
+			_, _ = fmt.Fprintf(writer, "\t\t- %s (%s): %s \n", field.Identifier, "?", v)
```
sideninja, line 135: extract this print into a method so it's not duplicated

internal/events/events.go
sukantoraymond: we don't want to print the values for parsed array? 
Currently with output in recover we get
````
Events:          
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooEvent
    Tx ID       5930133c65ee572bf4fd18ea0b2c0024680566d8a5a969c7db151810fa909af5
    Values
                - x (?): ["Event"] 

````
Without output we will get 
````
Events:          
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooEvent
    Tx ID       5930133c65ee572bf4fd18ea0b2c0024680566d8a5a969c7db151810fa909af5
    Values

````

File: internal/events/events.go (hunk 5 above)
Comment (sukantoraymond) [reply internal/events/events.go:146]: The hex was already there line 146 and below, which is triggered for array types like [String] (and not [String]?) Should we remove the hex entirely for both?

File: internal/events/events.go (hunk 4 above)
Comment (sideninja) [reply internal/events/events.go:137]: Oh nevermind, I thought the deffer still includes the error output as well. We want to get the values ofc, but it would be best if this output is uniform with the normal flow output and possibly extracted into a method, I would however drop the hex value output

File: internal/events/events.go (hunk 5 above)
Comment (sideninja) [reply internal/events/events.go:146]: I see. You are right. Lets just keep it as it was before but only for not known type ids

File: internal/events/events.go [hunk 7]
```diff
@@ -124,17 +124,28 @@ func eventString(writer io.Writer, event flow.Event) {
...
 	}
 
-	v := value.String()
 	if typeId == "" { // exception for not known typeId workaround for cadence arrays
 		v = fmt.Sprintf("%s\n\t\thex: %x", v, v)
+		typeId = "?"
 	}
-
-	_, _ = fmt.Fprintf(writer, "\t\t- %s (%s): %s \n", field.Identifier, typeId, v)
+	printValues(writer, field.Identifier, typeId, v)
```
sukantoraymond, line 150: Tested, for argument of optional type like [String]? output is 

````
Events:          
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooEvent
    Tx ID       eee4f24c6958a8978bacf2c0585608d5678b1d867e8dec3961e0a726ea714b5c
    Values
                - x (?): ["Event"] 

````

for non optional type like [String] output is 

````
Events:          
    Index       0
    Type        A.f8d6e0586b0a20c7.HelloWorld.FooEvent
    Tx ID       794e1ac2a5c5add70075e4b162bb84f15f61d7b5ac1ff560187d4d96d5e4b4c3
    Values
                - x (?): ["Event"]
                hex: 5b224576656e74225d 
````

internal/events/events.go
sukantoraymond: addressed

sideninja (approved):
Thank you, good job.

PR #518, opened 2022-06-09
Bugfix/ use emulator service account defined in flow.json for signing default

### LINKED ISSUE #346: Default value for signer flag
### Problem
The signer flag has a default value `emulator-account` which is not intuitive for cases where you change the service account in the config. Intuitively, if you change the config's service account, that account should be used as a default value for the signer. 

Imagine setting the service account to a different account and then sending a transaction without specifying the signer, you would assume it was sent using the service account set in config.

### Steps to Reproduce
1. Set service account in `flow.json` to newly created account
2. Send transaction without specifying service account

### Acceptance Criteria
The default signer should be the same as the service account.

### Proposed Solution
There are two possible solutions.
The first is where the signer flag doesn't have a default, and in case that is not specified the signer should resolve to the default service account. 
The second is where we have a special default value `service-account` and in case that is set it resolves to the account set under service account.

I believe the first solution is better as it avoids an edge case where the account you create is named `service-account`.

sukantoraymond:
Closes #346

## Description

This PR uses the emulator service account defined in emulator part of flow.json to sign transactions when there is no signer defined in the signer flag for sending transaction command
______

For contributor use:

### DIFF
--- a/internal/transactions/send.go
+++ b/internal/transactions/send.go
@@ -21,6 +21,8 @@ package transactions
 import (
 	"fmt"
 
+	"github.com/onflow/flow-cli/pkg/flowkit/config"
+
 	"github.com/spf13/cobra"
 
 	"github.com/onflow/cadence"
@@ -64,8 +66,7 @@ func send(
 	// if user has not specified signer in flag, we will use service account of default emulator as specified
 	// in flow.json
 	transactionSigner := sendFlags.Signer
-	if sendFlags.Signer == "emulator-account" {
-		fmt.Printf("signing info has default ")
+	if sendFlags.Signer == config.DefaultEmulatorServiceAccountName {
 		transactionSigner = state.Config().Emulators.Default().ServiceAccount
 	}
 	signer, err := state.Accounts().ByName(transactionSigner)
--- a/internal/transactions/send.go
+++ b/internal/transactions/send.go
@@ -60,8 +60,15 @@ func send(
 	state *flowkit.State,
 ) (command.Result, error) {
 	codeFilename := args[0]
-
-	signer, err := state.Accounts().ByName(sendFlags.Signer)
+	// if sendFlags.Signer != emulator-account, this means that user has specified a signer in command flag
+	// if user has not specified signer in flag, we will use service account of default emulator as specified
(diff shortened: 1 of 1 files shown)


internal/transactions/send.go
```diff
@@ -60,8 +60,15 @@ func send(
 	state *flowkit.State,
 ) (command.Result, error) {
 	codeFilename := args[0]
-
-	signer, err := state.Accounts().ByName(sendFlags.Signer)
+	// if sendFlags.Signer != emulator-account, this means that user has specified a signer in command flag
+	// if user has not specified signer in flag, we will use service account of default emulator as specified
+	// in flow.json
+	transactionSigner := sendFlags.Signer
+	if sendFlags.Signer == "emulator-account" {
```
sukantoraymond, line 67: Checks if signer flag value is "emulator-account" (if it is that means that user did not specify signer in flag, and we will use the emulator service account defined in emulator section of flow.json

sukantoraymond:
Screen Shot 2022-06-09 at 7 40 29 PM

Result before code change, where sending transaction still uses the address of emulator-account of f8d...c7 even though we defined it to use tempaccount2 in serviceAccount in flow.json

Screen Shot 2022-06-09 at 7 27 23 PM

Result after code change, where sending transaction  uses the address of emulator-account of 120...ab, which is the address of tempaccount2 in serviceAccount in flow.json

internal/transactions/send.go
sideninja, line 67: This should reference a variable instead of literal string

internal/transactions/send.go
```diff
@@ -60,8 +60,15 @@ func send(
...
+		fmt.Printf("signing info has default ")
```
sideninja, line 68: Why was this output added?

internal/transactions/send.go
sukantoraymond: addressed

internal/transactions/send.go
sukantoraymond: addressed

sideninja (approved):
Some small feedback, looks good. Thank you for providing the testing results.

PR #523, opened 2022-06-14
Feature/deploy update smarter

### LINKED ISSUE #498: make `flow project deploy` smarter
## Instructions

Right now flow project deploy either allow you to deploy all contracts or update all contracts. If you have an account with several contracts and change just one having to wait until everything deploys takes time. 

I propose to make the command smarter in that it will look at chain to see what is there and do the right thing. We can even add a --diff flag to allow it to show diffs that will be deployed incase of update. 

I have some code in overflow for doing this, granted this is used for mainnet deploy but the pattern is the same. 
https://github.com/findonflow/find/blob/main/tasks/deploy/main.go

### Issue To Be Solved

Make it faster and funnier to deploy contracts onflow!

sukantoraymond:
Closes #498 

## Description

We skip update for a contract if there is no diff between the contract to be deployed and the contract already deployed in account. This PR also updates the deploy CLI output to show during deployment which change is add and which change is an update.
______

For contributor use:


sukantoraymond:
This PR has been manually tested in the following scenarios: 

When first deploying project
````
Raymonds-MBP-3:flow raymondsukanto$ ./flow project deploy 

Deploying 1 contracts for accounts: emulator-account

HelloWorld -> 0xf8d6e0586b0a20c7 (45200faa85cd5cbe59ead076e999ebcc68cfab015c039d6410daf0416298bb9f) 


✨ All contracts deployed successfully

````

When updating an existing contract not containing diff
````
Raymonds-MBP-3:flow raymondsukanto$ ./flow project deploy --update

Deploying 1 contracts for accounts: emulator-account

no diff found in HelloWorld, skipping update

✨ All contracts deployed successfully

````


When updating an existing contract containing diff
````
Raymonds-MBP-3:flow raymondsukanto$ ./flow project deploy --update

Deploying 1 contracts for accounts: emulator-account

HelloWorld -> 0xf8d6e0586b0a20c7 (505bb819f424cb4db0003050487345e61813377f89f9a7468fecd318399a0398) (update)

1 contracts updated successfully

✨ All contracts deployed successfully

````

When updating an existing contract containing diff and adding a new contract
````
Raymonds-MBP-3:flow raymondsukanto$ ./flow project deploy --update

Deploying 2 contracts for accounts: emulator-account

HelloWorld -> 0xf8d6e0586b0a20c7 (fc72a313d5942b3860d2204335ee4c3408e1f33cff86533927ec473f359f6809) (update)

HelloWorldNew -> 0xf8d6e0586b0a20c7 (8f4ba2774d19aba97bc8318a5ecaf1ad2161918ae018b4217027fcb4e186d10e) 

1 contracts updated successfully

✨ All contracts deployed successfully
````

File: go.sum [hunk 1]
```diff
@@ -656,6 +656,8 @@ github.com/hashicorp/memberlist v0.3.0/go.mod h1:MS2lj3INKhZjWNqd3N0m3J+Jxf3DAOn
 github.com/hashicorp/serf v0.8.2/go.mod h1:6hOLApaqBFA1NXqRQAsxw9QxuDEvNxSQRwA/JwenrHc=
 github.com/hashicorp/serf v0.9.5/go.mod h1:UWDWwZeL5cuWDJdl0C6wrvrUwEqtQ4ZKBKKENpqIUyk=
 github.com/hashicorp/serf v0.9.6/go.mod h1:TXZNMjZQijwlDvp+r0b63xZ45H7JmCmgg4gpTwn9UV4=
+github.com/hexops/gotextdiff v1.0.3 h1:gitA9+qJrrTCsiCl7+kh75nPqQt1cx4ZkudSTLoUqJM=
```
sukantoraymond, line 659: library for showing diffs

pkg/flowkit/services/project.go
```diff
@@ -200,9 +208,29 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
...
-			if err != nil {
-				return nil, err
+			//only update contract if there is diff
+			if noDiffInContract {
+				p.logger.Info(fmt.Sprintf(
+					"no diff found in %s, skipping update",
+					contract.Name(),
+				))
+				continue
+			} else {
```
sideninja, line 218: You don't need else block here.

pkg/flowkit/services/project.go
```diff
@@ -200,9 +208,29 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
...
+				if !force {
+					edits := myers.ComputeEdits(span.URIFromPath(contract.Source()), string(existingContract), contract.Code())
+					diff := fmt.Sprint(gotextdiff.ToUnified("deployed", "changes", string(existingContract), edits))
+
+					fmt.Println("DIFF of " + contract.Name())
+					fmt.Println(diff)
+					fmt.Println("------------------------")
+					fmt.Println(" press <ENTER> to update")
+					fmt.Scanln()
```
sideninja, line 227: You shouldn't use `fmt.Println` anywhere in the service layer, if you need to log something you should use a logger, if you need an interactive prompt you should use the prompt in the `output` package https://github.com/onflow/flow-cli/blob/master/pkg/flowkit/output/prompt.go.

pkg/flowkit/services/project_test.go
```diff
@@ -102,7 +102,7 @@ func TestProject(t *testing.T) {
 			gw.SendSignedTransaction.Return(tests.NewTransaction(), nil)
```
sideninja, line 102: A test that includes this new feature should be added.

sideninja:
Docs should be updated. Do a mental check when you "check" the "update relevant documentation". A simple way to think about it is if there are any new flags, changes in commands, changes in arguments, that's for sure a request for doc update.

sideninja:
Good feature! but I would say the default action should be: go through contracts, if no changes are found skip and output (this is already done ✅) if a change is found then update it (this is different to how its implemented). Reasons why I believe this should be the default, first it doesn't break backward compatibility, secondly if a user runs an update command then they should expect an update by default, avoid interactive prompts whenever possible https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#interaction doing so allows chaining commands which is really important when building CLIs. Majority of usecases are like said, a user wants to update all the contracts so they shouldn't be expected to do additional work, in minority of cases when they want to approve each change we can provide this functionality, but I would argue at this time we could even skip this altogether, however if you think it's a good feature then I would advise going the other way around, so requiring a flag when user wants to approve the changes and otherwise by default updating without interactions. Also for future reference be aware of the global `-y` flag, this flag is meant to be as an approval of any commands that shows prompts.

pkg/flowkit/services/project.go
sukantoraymond: addressed

sukantoraymond:
Ya I don't think the prompting is necessary at this point. Will just do auto update

pkg/flowkit/services/project.go
sukantoraymond: don't need this anymore since we are not doing prompts

sideninja (approved):
Great work. LGTM 🚀

PR #528, opened 2022-06-17
add view to transactions to allow user to see all parts of a transact…

bthaile:
…ion RLP

Closes #???

## Description

Add `view` command to transactions to allow decoding transaction RLP to see code and signatures.
______

For contributor use:

### DIFF
--- a/internal/transactions/view.go
+++ b/internal/transactions/view.go
@@ -1,0 +1,71 @@
+/*
+ * Flow CLI
+ *
+ * Copyright 2019 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package transactions
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-cli/pkg/flowkit"
+
+	"github.com/spf13/cobra"
+
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flowkit/services"
+)
+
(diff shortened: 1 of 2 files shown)


internal/transactions/view.go
```diff
@@ -0,0 +1,71 @@
...
+
+type flagsView struct {
+	Include []string `default:"" flag:"include" info:"Fields to include in the output. Valid values: signatures, code, payload."`
+}
+
+var viewFlags = flagsView{}
+
+var ViewCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "view <transaction filename>",
```
janezpodhostnik, line 40: Interesting choice of verb.

I would say that this is more of a decode than a view...

the description should maybe also include that this is for decoding specifically rlp files

janezpodhostnik (commented):
The code looks good, but I have some comments about the naming and the description.

Would it also be possible to write a test for this somehow?

internal/transactions/view.go
Kay-Zee: Yeah, decode works as well, though originally my main thought for this function was to see the transaction in the file. So i feel like this more encapsulates what the user of this function really wants to do (see the tx), where as decode is the action that's performed to make that possible.

Any thoughts on this @sideninja

bthaile: `decode` does make sense since RLP is getting decoded. Also `view` makes sense viewing the transaction in human readable terms. The other transaction verbs are simple and in non-techie terms terms. I used view cuz Kan mentioned the term, so it was stuck in my head. I'm open to using a different verb.

sideninja: Apologies for the late response I was sick. I think this is a great addition to the transaction commands. I would say that using decode would be more consistent with the already existing `flow keys decode` command, but at the same time I would say view/decode are both OK and the only thing that is not subjective would be consistency as mentioned in guidelines https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#commands

sideninja (changes requested):
Great work. The only thing that is missing is documentation. I would kindly ask you to add it to the `/docs` folder, you can take a look at the https://github.com/onflow/flow-cli/blob/master/docs/get-transactions.md for an example.

internal/transactions/view.go
bthaile: Thanks all, I'll update to use `decode`

Kay-Zee (approved):
LGTM

sideninja (approved):
LGTM, thank you

PR #538, opened 2022-06-27
Interactive account creation for mainnet, testnet and emulator

### LINKED ISSUE #537: Interactive creation of accounts
We need to support creation of accounts on testnet and mainnet to be more seamless integrated into the Flow CLI and not required multiple steps.
We should have this notion of interactive commands in case where the command doesn't get passed the correct arguments or flags, which for this case would be just running the 
```
flow accounts create
```

This would open up a prompt first making you choose which network you want to create the account for, after it would create an account on:
- emulator: would just generate the keys, create an account by sending tx with service signer, save to flow.json
- testnet: would generate the key, open up a faucet with already profiled values and make you fill in the captcha, after that it would monitor the network for new account creation event and get the newly created address, it would then save that account in flow.json
- mainnet: it would be similar to testnet flow, but would use flow port for creation of account, and after we get back the address it would encrypt the key part of the account in flow.json

We should look into making faucet expose an API for above action, so the user doesn't have to click on the website.

sideninja:
Closes #537 

## Description
This PR implements the account creation interactive flow for mainnet, testnet and emulator networks. It also covers the encryption of keys for mainnet case. It also provides password prompts when running other commands with the encrypted keys.

The flow is as follows, you choose the network, if the chosen network is testnet then the faucet website is opened, if it's mainnet the flow port is opened, after you complete the required steps in browser to create an account, the CLI listens for new account creation event and stores the account in configuration. In the case of mainnet it offers you an option to encrypt.

Some screens:
Screenshot 2022-06-30 at 16 26 30
Screenshot 2022-06-30 at 16 26 47
Screenshot 2022-06-30 at 16 27 49
Screenshot 2022-06-30 at 16 26 11

DOD:
- [x] Implemneted
- [x] Tested
- [ ] Released

______

For contributor use:


pkg/flowkit/account.go
```diff
@@ -111,6 +111,45 @@ func toConfig(account Account) config.Account {
 	}
 }
 
+func AccountFromFlow(account *flow.Account, name string, key crypto.PrivateKey) (*Account, error) {
+	flowkitAccount := &Account{
+		name:    name,
+		address: account.Address,
+	}
+
+	if len(account.Keys) > 0 {
```
psiemens, line 120: I think that both the `AccountFromFlow` and `AccountFromFlowEncrypted` functions should be more strict and return an error if `len(account.Keys) != 1`. 

This prevents an exploit where an attacker front-runs account creation transactions and submits their own account creation transaction with the victim's public key, as well as their own additional malicious key. A user could then be tricked into sending funds to an account address that they only partially own.

psiemens: In addition to that, we should probably also throw an error if `defaultKey.Weight != 1000`. This is an unlikely edge case, but an account is unusable if it has only one key and that key has less than 1000 weight.

pkg/flowkit/util/utilities.go
```diff
@@ -116,3 +119,43 @@ func ValidateECDSAP256Pub(key string) error {
...
+
+	return link
+}
+
+func MainnetFlowPortURL(publicKey string) string {
+	const flowPortURL = "https://port.onflow.org/transaction?hash=a0a78aa7821144efd5ebb974bb52ba04609ce76c3863af9d45348db93937cf98&showcode=false&consent=true&pk="
+	return fmt.Sprintf("%s%s", flowPortURL, strings.TrimPrefix(publicKey, "0x"))
+}
+
+func RandomName() string {
```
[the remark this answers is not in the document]
sukantoraymond: addressed

pkg/flowkit/events.go
```diff
@@ -60,26 +73,49 @@ func newEvent(event flow.Event) Event {
...
+					}
+				}
+			}
+
+			// older format support
+			if p, ok := event.Values["publicKey"].(cadence.Array); ok {
+				var parsedKey []byte
+				_ = json.Unmarshal([]byte(p.String()), &parsedKey)
+
+				if publicKey.String() == fmt.Sprintf("0x%x", parsedKey[4:len(parsedKey)-5]) {
```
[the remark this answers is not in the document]
sukantoraymond: Yes the old format is being used by Testnet faucet in account creation. Placed print statements in old format section and they are printed for testnet. testing if they are parsed correctly.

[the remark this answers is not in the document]
sukantoraymond:
````
if strings.Contains(parsedKeyHex, publicKeyHex) {
````
doesn't work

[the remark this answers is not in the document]
Comment (sukantoraymond) [reply pkg/flowkit/events.go:113]: changing it to this instead 
````
				parsedKeyhex := fmt.Sprintf("%x", parsedKey)
				publicKeyHex := publicKey.String()
				//we have to remove 0x from beginning of publicKeyHex
				if strings.Contains(parsedKeyhex, publicKeyHex[2:]) {
````

File: pkg/flowkit/events.go [hunk 4]
```diff
@@ -60,26 +73,49 @@ func newEvent(event flow.Event) Event {
...
+					if bytes.Equal(parsedKey, publicKey.Encode()) {
+						return e.GetAddress()
+					}
+				}
+			}
+
+			// older format support
+			if p, ok := event.Values["publicKey"].(cadence.Array); ok {
+				var parsedKey []byte
+				_ = json.Unmarshal([]byte(p.String()), &parsedKey)
```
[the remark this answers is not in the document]
sukantoraymond: addressed

internal/accounts/create.go
```diff
@@ -134,3 +162,158 @@ func create(
...
+			os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			fmt.Printf("Error with .env file: " + err.Error())
+		}
+		defer f.Close()
+		newPKeyEnvVar := privateKeyEnvKey + "=" + strings.TrimPrefix(key.String(), "0x")
+		if _, err := f.WriteString("\n" + newPKeyEnvVar); err != nil {
+			fmt.Printf("Error with writing to .env file:" + err.Error())
+		}
+		log.Info(output.SuccessEmoji() + "Private key env variable stored in cmd/flow/private_keys.env")
```
sukantoraymond, line 267: Edited Gregor's implementation of using encryption for mainnet as discussed, and as intermediate solution to use environment variable to store private key, with the option to save in .env file if desired. 

Implementation here will save in the .env file in cmd/flow/private_key.env and will create the env file automatically if there is none detected in cmd/flow. If the file is detected, it will just add $PRIVATE_KEY_{address} = {privatekey} to the next line

internal/accounts/create.go
```diff
@@ -134,3 +162,158 @@ func create(
...
+		path, _ := os.Getwd()
+		envFilePath := path + "/private_keys.env"
+
+		f, err := os.OpenFile(envFilePath,
+			os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			fmt.Printf("Error with .env file: " + err.Error())
+		}
+		defer f.Close()
+		newPKeyEnvVar := privateKeyEnvKey + "=" + strings.TrimPrefix(key.String(), "0x")
```
sukantoraymond, line 263: I choose to use privatekey_{address} for env variable key instead of privatekey_{name}  due to possible name duplicates.

pkg/flowkit/output/prompt.go
```diff
@@ -507,3 +507,101 @@ func ReportCrash() bool {
...
+	encryptPrompt := promptui.Select{
+		Label: "Do you want to save the private key into a .env file?",
+		Items: []string{"Yes(IMPORTANT: Don't commit created env file)", "No"},
+	}
+	index, _, err := encryptPrompt.Run()
+	if err == promptui.ErrInterrupt {
+		os.Exit(-1)
+	}
+
+	return index == 0
```
sukantoraymond, line 539: Prompt for user whether they want to save private key env variable to .env file

pkg/flowkit/config/processor.go
```diff
@@ -43,8 +43,14 @@ func ProcessorRun(raw []byte) ([]byte, map[string]string) {
...
 func processEnv(raw string) string {
-	_ = godotenv.Load() // try to load .env file
-
+	path, _ := os.Getwd()
+	envFilePath := path + "/private_keys.env"
+	_, error := os.Stat(envFilePath)
+	// check if error is not "file not exists"
+	if error == nil {
+		godotenv.Load(envFilePath)
+	}
```
sukantoraymond, line 52: Will always load private_keys.env and read the env variables whenever we start up the command

pkg/flowkit/config/processor.go
```diff
@@ -43,8 +43,14 @@ func ProcessorRun(raw []byte) ([]byte, map[string]string) {
...
+	_ = godotenv.Load()
```
sukantoraymond, line 53: I have left this line behind because of https://docs.onflow.org/flow-cli/security/#store-configuration-in-environment-variables, but I don't think this line actually loads any env, because it reads only from the existing directory, which in this case will be config/processor.go. If we really want to read the env files, maybe it should be from cmd/flow or just in the flowcli directory itself.

pkg/flowkit/config/json/account.go
```diff
@@ -47,9 +49,21 @@ func transformAddress(address string) (flow.Address, error) {
...
+	if strings.Contains(a.Key, pKeyEnvKey) {
+		privateKey, ok := os.LookupEnv(pKeyEnvKey)
+		if !ok {
+			return nil, fmt.Errorf("could not find value of environment variable %s", a.Key)
+		}
+		configAccountKey = privateKey
+	}
+	if configAccountKey == "" {
+		configAccountKey = a.Key
+	}
```
sukantoraymond, line 63: implementation to look up the private key from the env variable

pkg/flowkit/config/json/account.go
```diff
@@ -173,27 +206,49 @@ func transformSimpleAccountToJSON(a config.Account) account {
...
+			Key:     getEnvKeyVariable(a.Address.String(), a.Key.PrivateKey.String()),
 		},
 	}
 }
+func getEnvKeyVariable(address string, privateKey string) string {
+	privateKeyEnvKey := "PRIVATE_KEY_" + strings.ToUpper(strings.TrimPrefix(address, "0x"))
+	_, ok := os.LookupEnv(privateKeyEnvKey)
+	if !ok {
+		return strings.TrimPrefix(privateKey, "0x")
+	}
```
sukantoraymond, line 218: If we can't find the private key from env variable, we will not store it in json as $PRIVATE_KEY_{address}, instead just the private key as how it has been done currenlty.

sukantoraymond:
For env variables, we will have the option whether or not to store the private key in .env file (as shown below). 
Screen Shot 2022-07-01 at 10 28 59 PM

If user doesn't want to store in .env file, user will need to run ` export PRIVATE_KEY_{address}="{private key}"` so that env variable is loaded up.

If user wants to use .env file, the .env file will be created (if it doesn't exist) and the env variable will be added in the .env file, as shown below. (Private key env variable will be shown in log output regardless of whether or not user chooses to save in .env file)

Screen Shot 2022-07-01 at 10 29 25 PM


The resulting flow.json is shown below
````
	

"accounts": {
		"1": {
			"address": "8e7bcf2294bd120e",
			"key": "8ac271230f78ad9d375c9203c250bc67ef07fd8d49e80ea5013c42f5889f7a14"
		},
		"Tom": {
			"address": "02e59aec9ff63f06",
			"key": "$PRIVATE_KEY_02E59AEC9FF63F06"
		}
	},
````

**Review by janezpodhostnik — COMMENTED** [review]
Great idea and PR.

Left some comments.

File: internal/accounts/create.go [hunk 12]
```diff
@@ -134,3 +162,158 @@ func create(
...
+	} else {
+		var link string
+		switch network {
+		case config.DefaultTestnetNetwork():
+			log.Info("\nA testnet faucet website will open, follow the steps to create an account: \n 1. Fill in the captcha, \n 2. Click on 'Create Account' button.\n")
+			link = util.TestnetFaucetURL(key.PublicKey().String(), crypto.ECDSA_P256)
+		case config.DefaultMainnetNetwork():
+			log.Info("\nA Flow Port website will open, follow the steps to create an account: \n 1. Click on 'Submit' button, \n 2. Connect existing Blocto or create a new account first, \n 3. Click on confirm, \n 4. Click on approve. \n")
+			link = util.MainnetFlowPortURL(key.PublicKey().String())
+		}
```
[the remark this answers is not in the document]
psiemens: Great feedback. I'll bring this up with Srinjoy today, but I'm inclined to say that we can keep the functionality simple for the first release of this feature and then cover these cases later.

pkg/flowkit/keys.go
```diff
@@ -248,3 +259,162 @@ func (a *HexAccountKey) Validate() error {
...
+func (a *EncryptedAccountKey) Validate() error {
+	_, err := a.decrypt()
+	return err
+}
+
+func (a *EncryptedAccountKey) SetPassword(password string) {
+	a.password = password
+}
+
+func encrypt(key, data []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
psiemens: Removed the encryption feature for now, so no need for Tarak to review.

internal/accounts/create.go
```diff
@@ -134,3 +148,258 @@ func create(
...
+		return nil, err
+	}
+	service := services.NewServices(gw, state, output.NewStdoutLogger(output.NoneLog))
+	log.Info("\nSTARTING ACCOUNT CREATION PROCESS: ")
+	log.Info("\n------------------------")
+	log.Info(fmt.Sprintf("%s (1/3) Generate public and private keys", output.GoEmoji()))
+	log.Info(fmt.Sprintf("%s (2/3) Create account on %s with generated keys", output.GoEmoji(), network.Name))
+	log.Info(fmt.Sprintf("%s (3/3) Save newly created account on flow.json", output.GoEmoji()))
+	log.Info("\n------------------------")
+	time.Sleep(time.Second * 2)
```
sukantoraymond, line 168: Improved interactivity by detailing each step that is happening and adding time lag in between steps to allow users to have "the feel of interactivity"

pkg/flowkit/account.go
```diff
@@ -101,6 +101,21 @@ func NewAccountFromOnChainAccount(
...
+		defaultKey := account.Keys[0] // todo check if ok to default
+		flowkitAccount.SetKey(
+			NewHexAccountKeyFromPrivateKey(defaultKey.Index, defaultKey.HashAlgo, key),
+		)
+	}
+
+	return flowkitAccount, nil
+}
```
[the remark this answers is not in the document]
sukantoraymond: addressed

pkg/flowkit/tests/resources.go
```diff
@@ -278,14 +278,16 @@ func Donald() *flowkit.Account {
 	return newAccount("Donald", "0x3", "seedseedseedseedseedseedseedseedseedseedseedseedDonald")
 }
 func newAccount(name string, address string, seed string) *flowkit.Account {
-	a := &flowkit.Account{}
```
sukantoraymond, line 281: change is from merge conflict resolution

internal/accounts/create.go
```diff
@@ -134,3 +149,277 @@ func create(
...
+	file, err := os.OpenFile(gitIgnoreDir, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	_, err = file.WriteString("\n" + filename)
+	if err != nil {
+		return err
+	}
```
sukantoraymond, line 422: auto add to gitignore

psiemens: It'd be good to move this into its own function to avoid making this function too long.

internal/accounts/create.go
```diff
@@ -134,3 +149,277 @@ func create(
...
+
+	currentWd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	flowCliDir := strings.TrimSuffix(currentWd, "cmd/flow")
```
psiemens, line 409: Hmm, you shouldn't have to trim the directory like this. How are you running the CLI while developing?

This might be needed because you're running from the `cmd/flow` directory. Try running `go run ./cmd/flow/main.go` instead.

internal/accounts/create.go
```diff
@@ -134,3 +149,277 @@ func create(
...
+	gitIgnoreDir := flowCliDir + ".gitignore"
```
psiemens, line 410: Best to avoid direct string manipulation on file paths. You can use the `path` package for this in Go.

```go
gitIgnorePath := path.Join(currentWd, ".gitignore")
```

internal/accounts/create.go
psiemens: I think we should ask the user before creating/updating the `.gitignore` file. Could you add a yes/no prompt that gives the user the option of updating `.gitignore`?

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
```diff
@@ -134,3 +149,277 @@ func create(
...
+		address, err = getAccountCreatedAddressWithPubKey(service, pubKey, startHeight)
+		if err != nil {
+			return nil, err
+		}
+
+		return address, nil
+	}
+
+	return address, nil
+}
```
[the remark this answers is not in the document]
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
```diff
@@ -134,3 +149,277 @@ func create(
...
+	state *flowkit.State,
+	account *flowkit.Account,
+	network config.Network,
+) error {
+	// If using emulator, save account private key to main flow.json configuration file
+	if network == config.DefaultEmulatorNetwork() {
+		return saveAccountToMainConfigFile(state, account)
+	}
+
+	// Otherwise, save to a separate {accountName}.private.jsonfile.
```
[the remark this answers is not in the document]
sukantoraymond: addressed

pkg/flowkit/config/account.go
```diff
@@ -30,6 +30,18 @@ type Account struct {
...
+	//
+	// This field is only set if the external "fromFile"
+	// syntax is used. Otherwise this field is empty.
+	//
+	// Ref: https://docs.onflow.org/flow-cli/security/#private-account-configuration-file
+	FromFile string
+	// saveAdvancedFormat is to save an account in advanced format even if it has default values as defined
+	// in isDefaultKeyFormat() in flowkit/config/json/account.go
+	// defaults to false
+	SaveAdvancedFormat bool
```
sukantoraymond, line 44: added this field so that account is saved as advanced format even if it has default account values (otherwise it will be saved as simple format), as per discussion to save account as advanced format in private json files.

pkg/flowkit/config/json/account_test.go
```diff
@@ -362,7 +362,22 @@ func Test_TransformAccountToJSON(t *testing.T) {
...
+	accounts, err := jsonAccounts.transformToConfig()
+	assert.NoError(t, err)
 
+	accounts[0].SaveAdvancedFormat = true
+	j := transformAccountsToJSON(accounts)
+	x, _ := json.Marshal(j)
+
+	// our output format is shorted - improve test
+	assert.Equal(t, string(b), string(x))
+}
```
psiemens, line 380: Thanks for adding a test case!

pkg/flowkit/account.go
```diff
@@ -41,6 +41,10 @@ type Account struct {
 	//
 	// Ref: https://docs.onflow.org/flow-cli/security/#private-account-configuration-file
 	fromFile string
+	// saveAdvancedFormat is to save an account in advanced format even if it has default values as defined
+	// in isDefaultKeyFormat() in flowkit/config/json/account.go
+	// defaults to false
+	saveAdvancedFormat bool
```
psiemens, line 47:
```suggestion
	// useAdvancedSaveFormat can be used to indicate that this account should be saved in advanced format
	//  even if it uses default values (as defined in isDefaultKeyFormat() in flowkit/config/json/account.go).
	// 
	// This field defaults to false.
	useAdvancedSaveFormat bool
```

`useAdvancedSaveFormat` reads a bit more clearly to me. Great way to implement the advanced saving!

pkg/flowkit/account.go
```diff
@@ -141,6 +145,10 @@ func (a *Account) SetFromFile(filename string) *Account {
 	return a
 }
 
+// SetAccountSaveAdvanced sets saveAdvancedFormat as true.
+func (a *Account) SetAccountSaveAdvanced() {
```
psiemens, line 149:
```suggestion
func (a *Account) EnableAdvancedSaveFormat() {
```

pkg/flowkit/config/json/account.go
```diff
@@ -166,7 +166,7 @@ func transformAccountsToJSON(accounts config.Accounts) jsonAccounts {
 	for _, a := range accounts {
 		if a.FromFile != "" {
 			jsonAccounts[a.Name] = transformFromFileAccountToJSON(a)
-		} else if isDefaultKeyFormat(a.Key) {
+		} else if isDefaultKeyFormat(a.Key) && !a.SaveAdvancedFormat {
```
psiemens, line 169: I would also update the `transformAdvancedToConfig` function to set the new `SaveAdvancedFormat` field to true. This way we prevent the advanced format JSON objects from being converted back to simple format.

psiemens (commented):
Nice changes!

internal/accounts/create.go
```diff
@@ -56,11 +62,23 @@ var CreateCommand = &command.Command{
...
+	// if user doesn't provide any flags go into interactive mode
+	if len(createFlags.Keys) == 0 {
+		account, err := createInteractive(state, loader)
+		if err != nil {
+			return nil, err
+		}
+		return &AccountResult{
+			Account: account,
+			include: createFlags.Include,
+		}, nil
```
psiemens, line 79: @sukantoraymond Here's how I think we can silence the noisy account result.

Rationale: This is an interactive command that already prints to stdout, there's no need to print out the account struct. It takes up a lot of space in the terminal. It's also unlikely that users will consume this command output from another script, so we can remove the JSON output feature.

```suggestion
		err := createInteractive(state, loader)
		if err != nil {
			return nil, err
		}
		return nil, nil
```

- `createInteractive` doesn't need to return an account value (it already prints it itself).
-  We return `nil` rather than returning an `AccountResult` value. This should tell the command framework not to print anything.

However, [this line in command.go](https://github.com/onflow/flow-cli/blob/master/internal/command/command.go#L132-L133) will throw an error if `result` is `nil`.

To avoid that, I think we can just update that function to skip logging results if `result == nil`:

```go
// internal/command/command.go line 129 to 137

handleError("Command Error", err)

if result == nil {
  return
}

// format output result
formattedResult, err := formatResult(result, Flags.Filter, Flags.Format)
handleError("Result", err)

// output result
err = outputResult(formattedResult, Flags.Save, Flags.Format, Flags.Filter)
handleError("Output Error", err)
```

sukantoraymond (dismissed):

pkg/flowkit/account.go
sukantoraymond: addressed

pkg/flowkit/account.go
sukantoraymond: addressed

pkg/flowkit/config/json/account.go
sukantoraymond: Good point. Doing that.

pkg/flowkit/config/json/account_test.go
```diff
@@ -344,7 +344,7 @@ func Test_TransformDefaultAccountToJSON(t *testing.T) {
 	x, _ := json.Marshal(j)
 
 	// our output format is shorted - improve test
-	assert.NotEqual(t, string(b), string(x))
+	assert.Equal(t, string(b), string(x))
```
sukantoraymond, line 347: Since we now set `SaveAdvancedFormat` to true after `transformAdvancedToConfig`, this will cause this test to be equal even if the json format has default values, if it appears as advanced format, i.e. `key` field being not a string

psiemens: I think this test is specifically meant to test the case where an account key with default values is transformed into the simplified JSON format, so I don't think we should change the expected value here.

I think the test can be improved, though (as the comment on line 346 suggests).

Rather than parsing the JSON, maybe just start with an `Account` object.

```go
// You can use a helper to create this account if you want to re-use it in other tests.
func defaultAccount(t *testing.T) config.Account {
  privateKey, err := crypto.DecodePrivateKeyHex(
    "1272967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47",
    crypto.ECDSA_P256,
  )
  assert.NoError(t, err)

  key := flowkit.NewHexAccountKeyFromPrivateKey(0, crypto.SHA3_256, privateKey)

  account := config.
    NewAccount("emulator-account).
    SetAddress(flow.HexToAddress("f8d6e0586b0a20c7")).
    SetKey(key)

  return account
}

// ...

account := defaultAccount(t)

accounts := []Account{account}

j := transformAccountsToJSON(accounts)
result, err := json.Marshal(j)
assert.NoError(t, err)

expected := []byte(`{"emulator-account":{"address":"f8d6e0586b0a20c7","key":"1272967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"}}`)

assert.Equal(t, result, expected)
```

pkg/flowkit/config/json/account_test.go
```diff
@@ -362,7 +362,22 @@ func Test_TransformAccountToJSON(t *testing.T) {
...
+func Test_TransformDefaultAccountToJSONAdvanced(t *testing.T) {
+	b := []byte(`{"emulator-account":{"address":"f8d6e0586b0a20c7","key":{"type":"hex","index":0,"signatureAlgorithm":"ECDSA_P256","hashAlgorithm":"SHA3_256","privateKey":"1272967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"}},"testnet-account":{"address":"3c1162386b0a245f","key":"2272967fd2bd75234ae9037dd4694c1f00baad63a10c35172bf65fbb8ad74b47"}}`)
+
+	var jsonAccounts jsonAccounts
+	err := json.Unmarshal(b, &jsonAccounts)
+	assert.NoError(t, err)
+	accounts, err := jsonAccounts.transformToConfig()
+	assert.NoError(t, err)
+
+	accounts[0].UseAdvancedSaveFormat = true
```
psiemens, line 374: I think this line is redundant now that the parser automatically identifies advanced accounts.

I would construct this test similarly to the one above:

```go
account := defaultAccount(t)
account.EnabledAdvancedSaveFormat()

// ... test that it results in advanced JSON
```

sukantoraymond: you are right. Removing it.

psiemens (commented):
@sukantoraymond Your latest changes look good!

Now that @sideninja is back, I think we should give him a chance to give a final review. I want him to make sure I didn't make any bad suggestions. 😛

pkg/flowkit/output/prompt.go
```diff
@@ -507,3 +505,57 @@ func ReportCrash() bool {
 
 	return chosen == 0
 }
+
+func CreateAccountNetworkPrompt() (string, config.Network) {
+	networkNames := make([]string, len(config.DefaultNetworks()))
+	for i, net := range config.DefaultNetworks() {
+		networkNames[i] = net.Name
+	}
```
sideninja, line 513: This is not needed anymore

pkg/flowkit/output/prompt.go
```diff
@@ -507,3 +505,57 @@ func ReportCrash() bool {
...
+
+	networkPrompt := promptui.Select{
+		Label: "Which network do you want to create the account on?",
+		Items: []string{"Local Emulator", "Flow Testnet", "Flow Mainnet"},
```
sideninja, line 517: I'ma bit nervous about hardcoding these values and assuming the indexes are right. Maybe add a test to make sure they don't get out of order.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
 		include: createFlags.Include,
 	}, nil
 }
+
+func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow.Account, error) {
+	log := output.NewStdoutLogger(output.InfoLog)
+
+	log.Info(fmt.Sprintf("%s What name would you like to give this new account?", output.GoEmoji()))
```
sideninja, line 154: This question should be part of the prompt. We should avoid emoji here, so we don't overload the users screen with emjois, they are meant to be used when we really need to get users attention to something important, like warning, or confirmation of action etc.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+		saveToGitIgnore = output.AddToGitIgnorePrompt(privateJsonFileName)
+	}
+	log.Info(fmt.Sprintf("%s This command will perform the following:\n", output.WarningEmoji()))
+	log.Info(fmt.Sprintf("- Generate a new ECDSA P-256 public and private key pair\n"))
+	if selectedNetwork != config.DefaultEmulatorNetwork() {
+		if saveToGitIgnore {
+			log.Info(fmt.Sprintf("- Save the private key to %s\n", privateJsonFileName))
+		}
+	}
+	log.Info(fmt.Sprintf("- Create a new account on %s and pair the public key with the new account\n", networkName))
```
sideninja, line 178:
```suggestion
	log.Info(fmt.Sprintf("- Create a new account on %s and pair the public key with the new account", networkName))
```

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	}
+	privateJsonFileName := ""
+	saveToGitIgnore := false
+	if selectedNetwork != config.DefaultEmulatorNetwork() {
+		privateJsonFileName = fmt.Sprintf("%s.private.json", name)
+		log.Info(fmt.Sprintf("%s For security purposes, the private key generated for this account will be "+
+			"stored in separate file: %s", output.WarningEmoji(), privateJsonFileName))
+		saveToGitIgnore = output.AddToGitIgnorePrompt(privateJsonFileName)
+	}
+	log.Info(fmt.Sprintf("%s This command will perform the following:\n", output.WarningEmoji()))
```
sideninja, line 171: I would drop the new lines, and just add it after the list, it makes the output nicer:
```
✔ Local Emulator

This command will perform the following:
- Generate a new ECDSA P-256 public and private key pair
- Create a new account on Local Emulator and pair the public key with the new account
- Save the newly created account configuration to flow.json
```

instead of:
```
✔ Local Emulator
⚠️ This command will perform the following:

- Generate a new ECDSA P-256 public and private key pair

- Create a new account on Local Emulator and pair the public key with the new account

- Save the newly created account configuration to flow.json

Use the arrow keys to navigate: ↓ ↑ → ←
```

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	log.Info(fmt.Sprintf("- Generate a new ECDSA P-256 public and private key pair\n"))
+	if selectedNetwork != config.DefaultEmulatorNetwork() {
+		if saveToGitIgnore {
+			log.Info(fmt.Sprintf("- Save the private key to %s\n", privateJsonFileName))
+		}
+	}
+	log.Info(fmt.Sprintf("- Create a new account on %s and pair the public key with the new account\n", networkName))
+	log.Info(fmt.Sprintf("- Save the newly created account configuration to flow.json\n"))
+	output.NextStepPrompt()
```
sideninja, line 180: This prompt should be changed so it's not a selection prompt:
```
Use the arrow keys to navigate: ↓ ↑ → ←
? Press <ENTER> to continue:
  ▸ CONTINUE
```

it should look like:
```
Press <ENTER> to continue
```
This way there is not a lot of screen property taken for this simple action.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	service := services.NewServices(gw, state, output.NewStdoutLogger(output.NoneLog))
+
+	log.Info(fmt.Sprintf("%s Generating public and private keys for %s", output.WarningEmoji(), name))
+
+	key, err := service.Keys.Generate("", crypto.ECDSA_P256)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info(fmt.Sprintf("%s Successfully generated public and private keys", output.OkEmoji()))
```
sideninja, line 191: I don't think we need to have two outputs for the generation of keys:
(Also the extra continue is confusing and not needed)
```
⚠️ Generating public and private keys for Foo
✅ Successfully generated public and private keys
Use the arrow keys to navigate: ↓ ↑ → ←
? Press <ENTER> to continue:
  ▸ CONTINUE
```

This should just be:
```
Successfully generated public and private keys
```

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	if err != nil {
+		return nil, err
+	}
+	log.Info(fmt.Sprintf("\n %s Successfully created account on %s", output.OkEmoji(), networkName))
+	output.NextStepPrompt()
+
+	log.Info(fmt.Sprintf("\n %s Saving created account in flow.json", output.WarningEmoji()))
+	log.StartProgress("")
+	time.Sleep(time.Second * 3)
+	log.StopProgress()
```
sideninja, line 272: I don't think this delay is better, I think good UX should always be as quick as possible and if these steps don't require any interactions there is no need to slow them down.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+			name, privateJsonFileName))
+		if saveToGitIgnore {
+			log.Info(fmt.Sprintf("%s %s added to .gitignore", output.OkEmoji(), privateJsonFileName))
+		}
+	}
+
+	log.StartProgress("")
+	time.Sleep(time.Second * 5)
+	log.StopProgress()
+	log.Info("\n------------------------")
```
sideninja, line 290: I don't like this separator, it's never used in the CLI.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	if selectedNetwork != config.DefaultEmulatorNetwork() {
+		log.Info(fmt.Sprintf("%s private key for %s successfully saved in %s", output.OkEmoji(),
+			name, privateJsonFileName))
+		if saveToGitIgnore {
+			log.Info(fmt.Sprintf("%s %s added to .gitignore", output.OkEmoji(), privateJsonFileName))
+		}
+	}
+
+	log.StartProgress("")
+	time.Sleep(time.Second * 5)
```
sideninja, line 288: Same as above for delays.

sideninja:
Left comments on the specific things. This is a general comment on the output. It looks a bit crowded and messy, there's a lot of inconsistency in spacing and indenting that makes this screen hard to glance.
```
🟢 What name would you like to give this new account?
Enter name: Foo
✔ Local Emulator
⚠️ This command will perform the following:

- Generate a new ECDSA P-256 public and private key pair

- Create a new account on Local Emulator and pair the public key with the new account

- Save the newly created account configuration to flow.json

✔ CONTINUE
⚠️ Generating public and private keys for Foo
✅ Successfully generated public and private keys
✔ CONTINUE
⚠️ Creating the account on Local Emulator with generated keys

Please note, that the newly created account will only be available until you keep the emulator service up and running, if you restart the emulator service all accounts will be reset. If you want to persist accounts between restarts you must use the '--persist' flag when starting the flow emulator.

 ✅ Successfully created account on Local Emulator
✔ CONTINUE

 ⚠️ Saving created account in flow.json
✅ Successfully saved account in flow.json

------------------------
🟢 Account creation summary:

 ✅ Generated a new ECDSA P-256 key pair

 ✅ Created a new account on Local Emulator and paired the public key with the new account

 	Account name: Foo

 	Address: 0x01cf0e2f2f715450

 ✅ Saved the newly created account configuration to flow.json
```

Here's a better example how the screen could look (the ticks and logs after each bullet point would be added as they are created).
```
What would you like to name the new account: Foo

Select the network:
✔ Local Emulator

This command needs to perform the actions (press <ENTER> to continue):
- Generate a new ECDSA P-256 public and private key pair 
   ✅ Keys generated, public key: '0x12312312312312312312412412'
- Create a new account on Local Emulator and pair the public key with the new account 
   ✅ Account created, address: '0x424528171245' with a name: 'Foo' 
- Save the newly created account configuration to flow.json 
   ✅ Account saved to 'flow.json' 

Please note, that the newly created account will only be available until you keep the emulator service up and running, 
if you restart the emulator service all accounts will be reset. If you want to persist accounts between restarts 
you must use the '--persist' flag when starting the flow emulator.
```

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	err = saveAccount(loader, state, account, selectedNetwork, saveToGitIgnore)
+	if err != nil {
+		return nil, err
+	}
+	log.Info(fmt.Sprintf("%s Successfully saved account in flow.json", output.OkEmoji()))
+	if selectedNetwork != config.DefaultEmulatorNetwork() {
+		log.Info(fmt.Sprintf("%s private key for %s successfully saved in %s", output.OkEmoji(),
+			name, privateJsonFileName))
+		if saveToGitIgnore {
+			log.Info(fmt.Sprintf("%s %s added to .gitignore", output.OkEmoji(), privateJsonFileName))
```
sideninja, line 283: This is not working correctly, I choose "No" in the gitignore prompt but it was still displayed:
```
	- Testnet-Bob.private.json added to .gitignore
```
But it wasn't added to gitignore.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	account.EnableAdvancedSaveFormat()
+	privateState := flowkit.NewEmptyState(loader)
+	privateState.Accounts().AddOrUpdate(account)
+
+	err := privateState.Save(fileName)
+	if err != nil {
+		return err
+	}
+	if saveToGitIgnore {
+		addToGitIgnore(fileName)
```
sideninja, line 420: Not handling errors.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+	filename string,
+) error {
+	currentWd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	gitIgnoreDir := path.Join(currentWd, ".gitignore")
+
+	//opens .gitignore so that we can add {name}.private.json to .gitignore, will create .gitignore if it doesn't exist
+	file, err := os.OpenFile(gitIgnoreDir, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
sideninja, line 436: For writing or loading files you should use the loader that is being passed to commands so we could have tests for this.

pkg/flowkit/account.go
```diff
@@ -28,11 +28,82 @@ import (
...
+	}
+}
+
+// NewAccountFromOnChainAccount creates a new flowkit account definition
+// that mirrors an already-existing on-chain Flow account.
+//
+// This function requires the on-chain account to have exactly one public key
+// with full signing weight (1000). This ensures that the user has complete
+// and sole control over the on-chain account.
+func NewAccountFromOnChainAccount(
```
sideninja, line 63: Nice

pkg/flowkit/account.go
```diff
@@ -28,11 +28,82 @@ import (
...
 	address flow.Address
 	key     AccountKey
+
+	// fromFile is the configuration file containing this account.
+	//
+	// This field is only set if the external "fromFile"
+	// syntax is used. Otherwise this field is empty.
+	//
+	// Ref: https://docs.onflow.org/flow-cli/security/#private-account-configuration-file
+	fromFile string
```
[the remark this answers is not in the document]
sideninja: I feel like that could be part of the state / config domain. You could set an account and its location, and during serialization that would be handled, that way we also don't need to implement saving a new state to another file outside of the config domain. @psiemens

sideninja:
So we removed the encryption altogether? I thought we would leave it just to extract the account into a separate file. Any thoughts on this?

sideninja:
Another general note/question is: since we dropped the encryption, I feel we should always add that newly created file to gitignore, it shouldn't be an option, because if the user has git (which I would say would need to be checked programmatically) then I don't see a case where they would want their accounts being committed.

pkg/flowkit/account.go
```diff
@@ -28,11 +28,82 @@ import (
...
+	// useAdvancedSaveFormat is to save an account in advanced format even if it has default values as defined
+	// in isDefaultKeyFormat() in flowkit/config/json/account.go
+	// defaults to false
+	useAdvancedSaveFormat bool
```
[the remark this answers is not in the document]
psiemens: When reviewing the UX, we thought that the contents of the `{name}.private.json` file were hard to understand for newcomers to Flow. In particular, it's not clear that the single "key" field is a private key. The implicit defaults also (ECDSA_P256, SHA3_256) aren't obvious. So we wanted to save the account in advanced format to make those things clear (we can revisit this decision, curious to get your opinion). I think the convenience is worth it on the emulator, but for production mainnet keys, the explicit format feels safer. 

When implementing that, we also noticed that if a user manually adds a key with default values in advanced format, the config module will re-save it in simplified format. I'm not sure if that's intentional, but it felt undesirable to automatically reformat a user's key.

All that said, I agree that this crosses domain boundaries. Maybe we don't need this, but if we do, I'm a bit stuck on how to implement it cleanly.

pkg/flowkit/account.go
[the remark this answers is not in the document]
psiemens: Oh I like that a lot. 

So this `fromFile` field could change to something like `location`. If `location != ""`, then the config package saves the account to that location and adds a `{ "fromFile": "..."}` entry to the main config. Is that what you're thinking?

And I guess the `location` filepath would be resolved relative to the location of `flow.json`.

pkg/flowkit/output/prompt.go
sukantoraymond: I think we should just use dictionary to address the index problem

pkg/flowkit/account.go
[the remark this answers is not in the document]
sideninja: > When implementing that, we also noticed that if a user manually adds a key with default values in advanced format, the config module will re-save it in simplified format. I'm not sure if that's intentional, but it felt undesirable to automatically reformat a user's key.

That was a conscious change, but I was on the fence I remember. I decided to do it that way to really migrate users to the new format. 

In my mind, we don't need to save default values, since we are consistently not doing that anymore, but I really feel this is subjective, so I don't want to push it. In my mind, if a user is during the process explained what the algos were he should understand that, but I even think that users won't really understand this, since they are using the "beginner friendly" method for account creation. 

If we decide to keep it we can implement it along with the change I suggested about saving to a separate file. That one is https://github.com/onflow/flow-cli/pull/538#discussion_r921993732 and I think that suggestion can work for this too, because state will know which accounts is saving separately and it can set that property on the config/account (which is fine if it's there) but I would remove it on flowkit/account.

internal/accounts/create.go
```diff
@@ -134,3 +147,301 @@ func create(
...
+			output.NextStepPrompt()
+			link = util.TestnetFaucetURL(key.PublicKey().String(), crypto.ECDSA_P256)
+		case config.DefaultMainnetNetwork():
+			log.Info(fmt.Sprintf("%s We will be creating the account on Flow Port website (%s), which has been prefilled with generated keys", output.WarningEmoji(), util.FlowPortUrl))
+			log.Info("\nPlease follow the steps: \n 1. Click on 'Submit' button, \n 2. Connect existing Blocto or create a new account first, \n 3. Click on confirm, \n 4. Click on approve. \n")
+			output.NextStepPrompt()
+			link = util.MainnetFlowPortURL(key.PublicKey().String())
+		}
+		log.StartProgress("Waiting for an account to be created, please finish all the steps in the browser...\n")
+		time.Sleep(time.Second * 2)
```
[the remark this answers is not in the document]
sukantoraymond: address

internal/accounts/create.go
sukantoraymond: Removing git ignore prompt so this is not necessary anymore

pkg/flowkit/output/prompt.go
sukantoraymond: addressed

pkg/flowkit/output/prompt.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

pkg/flowkit/account.go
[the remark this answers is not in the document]
sideninja: Not necessarily, you could create an account normally and then set on the state location for this account, so the state would have the locations of accounts (or at least non-default ones). I feel that way account stays unaware of persistence details whereas state includes those (which is within the domain).

internal/accounts/create.go
sukantoraymond: addressed

pkg/flowkit/state.go
```diff
@@ -88,12 +107,12 @@ func (p *State) SaveEdited(paths []string) error {
...
 	}
 
-	return p.Save(paths[0])
+	return p.Save(paths[0], p.accountLocations)
 }
 
 // Save saves the project configuration to the given path.
-func (p *State) Save(path string) error {
-	p.conf.Accounts = accountsToConfig(*p.accounts)
+func (p *State) Save(path string, accountLocations map[string]string) error {
```
sideninja, line 114: this shouldn't be an input arg but can just be accessed from the state reference like `p.accountLocations` in the method.

pkg/flowkit/state.go
```diff
@@ -51,10 +51,29 @@ type Contract struct {
...
 type State struct {
-	conf         *config.Config
-	confLoader   *config.Loader
-	readerWriter ReaderWriter
-	accounts     *Accounts
+	conf             *config.Config
+	confLoader       *config.Loader
+	readerWriter     ReaderWriter
+	accounts         *Accounts
+	accountLocations map[string]string
```
sideninja, line 58: I would make this be a map with accounts, so:
```
accountLocations map[*Account]string
```
And then all the changes that come with this

sideninja:
@srinjoyc if accounts are created using flow port they don't have enough funds to even send transactions. Shouldn't they be funded?
Also @psiemens during testing I noticed something really weird. The method that creates an account from the network, sometimes (maybe 10% of times) fails the check that the account only has one key. Sometimes that method returns bunch of public keys, the ones added are pretty random, but if you take a look on the network manually it's only one key on the account. I have a suspicion it might be some bug on AN. I plan to investigate.

pkg/flowkit/config/account.go
```diff
@@ -30,6 +30,15 @@ type Account struct {
...
 	Key     AccountKey
+
+	// Location is the configuration file containing this account.
+	//
+	// This field is only set if the external "location"
+	// syntax is used. Otherwise this field is empty.
+	//
+	// Ref: https://docs.onflow.org/flow-cli/security/#private-account-configuration-file
+	Location         string
+	UseAdvanceFormat bool
```
psiemens, line 41:
```suggestion
	UseAdvancedFormat bool
```

"Advance format" isn't quite valid grammar

internal/command/command.go
```diff
@@ -282,7 +289,7 @@ func initCrashReporting() {
 			if output.ReportCrash() {
 				return event
 			} else {
-				fmt.Printf("\nPlease help us improve the Flow CLI by opening an issue on https://github.com/onflow/flow-cli/issues, \nand pasting the output as well as description of the actions you took to cause this crash.\n\n")
+				fmt.Printf("\nPlease help us improve the Flow CLI by opening an issue on https://github.com/onflow/flow-cli/issues, \nand pasting the output as well as description of the actions you took that resulted in this crash.\n\n")
```
psiemens, line 292:
```suggestion
				fmt.Printf("\nPlease help us improve the Flow CLI by opening an issue on https://github.com/onflow/flow-cli/issues, \nand pasting the output as well as a description of the actions you took that resulted in this crash.\n\n")
```

internal/accounts/create.go
```diff
@@ -134,3 +145,218 @@ func create(
...
+	items = items[1:]
+	for n, item := range items {
+		sep := " -"
+		if numbered {
+			sep = fmt.Sprintf(" %d.", n+1)
+		}
+		log.Info(fmt.Sprintf("%s %s", sep, item))
+	}
+	log.Info("")
+}
```
psiemens, line 362: Should this go in the `output` package?

sideninja: It could yeah. Wil move it. Generally, I was thinking about refactoring this interactive UI into a separate file since more commands will introduce this mode and it could be nicer to have an interactive UI component handling that. Not sure yet exactly how, since I want to wait a bit to see the pattern evolving but I have in mind to refactor this down the line, I don't like the command itself becoming bloated with these interactive responses.

psiemens (approved):
Looks good to me overall.

That public key check bug is strange. It sounds like a data consistency issue coming from the AN… did you get a chance to investigate?

sideninja:
> Looks good to me overall.
> 
> That public key check bug is strange. It sounds like a data consistency issue coming from the AN… did you get a chance to investigate?

Yep investigated and fixed. It was in fact a bug in CLI as seen here: https://github.com/onflow/flow-cli/pull/538/commits/71f9fd66df477e76fbd216e8afddff9b19b06501
The problem was that since get address was used before only on a transaction that made a new account it didn't matter which address it returned so it did that on all events in that tx, whereas now we monitor more events and it's important that after we get the right event we extract address from that event not on all event range.

Issue #493, opened 2022-05-04
Refactor Flowkit APIs


bjartek:
## Instructions

Currently flowkit does two things that I do not belive that a pure library should do. 
 1. It uses a spinner logger that does not work well if you use flowkit as the base of a damon. (I know my syslog is ugly because of it)
 2. It panics in some places and do not return errors.

### Issue To Be Solved

Flowkit should fix the two above issues and ideally use the same logging framework as the other flow based go projects like the go-sdk uses (logrus)

bjartek:
Another option is to make it possible for StartProgressing to _not_ spin.

sideninja:
Very good feedback, I also had in mind to do some refactoring of flowkit in the future and make it more official, some of the things are there just because how it was developed, but I strongly agree with both points and there might be even more to better define the APIs etc.

bjartek:
Also context should be able to be sent in if that is appropriate. I want to control when to cancel things in my code.

sideninja:
> Also context should be able to be sent in if that is appropriate. I want to control when to cancel things in my code.

100%, it's on my to-do list to refactor API to take the context as the first argument as it is idiomatic in Go. My mistake making that in the first place when starting I wasn't as experienced in Go.

sideninja:
Also @bjartek I think this might be a great thread to list all the changes required to the API and then just make those in a release that would make the flowkit beta. I now see it as an alpha. 

Some other ideas:
- `Accounts.Create` API change: refactor `contractArgs []string` argument to just take in paths as names can be extracted from the source, or change to typed contract struct that defines the name and filepath
- `Blocks.GetBlock` API change: refactor `query string` to be an enum, similar to how that was done in the Go SDK
- `Blocks.GetLatestBlockHeight` API remove: potentially remove this since it can be achieve with the `GetBlock` method
- Idea: create event type which would include factory to create event types, so instead of passing values like `A.0x1.Foo` we could have:
```go
NewEvent(ForAccount(0x1), ForName("Foo"), ForNetwork(mainnet))
// or helper methods for native events
NewAccountCreatedEvent(ForNetwork(mainnet))
```

Will continue with more soon.

bjartek:
Since flowkit is the bridge between flow and other go programs could we consider making some code to parse events into something that is easier for go to read? I have some code in overflow for this but it currently outputs everything in the values as strings and that is not ideal. 

In general I think a session where we go through overflow and see what should be ported down and what should just stay in overflow could be useful.

[...]

PR #550, opened 2022-06-30
Move flowkit package to its own module

dsainati1:
This separates the flowkit package into its own module at `github.com/onflow/flow-cli/pkg/flowkit`. This way, the Cadence language server can depend on the flowkit without introducing a circular dependency between the CLI and the language server. 
______

For contributor use:


go.mod
```diff
@@ -1,27 +1,166 @@
...
+	google.golang.org/grpc v1.45.0 // indirect
+	google.golang.org/protobuf v1.28.0 // indirect
+	gopkg.in/ini.v1 v1.66.2 // indirect
+	gopkg.in/warnings.v0 v0.1.2 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	lukechampine.com/blake3 v1.1.7 // indirect
+)
+
+replace github.com/onflow/flow-cli/pkg/flowkit => ./pkg/flowkit
```
turbolent, line 166: This shouldn't be committed. I guess we need it temporarily until we can point to a commit?

turbolent (approved):
Modularizing flowkit would be great, as we can make use of it in various tools, e.g. in the future in the emulator for the debugger.

In addition to splitting it up inside of this repo, maybe we can also consider moving it to the Go SDK or a separate repo?

turbolent:
@dsainati1 Please wait until @sideninja had a chance to look at it

go.mod
dsainati1: Yeah; until we have a commit where this package exists I think we need this for now.

dsainati1:
> In addition to splitting it up inside of this repo, maybe we can also consider moving it to the Go SDK

We could consider moving this to another repo, but the SDK would not work, I think, because flowkit depends on the emulator. At the very least it needs to be higher up in the dependency chain than that.

bjartek:
What about putting it in a repo of its own?

bluesign:
agreed with @bjartek, I think it is better to separate into new repo too. 

In general it is a bit mess, especially fvm - flow-go , SDK - flow-go etc. I think need to think all ( not focused only here, and should make a re-org )

dsainati1:
Agreed, but this PR both makes that easier achieve and unblocks me on something else I am working on. We can follow up in another PR to make this its own repo.

psiemens (commented):
Gregor and I had talked about eventually moving flowkit to its own repository. I think this is a good first step!

Gregor's out until the 14th. @sukantoraymond and I can help merge this, but I'd like to wait until Wednesday as we are planning to cut a release today/tomorrow and would need a bit more time to test the upgrade to Go 1.18. @dsainati1 does that work for you?

dsainati1:
Yeah I can wait until then no problem.

psiemens (approved):

sideninja:
I would say that extracting flowkit more would be conditioned on making sure we refactor it in places according to this https://github.com/onflow/flow-cli/issues/493
It has been my plan to make these steps, but I feel we should first make sure it works like a well-behaved library.

PR #564, opened 2022-07-13
added new multisig sign command 

bthaile:
Added new flag `--from-remote-url` to indicate the argument passed into transactions sign command is a fully qualified url to transaction RLP. 

______

For contributor use:

### DIFF
--- a/docs/sign-transaction.md
+++ b/docs/sign-transaction.md
@@ -65,13 +65,13 @@ f90184f...a199bfd9b837a11a0885f9104b54014750f5e3e5bfe4a5795968b0df86769dd54c0
 ## Arguments
 
 ### Built Transaction Filename or Remote Server URL
-- Name: `built transaction filename | --remote-server-url <url>`
-- Valid inputs: Any filename and path valid on the system or fully qualified remote server url.
+- Name: `built transaction filename | --from-remote-url <url>`
+- Valid inputs: Any filename and path valid on the system or --from-remote-url flag and fully qualified remote server url.
 
 Specify the filename containing valid transaction payload that will be used for signing.
 To be used with the `flow transaction build` command.
 
-When --from-remote-url flag is used the argument needs to be a fully qualified url to transaction RLP
+When --from-remote-url flag is used the value needs to be a fully qualified url to transaction RLP
 Example: `flow transaction sign --from-remote-url https://fully/qualified/url --signer alice`
 ## Flags
 
@@ -80,7 +80,7 @@ Example: `flow transaction sign --from-remote-url https://fully/qualified/url --
 - Valid input: `http(s)://fully/qualified/server/url`
 
 Specify this flag with a fully qualified url to transaction RLP. The RLP will be fetched from server then signed. The resulting signed RLP is then posted to the remote url. This feature is to support protocol level multiple signature transaction coordination between multiple signers.
-Note: --yes flag is not supported and will fail `sign` command when this flag is used. 
+Note: --yes flag is not supported and will fail `sign` command when this flag is used. This forces the user to verify the cadence code.
 
 ### Include Fields
 
--- a/docs/sign-transaction.md
+++ b/docs/sign-transaction.md
@@ -64,14 +64,23 @@ f90184f...a199bfd9b837a11a0885f9104b54014750f5e3e5bfe4a5795968b0df86769dd54c0
 
 ## Arguments
 
-### Built Transaction Filename
-- Name: `built transaction filename`
-- Valid inputs: Any filename and path valid on the system.
+### Built Transaction Filename or Remote Server URL
+- Name: `built transaction filename | --remote-server-url <url>`
+- Valid inputs: Any filename and path valid on the system or fully qualified remote server url.
 
 Specify the filename containing valid transaction payload that will be used for signing.
 To be used with the `flow transaction build` command.
 
+When --from-remote-url flag is used the argument needs to be a fully qualified url to transaction RLP
+Example: `flow transaction sign --from-remote-url https://fully/qualified/url --signer alice`
 ## Flags
+
+### From Remote Url
+- Flag: `--from-remote-url`
+- Valid input: `http(s)://fully/qualified/server/url`
+
+Specify this flag with a fully qualified url to transaction RLP. The RLP will be fetched from server then signed. The resulting signed RLP is then posted to the remote url. This feature is to support protocol level multiple signature transaction coordination between multiple signers.
+Note: --yes flag is not supported and will fail `sign` command when this flag is used. 
 
 ### Include Fields
 
--- a/internal/transactions/sign.go
+++ b/internal/transactions/sign.go
@@ -26,22 +26,24 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flowkit/output"
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
 )
 
 type flagsSign struct {
-	Signer  string   `default:"emulator-account" flag:"signer" info:"name of the account used to sign"`
-	Include []string `default:"" flag:"include" info:"Fields to include in the output. Valid values: signatures, code, payload."`
+	Signer        string   `default:"emulator-account" flag:"signer" info:"name of the account used to sign"`
+	Include       []string `default:"" flag:"include" info:"Fields to include in the output. Valid values: signatures, code, payload."`
+	FromRemoteUrl string   `default:"" flag:"from-remote-url" info:"server URL where RLP can be fetched, signed RLP will be posted back to remote URL."`
 }
 
 var signFlags = flagsSign{}
 
 var SignCommand = &command.Command{
 	Cmd: &cobra.Command{
-		Use:     "sign <built transaction filename>",
+		Use:     "sign [<built transaction filename> | --from-remote-url <url>]",
 		Short:   "Sign built transaction",
 		Example: "flow transactions sign ./built.rlp --signer alice",
-		Args:    cobra.ExactArgs(1),
+		Args:    cobra.MaximumNArgs(1),
 	},
 	Flags: &signFlags,
 	RunS:  sign,
@@ -54,10 +56,26 @@ func sign(
 	services *services.Services,
 	state *flowkit.State,
 ) (command.Result, error) {
-	filename := args[0]
-	payload, err := readerWriter.ReadFile(filename)
+	var payload []byte
+	var err error
+	var filenameOrUrl string
+
+	if signFlags.FromRemoteUrl != "" {
+		if globalFlags.Yes {
+			return nil, fmt.Errorf("--yes is not supported with this flag")
+		}
+		filenameOrUrl = signFlags.FromRemoteUrl
+		payload, err = services.Transactions.GetRLP(filenameOrUrl)
+	} else {
+		if len(args) == 0 {
+			return nil, fmt.Errorf("filename argument is required")
+		}
+		filenameOrUrl = args[0]
+		payload, err = readerWriter.ReadFile(filenameOrUrl)
+	}
+
 	if err != nil {
-		return nil, fmt.Errorf("failed to read partial transaction from %s: %v", filename, err)
+		return nil, fmt.Errorf("failed to read partial transaction from %s: %v", filenameOrUrl, err)
 	}
 
 	signer, err := state.Accounts().ByName(signFlags.Signer)
@@ -68,6 +86,16 @@ func sign(
 	signed, err := services.Transactions.Sign(signer, payload, globalFlags.Yes)
 	if err != nil {
 		return nil, err
+	}
+
+	if signFlags.FromRemoteUrl != "" {
+		tx := signed.FlowTransaction()
+		err = services.Transactions.PostRLP(filenameOrUrl, tx)
+
+		if err != nil {
+			return nil, err
+		}
+		fmt.Printf("%s Signed RLP Posted successfully\n", output.SuccessEmoji())
 	}
 
 	return &TransactionResult{
--- a/internal/transactions/sign.go
+++ b/internal/transactions/sign.go
@@ -60,6 +60,10 @@ func sign(
 	var err error
 	var filenameOrUrl string
 
+	if signFlags.FromRemoteUrl != "" && len(args) > 0 {
+		return nil, fmt.Errorf("only use one, filename argument or --from-remote-url <url>")
+	}
+
 	if signFlags.FromRemoteUrl != "" {
 		if globalFlags.Yes {
 			return nil, fmt.Errorf("--yes is not supported with this flag")
--- a/pkg/flowkit/services/transactions.go
+++ b/pkg/flowkit/services/transactions.go
@@ -19,7 +19,11 @@
 package services
 
 import (
+	"bytes"
+	"encoding/hex"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 
 	"github.com/onflow/flow-cli/pkg/flowkit"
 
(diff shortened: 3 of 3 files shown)


bluesign:
I think `--yes`  this should not be an option here.

docs/sign-multisig.md
```diff
@@ -0,0 +1,96 @@
...
+The Flow CLI provides a command to sign a multsi transaction with options to specify
+signer account.
+
+Use this functionality after a multisig transaction has been generated:
+ - When a multisig transaction the keys needed to sign have already been identified
+ - This signing method will first pull down the transaction RLP from url to allow the user to sign
+ - After the transaction has been signed the resulting signed RLP will be posted to the server
+
+```shell
+flow multisig sign <transaction id ulr>
```
Kay-Zee, line 16: What was the thought process behind adding a new sub command, rather than having it as part of the `sign` sub command, and just signalling via a flag, i.e. `--remote-url` or similar?

internal/multisig/sign.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error downloading multisig identifier")
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
Kay-Zee, line 121: nitpick: can be `return ioutil.ReadAll(resp.Body)`

internal/multisig/sign.go
```diff
@@ -0,0 +1,142 @@
...
+	return &MultisigResult{
+		tx:     tx,
+		rlp:    string(payload),
+		signed: b.String(),
+	}, err
+}
+
+func retrieve(rlpUrl string) ([]byte, error) {
+
+	if rlpUrl == "" {
```
Kay-Zee, line 101: is this necessary? I feel like we checked before passing to the function, and sending a get to and empty URL will also error anyways.

internal/multisig/sign.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	tx := signed.FlowTransaction()
+	msg := tx.Encode()
+
+	var b bytes.Buffer
+	writer := util.CreateTabWriter(&b)
```
Kay-Zee, line 86: why do we need a tab writer here?

internal/multisig/sign.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func sign(
+	args []string,
+	readerWriter flowkit.ReaderWriter,
+	globalFlags command.GlobalFlags,
+	services *services.Services,
+	state *flowkit.State,
+) (command.Result, error) {
+	multisig_url := args[0]
```
Kay-Zee, line 61:
```suggestion
	multisigURL := args[0]
```

nitpick: golang prefers camel case

internal/multisig/sign.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/pkg/flowkit/services"
+	"github.com/onflow/flow-cli/pkg/flowkit/util"
+)
+
+type flagsSign struct {
+	Signer  string   `default:"emulator-account" flag:"signer" info:"name of the account used to sign"`
+	Include []string `default:"" flag:"include" info:"Fields to include in the output. Valid values: signatures, code, payload."`
```
Kay-Zee, line 38: Is this used?

Kay-Zee (commented):
Nice! thanks for doing this!

sideninja:
I'm missing context on the need for this functionality, so if you can quickly explain the use case it would be helpful. 
Another note on my assumption of this usage would be, I'm not sure if this makes sense to be a top command, looking at the implementation it looks like it does all the same as the `flow transactions sign` command, except for the part where it gets and uploads the transaction from a server, wouldn't then make sense for this to just be a flag on that mentioned command?

sideninja (changes requested):
I'm missing context on the need for this functionality, so if you can quickly explain the use case it would be helpful.
Another note based from my assumption of this usage would be, I'm not sure if this makes sense to be a top command, looking at the implementation it looks like it does all the same as the flow transactions sign command, except for the part where it gets and uploads the transaction from a server, wouldn't then make sense for this to just be a flag on that mentioned command? It also feels a bit tailored towards a specific usecase we might have (although I might be wrong since I lack context), but wouldn't this be possible to do with chaining/piping the commands like curl in order to do the same?

bluesign:
> wouldn't then make sense for this to just be a flag on that mentioned command? 

I think even we can make transaction sign command to accept both file and url , if it is URL it can get RLP from URL and post result

docs/sign-multisig.md
bthaile: My thinking was "multisig" is a new concept that might grow so wanted to keep it separate from existing commands.

internal/multisig/sign.go
bthaile: not needed, will remove

bthaile:
@sideninja 

This additional command is to better coordinate service account signers for system transactions which can have many signers, currently there is a bash script used to get and post to server which isn't ideal. Having this functionality baked into the flow cli will help.

sideninja:
> @sideninja
> 
> This additional command is to better coordinate service account signers for system transactions which can have many signers, currently there is a bash script used to get and post to server which isn't ideal. Having this functionality baked into the flow cli will help.

I understand. I feel this doesn't really make sense as a new top command tho, I feel it would be better as an option on the already existing sign command, either with a flag or even better as an argument and then the command would recognize the value as URL and handle that.

bthaile:
Looks like the sentiment is to add a new flag and not a top level command. I'll make that change. Thanks for the feedback.

bluesign:
> "service account signers for system transactions which can have many signers"

I once again repeat my warning about `--yes` flag in this context. In case web site is compromised, it is really big risk.

bthaile:
Updated PR to add `--from-remote-url` flag to existing transactions sign command. This flag will turn off --yes to force the user to view the tx they are signing.

internal/transactions/sign.go
```diff
@@ -54,22 +56,43 @@ func sign(
...
 
-	signed, err := services.Transactions.Sign(signer, payload, globalFlags.Yes)
+	signed, err := services.Transactions.Sign(signer, payload, yesOption)
 	if err != nil {
 		return nil, err
 	}
 
+	if signFlags.FromRemoteUrl {
+		tx := signed.FlowTransaction()
+		signedRlp := fmt.Sprintf("%x", string(tx.Encode()))
```
Kay-Zee, line 87: is this intended to stay in the final command? if so we may want to add more context?

internal/transactions/sign.go
```diff
@@ -54,22 +56,43 @@ func sign(
...
 	state *flowkit.State,
 ) (command.Result, error) {
-	filename := args[0]
-	payload, err := readerWriter.ReadFile(filename)
+	var payload []byte
+	var err error
+	yesOption := globalFlags.Yes
+	filenameOrUrl := args[0]
+
+	if signFlags.FromRemoteUrl {
```
Kay-Zee, line 64: I guess it's too un-reliable just to try and figure this out from the input itself? Probably safer with a flag, with only slightly worse UX. Probably worth the trade off, but wanted to mention it's probably possible.

Kay-Zee (approved):

docs/sign-transaction.md
```diff
@@ -64,15 +64,24 @@ f90184f...a199bfd9b837a11a0885f9104b54014750f5e3e5bfe4a5795968b0df86769dd54c0
...
+- Valid inputs: Any filename and path valid on the system or fully qualified remote server url.
 
 Specify the filename containing valid transaction payload that will be used for signing.
 To be used with the `flow transaction build` command.
 
+When --from-remote-url flag is used the argument needs to be a fully qualified url to transaction RLP
+Example: `flow transaction sign --from-remote-url https://fully/qualified/url --signer alice`
 ## Flags
 
+### From Remote Url
```
sideninja, line 78: <3

internal/transactions/sign.go
```diff
@@ -19,26 +19,29 @@
...
+	"github.com/onflow/flow-cli/pkg/flowkit/output"
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
 )
 
 type flagsSign struct {
-	Signer  string   `default:"emulator-account" flag:"signer" info:"name of the account used to sign"`
-	Include []string `default:"" flag:"include" info:"Fields to include in the output. Valid values: signatures, code, payload."`
+	Signer        string   `default:"emulator-account" flag:"signer" info:"name of the account used to sign"`
+	Include       []string `default:"" flag:"include" info:"Fields to include in the output. Valid values: signatures, code, payload."`
+	FromRemoteUrl bool     `default:"false" flag:"from-remote-url" info:"Indicates argument is a server Url where RLP can be fetched, signed RLP will be posted back to remote url."`
```
sideninja, line 37:
```suggestion
	FromRemoteUrl bool     `default:"false" flag:"from-remote-url" info:"Indicates argument is a server URL where RLP can be fetched, signed RLP will be posted back to remote URL."`
```

internal/transactions/sign.go
sideninja: I agree with above 100%, but I would add that we should avoid such usage of flags. I think it would be better suited to use the flag like: `flow transactions sign --from-remote-url <url>` instead of using it to indicate what the argument value is, usually flags are meant to either indicate a boolean value (with them being present or not) or to pass the value but doing so directly using the flag (as in my example), I think if you want to indicate the argument value is remote URL then the flag should be more like `--is-remove-url` but not sure if that is better, I would say it's better to just use the flag to provide the value and then make the argument value optional if that flag is present.

pkg/flowkit/services/transactions.go
```diff
@@ -276,3 +279,39 @@ func (t *Transactions) Send(
 
 	return sentTx, res, err
 }
+
+func (t *Transactions) GetRlp(rlpUrl string) ([]byte, error) {
```
sideninja, line 283:
```suggestion
func (t *Transactions) GetRLP(rlpUrl string) ([]byte, error) {
```

pkg/flowkit/services/transactions.go
```diff
@@ -276,3 +279,39 @@ func (t *Transactions) Send(
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error downloading RLP identifier")
+	}
+
+	return ioutil.ReadAll(resp.Body)
+}
+
+func (t *Transactions) PostRlp(rlpUrl string, signed string) error {
```
sideninja, line 304:
```suggestion
func (t *Transactions) PostRLP(rlpUrl string, signed string) error {
```

sideninja, line 304: I would accept the `Transaction` as an argument instead of string, this way you encapsulate the logic how to submit a transaction, because otherwise this is more of a function that knows how to send string to server than RLP for transaction.

internal/transactions/sign.go
```diff
@@ -54,22 +56,47 @@ func sign(
...
-	payload, err := readerWriter.ReadFile(filename)
+	var payload []byte
+	var err error
+	yesOption := globalFlags.Yes
+	var filenameOrUrl string
+
+	if signFlags.FromRemoteUrl != "" {
+		filenameOrUrl = signFlags.FromRemoteUrl
+		payload, err = services.Transactions.GetRlp(filenameOrUrl)
+		yesOption = false // force user to view remote tx
```
janezpodhostnik, line 67: IMO it's better to error if `globalFlags.Yes` is true instead of silently ignoring it.

docs/sign-transaction.md
```diff
@@ -64,15 +64,24 @@ f90184f...a199bfd9b837a11a0885f9104b54014750f5e3e5bfe4a5795968b0df86769dd54c0
...
+- Flag: `--from-remote-url`
+- Valid input: `http(s)://fully/qualified/server/url`
+
+Specify this flag with a fully qualified url to transaction RLP. After signing the result is posted to the remote url.
```
janezpodhostnik, line 82: Maybe a bit more explanation would be useful here. Without the code I don't know if I would have figured out what this meant.

janezpodhostnik (commented):
Some minor comments, looks good otherwise.

bthaile:
@sideninja 
How to fix these linting errors? I'm not seeing them locally in vscode or when I run `make lint`

image

sideninja:
> @sideninja How to fix these linting errors on this PR? I'm not seeing them locally in vscode or when I run `make lint`
> 
> image

First, I would suggest you on using the GoLand IDE if you plan to work more in Go as it will make your life much easier in detecting problems like these. You can obtain a company license as well for it. 
The problem above is that you have renamed the methods to `PostRLP` but you are referencing them as `PostRlp` which is a syntax error.

bthaile:
@sideninja updated the PR but still `services.Transactions.GetRLP` and `services.Transactions.PostRLP` are not found. I loaded up flow-cli in goLand IDE and don't see any errors. Does the PR compile `services` package so that PostRLP and PostRLP exist when its imported in Transactions package?

sideninja:
> @sideninja updated the PR but still `services.Transactions.GetRLP` and `services.Transactions.PostRLP` are not found. I loaded up flow-cli in goLand IDE and don't see any errors. Does the PR compile `services` package so that PostRLP and PostRLP exist when its imported in Transactions package?

Right, the problem is with the fact flowkit was created as a module, I was always during that time so I forget. Can you just try removing the `github.com/onflow/flow-cli/pkg/flowkit` as a dependency in go.mod

sideninja:
> > @sideninja updated the PR but still `services.Transactions.GetRLP` and `services.Transactions.PostRLP` are not found. I loaded up flow-cli in goLand IDE and don't see any errors. Does the PR compile `services` package so that PostRLP and PostRLP exist when its imported in Transactions package?
> 
> Right, the problem is with the fact flowkit was created as a module, I was away during that time so I forget. I will take a look into that now. The replace is needed but there's another problem with dep now. Will push you a branch soon.

This is proving to be quite a messy situation so it might take us some more time to untangle. Me and @dsainati1 are looking into it, will keep you updated.

sideninja:
@bthaile we finally untangled the dependency problems, so as soon as the PR is merged your PR should work, just make sure you pull and merge from master branch. Should be merged in today in next hours https://github.com/onflow/flow-cli/pull/578

sideninja:
@bthaile please try now.

bthaile:
@sideninja  updated PR

bthaile:
@sideninja  I don't see any linting errors in goLand IDE. Also the linting errors in this PR have to do with files I didn't change.

sideninja:
> @sideninja I don't see any linting errors in goLand IDE. Also the linting errors in this PR have to do with files I didn't change.

Yes, there was a problem with the linter version in the CI. I have a PR that will fix that, so just ignore the error here.

docs/sign-transaction.md
```diff
@@ -64,15 +64,24 @@ f90184f...a199bfd9b837a11a0885f9104b54014750f5e3e5bfe4a5795968b0df86769dd54c0
...
-- Name: `built transaction filename`
-- Valid inputs: Any filename and path valid on the system.
+### Built Transaction Filename or Remote Server URL
+- Name: `built transaction filename | --remote-server-url <url>`
+- Valid inputs: Any filename and path valid on the system or fully qualified remote server url.
 
 Specify the filename containing valid transaction payload that will be used for signing.
 To be used with the `flow transaction build` command.
 
+When --from-remote-url flag is used the argument needs to be a fully qualified url to transaction RLP
```
sideninja, line 74: This flag is not named the same as in the example above.

internal/transactions/sign.go
```diff
@@ -54,10 +56,26 @@ func sign(
 	services *services.Services,
 	state *flowkit.State,
 ) (command.Result, error) {
-	filename := args[0]
-	payload, err := readerWriter.ReadFile(filename)
+	var payload []byte
+	var err error
+	var filenameOrUrl string
+
+	if signFlags.FromRemoteUrl != "" {
```
sideninja, line 63: validate that both are not passed.

sideninja (approved):
There's a couple of comments you should address but otherwise LGTM. Just make sure you manually test it well.

internal/transactions/sign.go
sideninja: Yes that's better, but what I mean is that I don't see any validation for using the command like: `flow transaction sign ./tx.rlp --from-remote-url HTTP://test.com` so testing both are not passed

bthaile: update, tested and pushed.

sideninja:
Feel free to merge @bthaile

PR #577, opened 2022-07-20
Account creation improvements

sideninja:
This PR improves the output of the account creation flow as discussed and also refactors how the accounts are being saved using the private location, so the handling of that doesn't escape the config layer.

The new output recording can be found here:
https://asciinema.org/a/y0ryNXT4OMn6HfbdN9bwShcbK

______

For contributor use:

### DIFF
--- a/internal/accounts/create.go
+++ b/internal/accounts/create.go
@@ -166,7 +166,7 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 		log.Info(fmt.Sprintf("- Save the private key to %s.", privateFile))
 	}
 	log.Info(fmt.Sprintf("- Create a new account on %s paired with the public key.", output.Bold(networkName)))
-	log.Info(fmt.Sprintf("- Save the newly-created account to %s.", output.Bold("flow.json")))
+	log.Info(fmt.Sprintf("- Save the newly-created account to %s.\n", output.Bold("flow.json")))
 
 	if !output.WantToContinue() {
 		return nil, fmt.Errorf("process terminated")
@@ -212,7 +212,7 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 		switch selectedNetwork {
 		case config.DefaultTestnetNetwork():
 			log.Info("Please complete the following steps in the browser:")
-			log.Info("\n 1. Complete the captcha challenge. \n 2. Click the 'Create Account' button.\n 3. Return to this window.")
+			log.Info("\n 1. Complete the captcha challenge. \n 2. Click the 'Create Account' button.\n 3. Return to this window.\n")
 			link = util.TestnetFaucetURL(key.PublicKey().String(), crypto.ECDSA_P256)
 
 		case config.DefaultMainnetNetwork():
@@ -249,19 +249,25 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 		return nil, err
 	}
 
-	log.Info(fmt.Sprintf("%s New account created with address: %s, and name: %s", output.SuccessEmoji(), account.Address(), name))
+	log.Info(fmt.Sprintf(
+		"%s New account created with address: %s, and name: %s",
+		output.SuccessEmoji(),
+		output.Bold(account.Address().String()),
+		output.Bold(name)),
+	)
 
 	err = saveAccount(loader, state, account, selectedNetwork)
 	if err != nil {
 		return nil, err
 	}
 
-	log.Info("🎉 Here’s a summary of all the actions that were taken:")
+	log.Info("\nHere’s a summary of all the actions that were taken:")
 	log.Info(fmt.Sprintf("- Added the new account to %s.", output.Bold("flow.json")))
 	if selectedNetwork != config.DefaultEmulatorNetwork() {
 		log.Info(fmt.Sprintf("- Saved the private key to %s.", privateFile))
 		log.Info(fmt.Sprintf("- Added %s to %s.", privateFile, output.Bold(".gitignore.")))
 	}
+	log.Info("") // final new line
 
 	return onChainAccount, nil
 }
--- a/internal/accounts/create.go
+++ b/internal/accounts/create.go
@@ -165,7 +165,7 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 		"Generate a new ECDSA P-256 public and private key pair.",
 	}
 	if selectedNetwork != config.DefaultEmulatorNetwork() {
-		items = append(items, fmt.Sprintf("Save the private key to %s.", privateFile))
+		items = append(items, fmt.Sprintf("Save the private key to %s and add it to .gitignore.", privateFile))
 	}
 	items = append(items,
 		fmt.Sprintf("Create a new account on %s paired with the public key.", output.Bold(networkName)),
--- a/internal/accounts/create.go
+++ b/internal/accounts/create.go
@@ -161,7 +161,7 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 	privateFile := output.Bold(fmt.Sprintf("%s.private.json", name))
 
 	items := []string{
-		fmt.Sprintf("%s This command will perform the following", output.WarningEmoji()),
+		fmt.Sprintf("%sThis command will perform the following", output.WarningEmoji()),
 		"Generate a new ECDSA P-256 public and private key pair.",
 	}
 	if selectedNetwork != config.DefaultEmulatorNetwork() {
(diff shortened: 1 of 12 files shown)


sukantoraymond:
This change doesn't save the account as advanced format

sideninja:
> This change doesn't save the account as advanced format

Yeah, that I intentionally removed. Unless @psiemens feels strongly about keeping I would rather be consistent. It's an easy add tho.

psiemens:
> Yeah, that I intentionally removed. Unless @psiemens feels strongly about keeping I would rather be consistent. It's an easy add tho.

I prefer the advanced format but I don't feel strongly about keeping it. I'd say let's get the feature out and then later decide if you want to add that change.

internal/accounts/create.go
```diff
@@ -221,21 +211,22 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
...
-		time.Sleep(time.Second * 2)
+
+		output.ConfirmOpenBrowser()
+
+		log.StartProgress("Waiting for your account to be created, please finish all the steps in the browser...\n")
 		err := util.OpenBrowserWindow(link)
 		if err != nil {
-			return nil, err
+			log.Info(fmt.Sprintf("Could not open a browser window, please navigate to: %s manually", link))
 		}
```
psiemens, line 230: Do we know with certainty that this will return an error if the browser can't open? I'm worried about a case where the browser won't open (i.e. due to OS permissions) and this fails silently.

internal/accounts/create.go
```diff
@@ -256,21 +247,27 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 	if err != nil {
 		return nil, err
 	}
-	log.Info(fmt.Sprintf("\n %s Successfully created account on %s", output.OkEmoji(), networkName))
-	output.NextStepPrompt()
+
+	log.Info(fmt.Sprintf(
+		"%s New account created with address: %s, and name: %s",
```
psiemens, line 252:
```suggestion
		"%s New account created with address 0x%s and name %s",
```

No need for the extra punctuation IMO.

internal/accounts/create.go
```diff
@@ -256,21 +247,27 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
...
-	log.Info(fmt.Sprintf("%s Successfully saved account in flow.json", output.OkEmoji()))
+
+	log.Info("\nHere’s a summary of all the actions that were taken:")
+	log.Info(fmt.Sprintf("- Added the new account to %s.", output.Bold("flow.json")))
 	if selectedNetwork != config.DefaultEmulatorNetwork() {
-		log.Info(fmt.Sprintf("%s private key for %s successfully saved in %s", output.OkEmoji(),
-			name, privateJsonFileName))
-		log.Info(fmt.Sprintf("%s %s added to .gitignore", output.OkEmoji(), privateJsonFileName))
+		log.Info(fmt.Sprintf("- Saved the private key to %s.", privateFile))
+		log.Info(fmt.Sprintf("- Added %s to %s.", privateFile, output.Bold(".gitignore.")))
```
psiemens, line 267:
```suggestion
		log.Info(fmt.Sprintf("- Added %s to %s.", privateFile, output.Bold(".gitignore")))
```

Extra period :)

internal/accounts/create.go
```diff
@@ -151,29 +149,28 @@ func create(
...
 	gw, err := gateway.NewGrpcGateway(selectedNetwork.Host)
 	if err != nil {
 		return nil, err
 	}
-	privateJsonFileName := ""
+
+	privateFile := output.Bold(fmt.Sprintf("%s.private.json", name))
+
+	log.Info(fmt.Sprintf("\n%s This command will perform the following:", output.WarningEmoji()))
+	log.Info(fmt.Sprintf("- Generate a new ECDSA P-256 public and private key pair."))
```
psiemens, line 164: Super nitpicky, but I like that the numbered steps below are indented with one space. It'd be good to make these ticks match (and same with the final summary).
Screen Shot 2022-07-20 at 1 21 43 PM

pkg/flowkit/state.go
```diff
@@ -107,14 +106,28 @@ func (p *State) SaveEdited(paths []string) error {
...
 }
 
 // Save saves the project configuration to the given path.
-func (p *State) Save(path string, accountLocations map[string]string) error {
-	p.conf.Accounts = accountsToConfig(*p.accounts, accountLocations)
+func (p *State) Save(path string) error {
+	p.conf.Accounts = accountsToConfig(*p.accounts, p.confLoader.AccountsFromFile())
 	err := p.confLoader.Save(p.conf, path)
 
+	// if we have defined accounts to be saved to private file iterate over them and save them separately
```
psiemens, line 117:
```suggestion
	// if we have defined accounts to be saved to an external file, iterate over them and save them separately
```

This feature isn't just for private files, so don't want to confuse future maintainers

pkg/flowkit/util/utilities.go
```diff
@@ -154,3 +156,45 @@ func TestnetFaucetURL(publicKey string, sigAlgo crypto.SignatureAlgorithm) strin
...
+	}
+	gitIgnoreDir := path.Join(currentWd, ".gitignore")
+
+	gitIgnoreFiles := ""
+	if DirExists(gitIgnoreDir) {
+		gitIgnoreFilesRaw, err := loader.ReadFile(gitIgnoreDir)
+		gitIgnoreFiles = string(gitIgnoreFilesRaw)
+		if err != nil {
+			return err
+		}
```
psiemens, line 190:
```suggestion
		if err != nil {
			return err
		}
		gitIgnoreFiles = string(gitIgnoreFilesRaw)
```

psiemens (commented):
Nice 👌

internal/accounts/create.go
sideninja: Not nitpicky at all, I agree 100% just overlooked it.

pkg/flowkit/util/utilities.go
```diff
@@ -154,3 +156,45 @@ func TestnetFaucetURL(publicKey string, sigAlgo crypto.SignatureAlgorithm) strin
...
+	return !info.IsDir()
+}
+
+type ReaderWriter interface {
+	ReadFile(source string) ([]byte, error)
+	WriteFile(filename string, data []byte, perm os.FileMode) error
+}
+
+// AddToGitIgnore adds a new line to the .gitignore if one doesn't exist it creates it.
+func AddToGitIgnore(
```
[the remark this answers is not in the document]
sideninja: At first, I was thinking no, and even commented that but then I was thinking more in terms of gain/loss, and I think that loss of not doing that is huge, imagine some cases where people will first init the flow and create accounts, start playing around and only then create a project on git (I know sometimes when I prototype I create git much later), the loss there would be their accounts exposed without maybe even realizing it. The gain of not doing it would be not creating an additional file, and although I agree that creating some files in the background is not great, we are actually not doing that since we are telling the user we will do that. I feel if we prevent one case of posting this account to git is worth it. I was actually thinking we will keep the encryption tbh and add the to a separate file.

internal/accounts/create.go
```diff
@@ -151,29 +149,28 @@ func create(
...
 	}
-	log.Info(fmt.Sprintf("\n This command will perform the following:"))
-	log.Info(fmt.Sprintf("- Generate a new ECDSA P-256 public and private key pair"))
-	if selectedNetwork != config.DefaultEmulatorNetwork() {
-		log.Info(fmt.Sprintf("- Save the private key to %s", privateJsonFileName))
+	log.Info(fmt.Sprintf("- Create a new account on %s paired with the public key.", output.Bold(networkName)))
+	log.Info(fmt.Sprintf("- Save the newly-created account to %s.\n", output.Bold("flow.json")))
+
+	if !output.WantToContinue() {
+		return nil, fmt.Errorf("process terminated")
```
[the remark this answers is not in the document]
sideninja: Yes, this is just a generic error. I mean we can return nil as an error and just cancel. I feel if a CLI command fails to do the job it should return an exit status. But it's a good question whether we should output that exit status, and the answer is probably no :D maybe I should remove that generally. As far as the output here, unfortunately, the error handler is generic to all commands and there are only two options, we output command was terminated and not display it as an error and return nil as an error (but then we don't return exit status) or we do it like this but just remove the exit status output.

sideninja:
@psiemens This is the updated output:
Screenshot 2022-07-21 at 13 06 51

With advance schema saved in the private files:
Screenshot 2022-07-21 at 13 07 13

pkg/flowkit/util/utilities.go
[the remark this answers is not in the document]
psiemens: That reasoning makes sense to me! 👍

psiemens (approved):
LGTM

psiemens (approved):

PR #581, opened 2022-07-21
make it possible to send in any emulator option 

### LINKED ISSUE #580: Gateway does not turn on transactionFees by default and it is impossible to turn them on
The way computation is computed is different on emulator since it is not turned on.

bjartek:
Closes #580

### DIFF
--- a/pkg/flowkit/gateway/emulator.go
+++ b/pkg/flowkit/gateway/emulator.go
@@ -103,7 +103,7 @@ func newEmulator(serviceAccount *flowkit.Account, emulatorLogger *zerolog.Logger
 			(*privKey).PublicKey(),
 			serviceAccount.Key().SigAlgo(),
 			serviceAccount.Key().HashAlgo(),
-		), emulator.WithLogger(*emulatorLogger))
+		), emulator.WithLogger(*emulatorLogger), emulator.WithTransactionFeesEnabled())
 	}
 
 	b, err := emulator.NewBlockchain(opts...)
--- a/pkg/flowkit/gateway/emulator.go
+++ b/pkg/flowkit/gateway/emulator.go
@@ -103,7 +103,7 @@ func newEmulator(serviceAccount *flowkit.Account, emulatorLogger *zerolog.Logger
 			(*privKey).PublicKey(),
 			serviceAccount.Key().SigAlgo(),
 			serviceAccount.Key().HashAlgo(),
-		), emulator.WithLogger(*emulatorLogger), emulator.WithTransactionFeesEnabled())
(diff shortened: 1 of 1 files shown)


bjartek:
I am a little bit confused. `flow emulator` does not use gateway, why is it there? What is the use case for it?

janezpodhostnik:
we have the `--transaction-fees` flag for this would you prefer this would be the default?

bluesign:
title should be `transaction fees` I guess. Should we also change the cli default for `transaction-fees`. I mean they have different start up options would be confusing.

sideninja:
> I am a little bit confused. flow emulator does not use gateway, why is it there? What is the use case for it?

Can you elaborate, please? I don't understand what you mean by `flow emulator` having a gateway.

> Should we also change the cli default for transaction-fees

Where is that?

sideninja:
Also @bjartek make sure the test work, but good improvement.

bjartek:
I changed my pr so that it possible to send in any emulator option so that we do not have to send pr if/when we want to change more.

bjartek:
@sideninja I just find the abstractions strange to reason about that is all. Flowkit has an gateway package that can start an emulator. but `flow emulator` does not use it.

sideninja:
> @sideninja I just find the abstractions strange to reason about that is all. Flowkit has an gateway package that can start an emulator. but `flow emulator` does not use it.

Ohhh I see yeah... Yeah I was thinking about changing that a while ago too. I even had a branch already with that feature but didn't release it. I paused doing that since I had questions about UX but I think I now could do a change. At that time you would have to pass the --host hosted or smth like that to indicate you want to run internal emulator. But I think now I have better idea how to do it. So I just might. Thanks for remembering me :D It would work similarly to how you use pm2, that one I always liked how you can then check logs, start/stop service etc. Or maybe how docker works with containers. The thing is we need to add more logic around it, it's not just switching to hosted version. Btw you can already use hosted emulator there, I just remembered. There is a --hosted flag

sideninja:
> I changed my pr so that it possible to send in any emulator option so that we do not have to send pr if/when we want to change more.

That's way better 👍

sideninja:
> It would also be nice to be able to use a gateway in combinated with dev-wallet. That is set up a server from flowkit. Is that possible? And also maybe connect to it using flow-cli

That unfortunately wouldn't be possible in the current form since dev-wallet is written in JS and expects to connect to a port, and when CLI runs the hosted emulator gateway there's no ports exposed. In any case I feel this convo should be moved to an issue as it doesn't relate to this PR.

sideninja (approved):

Issue #330, opened 2021-07-21
Test the commands layer 


sideninja:
### Issue To Be Solved
The command layer: https://github.com/onflow/flow-cli/tree/master/internal is lacking tests which has lead to multiple bugs with typos etc because it requires manual testing to detect that.

### Suggest A Solution
Write tests that execute functions in the command layer (functions were extracted with the purpose of testability).

Transaction example:
Test for send function here https://github.com/onflow/flow-cli/blob/master/internal/transactions/send.go#L55
Use mock `ReaderWriter` with afero mock loader to mock file loading like done here: https://github.com/onflow/flow-cli/blob/master/pkg/flowkit/config/loader_test.go#L88-L89
Test returned results in different formats.

neilZon:
@sideninja For comparing the console print outputs of commands for testing, I was thinking about using some 2 pointer approach that compares equivalence of non-format characters (ie. `\n`, `\t`, `<whitespace character>`). However this approach totally disregards formatting so I'm not too sure of how viable this is, however I have some ideas of how we can lightly regard formatting, but what are your thoughts on this approach?

sideninja:
I would say that once we have the service layer mocked out you can to an exact string match, since the result from the service layer will be predictable you can also exactly predict the output in different formats. If a formating changes (like adding a new line) then it's ok for the test to also break and require being updated.

PR #583, opened 2022-07-22
[WIP] get latest block test

neilZon:
Closes: #330 

## Description

This PR tests internal cmds 
______

For contributor use:

### DIFF
--- a/internal/blocks/blocks_test.go
+++ b/internal/blocks/blocks_test.go
@@ -2,34 +2,48 @@ package blocks
 
 import (
 	"testing"
+	"time"
 
 	"github.com/onflow/flow-cli/internal/command"
-	"github.com/onflow/flow-cli/pkg/flowkit/tests"
-	"github.com/stretchr/testify/assert"
-
-	"github.com/onflow/flow-cli/pkg/flowkit"
-	"github.com/onflow/flow-cli/pkg/flowkit/output"
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
-
-	"github.com/onflow/flow-go-sdk/crypto"
+	"github.com/onflow/flow-cli/pkg/flowkit/tests"
+	"github.com/onflow/flow-go-sdk"
+	"github.com/stretchr/testify/assert"
 )
 
-func setup() (*flowkit.State, *services.Services, *tests.TestGateway) {
-	readerWriter := tests.ReaderWriter()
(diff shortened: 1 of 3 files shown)


internal/blocks/blocks_test.go
```diff
@@ -0,0 +1,58 @@
...
+			Blocks: &MockBlocks{},
+		}
+
+		readerWriter := tests.ReaderWriter()
+		res, err := get([]string{"latest"}, readerWriter, command.Flags, s)
+		if err != nil {
+			panic(err)
+		}
+
+		expected := "Block ID\t6162000000000000000000000000000000000000000000000000000000000000\nParent ID\t6162000000000000000000000000000000000000000000000000000000000000\nTimestamp\t2022-07-27 14:27:49.585751 -0600 MDT m=+0.025544111\nHeight\t123\nTotal Seals\t0\nTotal Collections\t0"
```
sideninja, line 54: first, you should assert the values in the result struct. The String comparison is a bit fragile, so I would say it would be better to have some kind of regex maybe to detect values in there.

pkg/flowkit/services/blocks.go
```diff
@@ -50,6 +50,15 @@ func NewBlocks(
 	}
 }
 
+type IBlocks interface {
```
sideninja, line 53: Uff you from Java? :D 
I think it's a good idea to make an interface for services, I think it would make sense to even have a single interface in front of all services, I don't like how I designed this a while ago. It would however be a big breaking change so I think it should be as a separate PR that then allows this tests PR.

neilZon: > Uff you from Java? :D

Oh no is that a Javaism I'll come up with a better name 🤣
I'm having troubles thinking of a way to refactor the `Services` struct to be be defined by a single interface. Here is what I could think of https://github.com/onflow/flow-cli/compare/neil/refactor-services-for-testing

sideninja: I think I would rather refactor that the whole service layer has one interface and then methods like `GetTransaction`, `GetBlock` etc. I think we should first discuss this in an issue, so open an issue, write your proposal, I will leave comments and lets define things first there and then move on with implementation. The issue can be generic "refactor of services to use an interface" not just for testing, since this will improve other things as well.

sideninja:
Closing this in favour of defining first things in an issue.

PR #590, opened 2022-07-30
Optional arguments parsing fix

bluesign:
Closes #???

## Description

Fixes : 

When running a script with Optional String argument without quoting: 
`flow transactions send myfile.cdc someStringArg1 --network=emulator`

We get an error: `arg  is not expected type String?`

ref: https://discord.com/channels/613813861610684416/621847426201944074/1002734104141365250

PS: I had some problems with flowkit pkg and build process, so sending this PR little blind. 
PS2: There was some locationID problems with newer Cadence I guess, but didn't make sense to include in this PR

______

For contributor use:

- [] Added appropriate labels

### DIFF
--- a/pkg/flowkit/arguments.go
+++ b/pkg/flowkit/arguments.go
@@ -158,19 +158,25 @@ func ParseArgumentsWithoutType(fileName string, code []byte, args []string) (scr
 		astType := parameterList[index].TypeAnnotation.Type
 		semaType := checker.ConvertType(astType)
 
-		switch semaType {
-		case sema.StringType:
-			if len(argumentString) > 0 && !strings.HasPrefix(argumentString, "\"") {
-				argumentString = "\"" + argumentString + "\""
-			}
-		}
+		for {
+			switch v := semaType.(type) {
+			case *sema.OptionalType:
+				semaType = v.Type
+				continue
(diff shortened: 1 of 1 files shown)


pkg/flowkit/arguments.go
```diff
@@ -158,19 +158,25 @@ func ParseArgumentsWithoutType(fileName string, code []byte, args []string) (scr
 		astType := parameterList[index].TypeAnnotation.Type
 		semaType := checker.ConvertType(astType)
 
-		switch semaType {
-		case sema.StringType:
-			if len(argumentString) > 0 && !strings.HasPrefix(argumentString, "\"") {
-				argumentString = "\"" + argumentString + "\""
+		for {
```
sideninja, line 161: why is this for loop needed? it breaks on the first loop in any case

bluesign: Looping 2 times if type is Optional. I could have checked if optional type and unwrap the type before switch. I can change to that if it is more readable.

sideninja: ohh completely missed that `continue`. Makes sense it's ok that way.

sideninja (approved):

PR #606, opened 2022-08-05
Use Flow Account API for account creation

### LINKED ISSUE #597: Support creation of accounts on testnet and mainnet using APIs
We want to support the creation of testnet and mainnet accounts using APIs instead of manually going through the processes. 

Currently, you can create an account by running: `flow accounts create`, which will allow you to choose the network and open a faucet website where you have to fill in the captcha and copy back the keys. This can all be automated so the user can just run the command and the account is created using the APIs.

**Testnet**
We should completely remove the manual process for testnet and use the faucet API instead.

**Mainnet**
For mainnet case we should provide the user option to choose, either choose the current Flowport manual process or the automatic API option. API integration can be done according to this https://github.com/Outblock/flow-account-api

sukantoraymond:
Closes #597 

## Description

Replaces current implementation of manual account creation with the Flow Account API stated in https://github.com/Outblock/flow-account-api

______

For contributor use:


internal/accounts/create.go
```diff
@@ -213,32 +228,12 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
...
 
-		log.StartProgress("Waiting for your account to be created, please finish all the steps in the browser...\n")
-		_ = util.OpenBrowserWindow(link)
-		log.Info(output.Italic(fmt.Sprintf("You can also navigate to the link manually: %s\n", link)))
+		err = createNewAccount(selectedNetwork.Name, key.PublicKey().String())
+		if err != nil {
+			return nil, err
+		}
 
 		addr, err := getAccountCreatedAddressWithPubKey(service, key.PublicKey(), startHeight)
```
sukantoraymond, line 238: the API returns transaction ID as response, but I think we can just keep the current implementation of scanning through the events and match it with the public key

sideninja: We should track that transaction ID, it's more optimal and reliable.

internal/accounts/create.go
```diff
@@ -213,32 +228,12 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 
 		address = account.Address
 	} else {
-		var link string
-		switch selectedNetwork {
-		case config.DefaultTestnetNetwork():
```
sideninja, line 218: This shouldn't be removed. User should be able to choose a provider for account creation once they choose the network. In case of mainnet it should be blockport or lillico.

internal/accounts/create.go
```diff
@@ -286,6 +281,58 @@ func createInteractive(state *flowkit.State, loader flowkit.ReaderWriter) (*flow
 	return onChainAccount, nil
 }
 
+func createNewAccount(network, publicKey string) error {
```
sideninja, line 284: this should be extracted elswhere. I suggest creating a "gateway" module outside of flowkit in the internal module. In there we should move all the APIs implementation (mixpanel, sentry, this one). Then we just call those methods where needed.

sideninja (changes requested):
A user should be able to choose a provider. This is not meant to replace the current flow, but to add an option when choosing a provider. Left some other comments as well.

internal/accounts/create.go
```diff
@@ -36,6 +40,12 @@ import (
 	"github.com/onflow/flow-cli/pkg/flowkit/util"
 )
 
+const (
+	DefaultFlowAccountSigAlg  = "ECDSA_P256"
+	DefaultFlowAccountHashAlg = "SHA3_256"
```
sideninja, line 45: You can get default from flags

sukantoraymond: addressed

internal/accounts/create.go
sukantoraymond: addressed

sukantoraymond: @sideninja Since there is issue with getting transactions from testnet, let's just keep the current implementation of monitoring the events and get the account that matches the public key to get the new account. What do you think?

sideninja: I wouldn't do that. If there's an issue on testnet that will be fixed I'm sure, we shouldn't work around that. What is the issue by the way?

pkg/flowkit/services/transactions.go
```diff
@@ -73,13 +72,21 @@ func (t *Transactions) GetStatus(
 		t.logger.StartProgress("Waiting for transaction to be sealed...")
 	}
 
-	result, err := t.gateway.GetTransactionResult(tx, waitSeal)
-
+	result, err := t.gateway.GetTransactionResultByID(id, waitSeal)
```
sukantoraymond, line 75: Using GetTransactionResultByID instead so that we search for the transaction object directly by using the id provided through the argument. 

Current implementation using GetTransactionResult means that we are getting the transaction object from the given id in the argument and then using the transaction object to derive the id. In the case of lilico api, they do not result in the same value of ids

internal/accounts/create.go
```diff
@@ -326,6 +427,38 @@ func getAccountCreatedAddressWithPubKey(
...
 }
 
+// getAccountCreatedWithAPI gets the events from the transaction ID generated from the account creation API. It
+// extracts the newly created address from the account creation event payload.
+func getAccountCreatedWithAPI(
+	service *services.Services,
+	pubKey crypto.PublicKey,
+	txID string,
+) (*flow.Address, error) {
+	result, err := retry(txID, 10, time.Second, service.Transactions.GetTransactionResultByID)
```
sukantoraymond, line 437: Need retry because it takes time for the txID to show up on the chain. From my testing, I find that it takes around 3 retries with 2 seconds stop in between to find the transaction. Using 10 retries just to be safe.

pkg/flowkit/gateway/emulator.go
```diff
@@ -146,6 +147,20 @@ func (g *EmulatorGateway) GetTransactionResult(tx *flow.Transaction, waitSeal bo
...
+	result, err := g.backend.GetTransactionResult(g.ctx, txID)
+	if err != nil {
+		return nil, UnwrapStatusError(err)
+	}
+
+	if result.Status != flow.TransactionStatusSealed && waitSeal {
+		time.Sleep(time.Second)
+		return g.GetTransactionResultByID(txID, waitSeal)
+	}
+
```
sukantoraymond, line 161: I timed it and it takes 1 min + to get transaction since we need to wait for it to be sealed.

sideninja:
We shouldn't change APIs just to facilitate an issue with transaction iD. As soon as I finish some work on the playground I will help you solve the transaction ID issue.

sideninja:
@sukantoraymond I don't think we need both. In any way that's a different topic, we can discuss changing the API as part of the https://github.com/onflow/flow-cli/issues/493 but again I don't think we should be doing this now just to make the API work. We need to find the problem there. Until then this should wait.

PR #609, opened 2022-08-08
Fix bug for setting metric settings

sukantoraymond:
Currently running flow config metrics disable enables tracking. This PR fixes it by flipping from checking if command is  "disable" to if it is "enable"

______

For contributor use:


Makefile
```diff
@@ -56,14 +56,14 @@ install:
 	GO111MODULE=on go install \
 		-trimpath \
 		-ldflags \
-		"-X github.com/onflow/flow-cli/build.commit=$(COMMIT) -X github.com/onflow/flow-cli/build.semver=$(VERSION) -X github.com/onflow/flow-cli/pkg/flowkit/util.MIXPANEL_SERVICE_ACCOUNT_SECRET=$(MIXPANEL_SERVICE_ACCOUNT_SECRET) -X github.com/onflow/flow-cli/pkg/flowkit/util.MIXPANEL_PROJECT_TOKEN=$(MIXPANEL_PROJECT_TOKEN)" \
+		"-X github.com/onflow/flow-cli/build.commit=$(COMMIT) -X github.com/onflow/flow-cli/build.semver=$(VERSION) -X github.com/onflow/flow-cli/pkg/flowkit/util.MIXPANEL_SERVICE_ACCOUNT_SECRET=${MIXPANEL_SERVICE_ACCOUNT_SECRET} -X github.com/onflow/flow-cli/pkg/flowkit/util.MIXPANEL_PROJECT_TOKEN=${MIXPANEL_PROJECT_TOKEN}" \
```
sideninja, line 59: This is not the problem. The problem is that these variables weren't set as part of the environment running this make command. Please set them https://github.com/onflow/flow-cli/blob/3284a5b28330350bffbcdea1a2b4f6b84cf05c3e/.github/workflows/cd.yml#L30 and please test this as part of the PR.

sukantoraymond: I think we can run the test as part of CI, so we set the env variables with the secrets for make test, to make sure that they are there.

Makefile
```diff
@@ -106,3 +106,12 @@ check-headers:
...
 	cd pkg/flowkit; go mod tidy
+
+.PHONY: check-env
+check-env:
+ifndef MIXPANEL_SERVICE_ACCOUNT_SECRET
+	$(error MIXPANEL_SERVICE_ACCOUNT_SECRET is undefined)
+endif
+ifndef MIXPANEL_PROJECT_TOKEN
+	$(error MIXPANEL_PROJECT_TOKEN is undefined)
+endif
```
sukantoraymond, line 117: Test checks that MIXPANEL_SERVICE_ACCOUNT_SECRET and MIXPANEL_PROJECT_TOKEN will always be passed in. otherewise, ci will fail.

sideninja: This test is not necessary, since if this works then it will always work unless tokens are removed, which won't just happen by accident. The test that is necessary is making sure passing env variables work, so I would suggest adding a temporary action in the CI that checks if those variables are present in the Makefile, you can add an action that checks if values are present in makefile and then run that action as part of the CI, after the change validate that action has been run successfully on the CI and then you can remove it.

sideninja: I see. That's good test then. I would say you can remove this altogether from the CI since CI doesn't need to test this everytime. But good that we know it works this way.

pkg/flowkit/util/mixpanel.go
```diff
@@ -31,7 +31,7 @@ import (
 
 const (
 	MIXPANEL_TRACK_URL   = "https://api.mixpanel.com/track"
-	MIXPANEL_QUERY_URL   = "https://mixpanel.com/api/2.0/engage?project_id=2154593"
+	MIXPANEL_QUERY_URL   = "https://mixpanel.com/api/2.0/engage?project_id=2123215"
```
sideninja, line 34: how come this changed now? don't we use the same token?

sukantoraymond: We change to use flow production project in mixpanel

sideninja: Ok I see.

sideninja (approved):
LGTM. Good thinking for testing.

PR #621, opened 2022-08-15
Use config json in home config directory to check metrics opt in status

### LINKED ISSUE #615: Refactor metric preference setting
The CLI is getting rate-limited when requesting the user preference from the mixpanel. We need to refactor the way this works so we don't send all those requests. We could keep opt-in behavior and only disable sending of requests if the value is set in a configuration file we add. The configuration file would be added once a user chooses to opt out and we should also let the user know we will create that file for them and that if they switch the operating system they need to again disable this setting. That file should be saved somewhere in a users home directory (`~/.config`), the file should be named something like `.flow-cli.config` and it should contain JSON with values (currently only the metrics: "disabled/enabled"). We should completely remove the part where we check for those values from mixpanel.

sukantoraymond:
Closes #615 

## Description

This PR uses config.json stored in user's local machine home config directory to check if user is opted in for metrics tracking

______

For contributor use:


pkg/flowkit/util/utilities.go
```diff
@@ -187,3 +190,57 @@ func AddToGitIgnore(filename string, loader ReaderWriter) error {
...
+	Metrics bool `json:"metrics"`
+}
+
+// AddToConfig adds a new line to user's config dir stating the user's preference for command tracking
+func AddToConfig(loader ReaderWriter, metricsEnabled bool) error {
+	homeConfigDir, err := os.UserConfigDir()
+	if err != nil {
+		return err
+	}
+	cliDir := filepath.Join(homeConfigDir, FLOW_CLI)
```
sukantoraymond, line 204: we use filepath.join because it will automatically append \ for windows and / for other os.

pkg/flowkit/util/utilities.go
```diff
@@ -187,3 +190,57 @@ func AddToGitIgnore(filename string, loader ReaderWriter) error {
...
 	)
 }
+
+type jsonConfig struct {
+	Metrics bool `json:"metrics"`
+}
+
+// AddToConfig adds a new line to user's config dir stating the user's preference for command tracking
+func AddToConfig(loader ReaderWriter, metricsEnabled bool) error {
+	homeConfigDir, err := os.UserConfigDir()
```
sukantoraymond, line 200: using user config dir, in my machine it is $HOME/Library/Application Support, so that it works automatically for windows and other os

sukantoraymond:
Sample output of json in config.json 

````
{
 "metrics": true
}
````

**Comment by sukantoraymond** [comment]
I also don't think we need go routine for sending command usage, because we are not aggregating the command usage. There is no limit on sending API events on mixpanel. Current implementation is fine.

File: pkg/flowkit/util/utilities.go [hunk 3]
```diff
@@ -187,3 +190,57 @@ func AddToGitIgnore(filename string, loader ReaderWriter) error {
...
+	var jsonConf jsonConfig
+	homeConfigDir, err := os.UserConfigDir()
+	if err != nil {
+		return false, err
+	}
+	cliDir := filepath.Join(homeConfigDir, FLOW_CLI)
+	cliConfigPath := filepath.Join(cliDir, FlowCLIConfigFile)
+
+	_, err = os.Stat(cliConfigPath)
+	if os.IsNotExist(err) { // if config.json exists, return metrics enabled by default
```
sukantoraymond, line 235: if user has never opted out, tracking enabled by default.

internal/config/metrics-settings.go
```diff
@@ -47,17 +46,18 @@ var MetricsSettings = &command.Command{
...
 		return nil, err
 	}
 
-	return &Result{
-		fmt.Sprintf("Metrics have been %sd", args[0]),
-	}, nil
+	output := fmt.Sprintf("Metrics tracking is %sd and its settings is stored in %s \n", args[0], configPath) +
+		"Please note that you will also need to opt out of metrics tracking on other devices that you also use flow-cli on \n"
+
+	return &Result{output}, nil
```
sukantoraymond, line 62: output stating where the tracking preferences is stored

internal/config/metrics-settings.go
```diff
@@ -47,17 +47,18 @@ var MetricsSettings = &command.Command{
...
+	configPath, err := util.AddToConfig(loader, enabled)
 	if err != nil {
 		return nil, err
 	}
 
-	return &Result{
-		fmt.Sprintf("Metrics have been %sd", args[0]),
-	}, nil
+	output := fmt.Sprintf("Metrics tracking is %sd and its settings is stored in %s \n", args[0], configPath) +
+		"Please note that you will also need to opt out of metrics tracking on other devices that you also use flow-cli on \n"
```
sideninja, line 61:
```suggestion
	output := fmt.Sprintf("Command usage tracking is %sd. The preference was saved in a file located at %s \n", args[0], configPath) +
		"Please be aware that your command tracking preference will only be set for your current user. \n"
```

internal/config/metrics-settings.go
```diff
@@ -47,17 +47,18 @@ var MetricsSettings = &command.Command{
...
+	configPath, err := util.AddToConfig(loader, enabled)
 	if err != nil {
 		return nil, err
 	}
 
-	return &Result{
-		fmt.Sprintf("Metrics have been %sd", args[0]),
-	}, nil
+	output := fmt.Sprintf("Command usage tracking is %sd. The preference was saved in a file located at %s \n", args[0], configPath) +
+		"Please be aware that your command tracking preference will only be set for your current user. \n"
```
sideninja, line 61:
```suggestion
		"Please be aware that your command tracking preference will only be set for your current user."
```

sideninja (approved):

PR #623, opened 2022-08-17
Resolve imports when deploying a contract

sideninja:
This PR makes it possible to resolve any imports defined in a contract during the deployment of the contract on account.

______

For contributor use:

### DIFF
--- a/pkg/flowkit/services/accounts.go
+++ b/pkg/flowkit/services/accounts.go
@@ -316,7 +316,9 @@ func (a *Accounts) AddContract(
 		return nil, err
 	}
 
-	err = contract.validate(resolver.HasFileImports())
+        hasFileImports = resolver.HasFileImports()
+        
+	err = contract.validate(hasFileImports)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/flowkit/services/accounts.go
+++ b/pkg/flowkit/services/accounts.go
@@ -316,14 +316,13 @@ func (a *Accounts) AddContract(
 		return nil, err
 	}
 
-        hasFileImports = resolver.HasFileImports()
-        
+	hasFileImports := resolver.HasFileImports()
 	err = contract.validate(hasFileImports)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


pkg/flowkit/services/accounts.go
```diff
@@ -281,9 +283,39 @@ func (a *Accounts) AddContract(
 	account *flowkit.Account,
 	contractName string,
 	contractSource []byte,
-	updateExisting bool,
 	contractArgs []cadence.Value,
+	contractFilename string,
```
psiemens, line 287: Would it make sense for these 4 arguments to be a `Contract` struct?

The function signature is a bit large now -- this could make it easier to call IMO.

pkg/flowkit/services/accounts.go
```diff
@@ -276,19 +278,70 @@ func (a *Accounts) Create(
...
+	contract *Contract,
 	updateExisting bool,
-	contractArgs []cadence.Value,
 ) (*flow.Account, error) {
+	resolver, err := contracts.NewResolver(contract.Source)
+	if err != nil {
+		return nil, err
+	}
+
+	err = contract.validate(resolver.HasFileImports())
```
psiemens, line 319:
```suggestion
        hasFileImports = resolver.HasFileImports()
        
	err = contract.validate(hasFileImports)
```

I can't remember what the cost is of calling `HasFileImports()` , but might make sense to save to a variable here.

psiemens (approved):

pkg/flowkit/services/accounts.go
```diff
@@ -276,19 +278,72 @@ func (a *Accounts) Create(
...
+	}
+
+        hasFileImports = resolver.HasFileImports()
+        
+	err = contract.validate(hasFileImports)
+	if err != nil {
+		return nil, err
+	}
+
+	if resolver.HasFileImports() {
```
psiemens, line 326: Can use `hasFileImports` here too
```suggestion
	if hasFileImports {
```

PR #627, opened 2022-08-19
Gpavlov/fix dev wallet refs

### LINKED ISSUE #620: fcl-dev-wallet folder structure changes break flow-cli build
In the current master of the [fcl-dev-wallet](https://github.com/onflow/fcl-dev-wallet), the go files were moved into internal folder go/wallet but the import in [wallet.go](https://github.com/onflow/flow-cli/blob/master/internal/tools/wallet.go) was not updated which leads to an error during build.
The breaking update to fcl-dev-wallet was not released yet so build doesn't fail while using the latest official version 0.39 . When new version is created off of master it will break flow-cli builds

gpavlov2016:
Closes #620 

## Description

Updates the imports and usage of fcl-dev-wallet to be consistent with the latest changes in fcl-dev-wallet

______

For contributor use:


.github/workflows/detect-renamed-docs.yml
```diff
@@ -0,0 +1,34 @@
+name: 🚀 Deploy
```
sideninja, line 1: I think this should be removed, I believed it escaped from the other PR you made. I left some comments on that, so deleting this you will avoid potential conflicts.

internal/tools/wallet.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/onflow/flow-cli/pkg/flowkit/output"
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
 
-	devWallet "github.com/onflow/fcl-dev-wallet"
+	devWallet "github.com/onflow/fcl-dev-wallet/go/wallet"
```
sideninja, line 30: you need to use `go get` command to update the dev-wallet dependency so this linter doesn't complain and there are go.sum entries.

sideninja: We should also release a tagged version of dev-wallet to do this update.

.github/workflows/detect-renamed-docs.yml
gpavlov2016: Absolutely, it shouldn't be here. Thanks for noticing

internal/tools/wallet.go
gpavlov2016: The module import in go.mod didn't change, it's still github.com/onflow/fcl-dev-wallet
I think the linter complains because it uses a tagged version of dev wallet that doesn't have the corresponding changes yet. 
I am coordinating the release of dev wallet so once it's done this error should not happen

internal/tools/wallet.go
```diff
@@ -63,21 +63,11 @@ func wallet(
...
-		AccountKeyID:          "0",
-		BaseURL:               "http://localhost:8701",
-		ContractFungibleToken: "0xee82856bf20e2aa6",
-		ContractFlowToken:     "0x0ae53cb6e3f42a79",
-		ContractFUSD:          "0xf8d6e0586b0a20c7",
-		ContractFCLCrypto:     "0xf8d6e0586b0a20c7",
-		AvatarURL:             "https://avatars.onflow.org/avatar/",
-		FlowInitAccountsNo:    "0",
-		TokenAmountFLOW:       "100.0",
-		TokenAmountFUSD:       "100.0",
```
psiemens, line 80: Are these config variables no longer needed?

gpavlov2016: They are loaded dynamically by the fcl-dev-wallet

psiemens: Hmm, loaded dynamically from where? Does the dev wallet define defaults for these values?

I ask only because we previously had issues with missing config variables and these lines (I believe) were added as a fix. So want to make sure this doesn't cause a regression.

sideninja: @psiemens I made a PR some time back addressing this configurations, the problem was that everytime fcl team added a new required variable things broke since they had to add that config variable in multiple places (js app, go module in dev-wallet, cli), so I made it so there's only one place acting as a source to variables that are not CLI specific. The process is described here https://github.com/onflow/fcl-dev-wallet/tree/main/go

sideninja (approved):

PR #637, opened 2022-08-23
use custom emulator

### LINKED ISSUE #520: Flow emulator cannot be started with custom emulator
### Problem

When emulator name is changed from "default" to something else, we get an error message saying “no default emulator account”. This means that currently flow cli users can’t start emulator with their custom emulator. 

### Steps to Reproduce

- change the emulator name from “default” to something else 
- start emulator

### Acceptance Criteria

Users should be able to start emulator with non default emulator.

sukantoraymond:
Closes #520

This PR enables user to start emulator with non default emulator in flow.json

______

For contributor use:

### DIFF
--- a/pkg/flowkit/state.go
+++ b/pkg/flowkit/state.go
@@ -173,11 +173,13 @@ func (p *State) SetAccountFileLocation(account Account, location string) {
 
 // EmulatorServiceAccount returns the service account for the default emulator profile.
 func (p *State) EmulatorServiceAccount() (*Account, error) {
-	emulator := p.conf.Emulators.Default()
-	if emulator == nil {
-		return nil, fmt.Errorf("no default emulator account")
-	}
+	emulator := config.Emulator{}
 
+	if len(p.conf.Emulators) > 0 && p.conf.Emulators[0].Name != config.DefaultEmulatorConfigName {
+		emulator = p.conf.Emulators[0]
+	} else {
+		emulator = *p.conf.Emulators.Default()
+	}
 	return p.accounts.ByName(emulator.ServiceAccount)
 }
 


sukantoraymond:
As I am working on this PR, I'm just thinking more about the reasons for a user to create a custom emulator in flow.json. If it's to use a different port number, current implementation of using different port number in flow.json won't work since currently it is using the default value in start.go in emulator repo. We can just pass the custom port number using flag instead of using flow.json. 

If the user wants to change the emulator service account, this means that the user will have to create a new emulator account. Currently from emulator repo code, seems that the address used in emulator is default, so changes in flow.json won't even be reflected unless we make changes to emulator repo code to use the custom address value in flow.json for emulator. 

My concluding thought I guess is if we even want to enable users to use custom emulator? @sideninja  

Screen Shot 2022-08-23 at 6 30 46 PM

sukantoraymond:
What this PR does is basically just change the name of the default emulator. If we want to use custom values like port, address on emulator, we would need to make changes on the emulator repo.

pkg/flowkit/state.go
```diff
@@ -173,11 +173,13 @@ func (p *State) SetAccountFileLocation(account Account, location string) {
 
 // EmulatorServiceAccount returns the service account for the default emulator profile.
 func (p *State) EmulatorServiceAccount() (*Account, error) {
-	emulator := p.conf.Emulators.Default()
-	if emulator == nil {
-		return nil, fmt.Errorf("no default emulator account")
-	}
+	emulator := config.Emulator{}
```
sideninja, line 176:
```suggestion
	 var emulator config.Emulator
```

pkg/flowkit/state.go
```diff
@@ -173,11 +173,13 @@ func (p *State) SetAccountFileLocation(account Account, location string) {
...
 
+	if len(p.conf.Emulators) > 0 && p.conf.Emulators[0].Name != config.DefaultEmulatorConfigName {
```
sideninja, line 178: Why do we need to change this? Not sure how would the first emulator be important in case there are multiple defined?

sideninja:
We could set the private key using the configuration, but I agree. I think we should remove tool-specific configurations from flow.json.

sukantoraymond:
I think we can just add a message saying "default emulator is required" is we detect a custom emulator in flow.json

sideninja:
I think we can close both for now.

PR #640, opened 2022-08-24
Integrate Cadence testing framework to the CLI

SupunS:
Work towards https://github.com/onflow/cadence/issues/331

## Description

This PR is to integrate the new test-framework into the CLI.
Command:
```
flow tests run /path/to/test_script.cdc
```

Sample output:
```
Running tests...

Test results:
- PASS: testImportingContract
- PASS: testSimpleScript
- PASS: testSimpleTransaction
- FAIL: testAssertionFailure
		assertion failed: manually failed the test case
- PASS: testInvokingContractFunction
- PASS: testReadingFile
- FAIL: testDeployingContractWithImports
		panic: [Error Code: 1101] cadence runtime error Execution failed:
			error: cannot deploy invalid contract
			 --> 9df24568a2615ba869f4b00abde9fb17f9b1ebe1d3cdd45d3f191f7d579e415e:4:7
```
______

For contributor use:


go.mod
```diff
@@ -175,3 +179,5 @@ require (
 )
 
 replace github.com/onflow/flow-cli/pkg/flowkit => ./pkg/flowkit
+
+replace github.com/onflow/cadence/test-framework => ../cadence/test-framework
```
SupunS, line 183: This is temporary. Will be removing this once the cadence tag is created.

internal/tests/tests.go
```diff
@@ -0,0 +1,75 @@
...
+
+	_, _ = fmt.Fprintf(writer, test_framework.PrettyPrintResults(r.Results))
+
+	_ = writer.Flush()
+
+	return b.String()
+}
+
+func (r *TestResult) Oneliner() string {
+	return test_framework.PrettyPrintResults(r.Results)
```
SupunS, line 74: It is not possible to get the test results in one-line, so I just used the default pretty-printing which is multi-line. Could that be a problem?
Any alternatives we could do?

sideninja: This is not so important in this case, but you could maybe just print out one liner like so:
`Test results: PASS: 5, FAIL: 2` 
But even if you just default to the multi-line output is ok.

internal/tests/execute.go
```diff
@@ -0,0 +1,75 @@
...
+	// Nothing for now
+}
+
+var testFlags = flagsTests{}
+
+var ExecuteCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "run <filename>",
+		Short:   "Run a test script",
+		Example: `flow tests run script.cdc`,
```
sideninja, line 41: I feel `flow test run` would be better name (following others). I also wonder if we could just have it `flow test` without the `run`? Do you think there is another test subcommand that will come in handy in the future? I feel that we could add features like coverage report, benchmarking etc all as flags.

internal/tests/execute.go
```diff
@@ -0,0 +1,75 @@
...
+type flagsTests struct {
+	// Nothing for now
+}
+
+var testFlags = flagsTests{}
+
+var ExecuteCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "run <filename>",
+		Short:   "Run a test script",
```
sideninja, line 40: Maybe more like: "Run Cadence tests"? anyway we can say they are cadence tests not some other type

pkg/flowkit/services/tests.go
```diff
@@ -0,0 +1,79 @@
...
+
+			return string(content), nil
+		})
+
+	t.logger.Info("Running tests...")
+
+	return runner.RunTests(string(code))
+}
+
+func absolutePath(basePath, filePath string) string {
```
sideninja, line 73: maybe move this to utils

pkg/flowkit/services/tests.go
```diff
@@ -0,0 +1,79 @@
...
+		logger: logger,
+	}
+}
+
+// Execute test scripts.
+//
+func (t *Tests) Execute(code []byte, scriptPath string, readerWriter flowkit.ReaderWriter) (test_framework.Results, error) {
+	runner := test_framework.NewTestRunner().
+		WithImportResolver(func(location common.Location) (string, error) {
+			stringLocation, isFileImport := location.(common.StringLocation)
```
sideninja, line 53: address imports are not possible?

sideninja:
@SupunS this is really awesome!!! 🔥 Thank you. Left some comments but nothing major. The one thing that is also lacking is documentation. Make sure to add that as you finish things up.

internal/tests/execute.go
SupunS: Actually, I also wanted to go with just `flow test`, but wasn't sure / didn't know if it's possible to drop the second portion of the command. If we can, that would be great.

pkg/flowkit/services/tests.go
SupunS: Address imports are handled at the cadence/test-framework side. They won't reach here.
This resolver is only used for resolving local file imports.

internal/tests/execute.go
SupunS: Added with f8dbe2e41249002d54d7fcf68ba1397d1ba8ba2b

SupunS:
@sideninja I'll open a separate PR to add documentation, so that this PR doesn't get too big.

sideninja:
The only question left here is, should we have this as top commands or not? why this changed from last time you were working on it is because I believe we will also add linter, so I'm now wondering should those commands be `flow test` `flow lint` or `flow cadence test` `flow cadence lint`. In some way lint and test are indicating doing that for flow so in some sense that is justifiable to be as a top-level command. Do you have opinion?

SupunS:
I'm happy with either way. 

I'm slightly leaning towards having the `test` command at the top-level, because it's not only a cadence-related command, but also involves other aspects of flow (uses flow.json, test cases may actually test all blockchain-related operations, etc.). But for some of the other tools like `lint`, and maybe document-generator (in future), grouping them under `cadence` makes sense.

sideninja (approved):
Looks good. Thank you! Nice work

PR #645, opened 2022-08-30
Warning when deploying standard contract to mainnet accidentally

### LINKED ISSUE #599: Warning on Standard Contract deploy on mainnet
### Issue To Be Solved

Many new comers accidentally deploy FungibleToken, NonFungibleToken, NFTStoreFront contracts to their accounts, either because of a mistake or totally because they are uninformed.

### Suggest A Solution

We can show a warning and prompt for confirmation in a case like this

bluesign:
Closes #599

______

For contributor use:

### DIFF
--- a/pkg/flowkit/services/project.go
+++ b/pkg/flowkit/services/project.go
@@ -102,6 +102,12 @@ func (p *Project) Deploy(network string, update bool) ([]*contracts.Contract, er
 		return nil, config.ErrDoesNotExist
 	}
 
+	// check for accidental standard contract deployments
+	err := p.state.CheckForStandardContractUsageOnMainnet()
+	if err != nil {
+		return nil, err
+	}
+
 	// check there are not multiple accounts with same contract
 	if p.state.ContractConflictExists(network) {
 		return nil, fmt.Errorf( // TODO(sideninja) specify which contract by name is a problem
--- a/pkg/flowkit/state.go
+++ b/pkg/flowkit/state.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path"
 
+	"github.com/manifoldco/promptui"
 	"github.com/onflow/cadence"
 	"github.com/onflow/flow-go-sdk"
 	"github.com/onflow/flow-go-sdk/crypto"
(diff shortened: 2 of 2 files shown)


pkg/flowkit/state.go
```diff
@@ -117,6 +118,111 @@ func (p *State) Save(path string) error {
...
+	if err != nil {
+		return err
+	}
+	for _, contract := range contracts {
+		standardContract, ok := standardContracts[contract.Name]
+		if ok {
+
+			fmt.Printf("It seems like you are trying to deploy %s to Mainnet \n", contract.Name)
+			fmt.Printf("It is a standard contract already deployed at address 0x%s \n", standardContract.Address.String())
+			fmt.Printf("You can read more about it here: %s \n", standardContract.InfoLink)
```
sideninja, line 187: any output should be done using the logger.

sideninja (changes requested):
I think this should be refactored a bit. What I suggest is to have a function that checks whether the deployment is of a standard contract if that is true then it prompts the user if they want to just use the standard contract already deployed and if they decide so then there's another function on the state that replaces that contract, but all that should be called from the command layer. This will avoid polluting state with prompts and also make the API generic.

bluesign:
> I think this should be refactored a bit. What I suggest is to have a function that checks whether the deployment is of a standard contract if that is true then it prompts the user if they want to just use the standard contract already deployed and if they decide so then there's another function on the state that replaces that contract, but all that should be called from the command layer. This will avoid polluting state with prompts and also make the API generic.

Actually that was my first plan https://github.com/onflow/flow-cli/issues/599#issuecomment-1222187738

> As we are not exposing direct access to state, should I add some helpers to Project, such as "ContractsToDeploy() and ReplaceContractWithAlias` ? or is there a better place for this?

As state is not exposed to command layer, I need to add something like: ContractsToDeploy and ReplaceContractWithAlias methods somewhere. With those I can actually move prompt to the Command layer too.

bluesign:
Sure I will adjust today

pkg/flowkit/services/project.go
```diff
@@ -92,6 +95,131 @@ func (p *Project) Init(
...
+		if err == promptui.ErrInterrupt {
+			os.Exit(-1)
+		}
+
+		if useMainnetVersion == "Yes" {
+			err := p.ReplaceStandardContractReferenceToAlias(standardContract)
+			if err != nil {
+				return err
+			}
+		}
```
sideninja, line 218: could we extract this to just return to a function in prompts.go and just return bool.

internal/project/deploy.go
```diff
@@ -50,6 +50,16 @@ func deploy(
 	services *services.Services,
 	_ *flowkit.State,
 ) (command.Result, error) {
+
+	//precheck for standard contract on Mainnet
+	if globalFlags.Network == "mainnet" {
```
sideninja, line 55: use the constant instead of string literal

sideninja (approved):

PR #677, opened 2022-10-17
Transaction Send multisign support 

### LINKED ISSUE #667: Multiple signers don't work in 0.40.1
### Problem

Multiple signers are still not supported in 0.40.1 (installed via homebrew)
Additionally, flow-cli reports that there is a newer version (v0.39.3) whereas "flow version" reports "Version: v0.40.1"

### Steps to Reproduce

flow transactions send ./transactions/test.cdc --signer account1 --signer account2 --network mainnet

also tried this format since documentation for multiple signers is not updated yet:

flow transactions send ./transactions/test.cdc --signer account1 account2 --network mainnet

### Expected

Transaction to be signed by two authorizers

### Actual output

❗   Version warning: a new version of Flow CLI is available (v0.39.3).
   Read the installation guide for upgrade instructions: https://docs.onflow.org/flow-cli/install

❌ Command Error: provided authorizers length mismatch, required authorizers 2, but provided 1

### More info

BTW when I downloaded sources for 0.40.1 (https://github.com/onflow/flow-cli/archive/refs/tags/v0.40.1.tar.gz) version.txt inside says "v0.39.3". It's the same for version 0.40.0

bluesign:
Closes #667 

## Description

Added with some assumptions as follows:

- the first signer is the proposer 
- the last signer is the payer 
- authorizer count is taken from the transaction code 

this should also allow different weight keys to be used with multisign. So account can have more than one key.

______

For contributor use:

### DIFF
--- a/docs/send-transactions.md
+++ b/docs/send-transactions.md
@@ -131,7 +131,7 @@ Specify the name of the account that will be used as payer in the transaction.
 ### Authorizer
 
 - Flag: `--authorizer`
-- Valid inputs: the name of an account defined in the configuration (`flow.json`)
+- Valid inputs: the name of a single or multiple comma-separated accounts defined in the configuration (`flow.json`)
 
 Specify the name of the account(s) that will be used as authorizer(s) in the transaction. If you want to provide multiple authorizers separate them using commas (e.g. `alice,bob`)
 
--- a/docs/send-transactions.md
+++ b/docs/send-transactions.md
@@ -133,7 +133,7 @@ Specify the name of the account that will be used as payer in the transaction.
 - Flag: `--authorizer`
 - Valid inputs: the name of an account defined in the configuration (`flow.json`)
 
-Specify the name of the account(s) that will be used as authorizer(s) in the transaction.
+Specify the name of the account(s) that will be used as authorizer(s) in the transaction. If you want to provide multiple authorizers separate them using commas (e.g. `alice,bob`)
 
 ### Arguments JSON
 
--- a/docs/send-transactions.md
+++ b/docs/send-transactions.md
@@ -114,6 +114,27 @@ Specify fields to exclude from the result output. Applies only to the text outpu
 
 Specify the name of the account that will be used to sign the transaction.
 
+### Proposer
+
+- Flag: `--proposer`
+- Valid inputs: the name of an account defined in the configuration (`flow.json`)
+
+Specify the name of the account that will be used as proposer in the transaction.
+
+### Payer
+
+- Flag: `--payer`
+- Valid inputs: the name of an account defined in the configuration (`flow.json`)
+
+Specify the name of the account that will be used as payer in the transaction.
+
+### Authorizer
+
+- Flag: `--authorizer`
+- Valid inputs: the name of an account defined in the configuration (`flow.json`)
+
+Specify the name of the account(s) that will be used as authorizer(s) in the transaction.
+
 ### Arguments JSON
 
 - Flag: `--args-json`
--- a/internal/transactions/send.go
+++ b/internal/transactions/send.go
@@ -21,15 +21,14 @@ package transactions
 import (
 	"encoding/hex"
 	"fmt"
-	"github.com/onflow/flow-cli/pkg/flowkit/config"
-
-	"github.com/spf13/cobra"
-
-	"github.com/onflow/cadence"
 
 	"github.com/onflow/flow-cli/internal/command"
 	"github.com/onflow/flow-cli/pkg/flowkit"
+	"github.com/onflow/flow-cli/pkg/flowkit/config"
 	"github.com/onflow/flow-cli/pkg/flowkit/services"
+
+	"github.com/onflow/cadence"
+	"github.com/spf13/cobra"
 )
 
 type flagsSend struct {
--- a/internal/transactions/send.go
+++ b/internal/transactions/send.go
@@ -34,9 +34,9 @@ type flagsSend struct {
 	ArgsJSON  string   `default:"" flag:"args-json" info:"arguments in JSON-Cadence format"`
 	Arg       []string `default:"" flag:"arg" info:"⚠️  Deprecated: use command arguments"`
 	Signer    string   `default:"" flag:"signer" info:"Account name from configuration used to sign the transaction as proposer, payer and suthorizer"`
-	Proposer  string   `default:"" flag:"signer" info:"Account name from configuration used as proposer"`
-	Payer     string   `default:"" flag:"signer" info:"Account name from configuration used as payer"`
-	Autorizer []string `default:"" flag:"signer" info:"Account name(s) from configuration used as authorizer(s)"`
+	Proposer  string   `default:"" flag:"proposer" info:"Account name from configuration used as proposer"`
+	Payer     string   `default:"" flag:"payer" info:"Account name from configuration used as payer"`
+	Autorizer []string `default:"" flag:"authorizer" info:"Account name(s) from configuration used as authorizer(s)"`
 	Include   []string `default:"" flag:"include" info:"Fields to include in the output"`
 	Exclude   []string `default:"" flag:"exclude" info:"Fields to exclude from the output (events)"`
 }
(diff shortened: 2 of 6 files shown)


internal/transactions/send.go
```diff
@@ -63,20 +64,52 @@ func send(
...
+	}
+
+	//remove extra signers
+	authorizers = authorizers[:authorizerCount]
+
+	//proposer is the first signer
+	proposer := signers[0]
+
+	//payer signs last
+	payer := signers[len(signers)-1]
```
sideninja, line 111: I feel this should be encapsulated in the transaction model.

sideninja: in here https://github.com/onflow/flow-cli/blob/master/pkg/flowkit/transaction.go#L289

bluesign: @sideninja How about if we add something like `NewTransactionFromPayloadWithSigners(payload, signers)`?

I am worried about some confusion for flowkit users as a method. ( as it sets, payer, proposer and authorizers only from signers )

sideninja: Oh sorry I meant just the first part of validation of authorizer count, the second part of setting those values can stay there or be extracted in that method. I feel those methods will have to be refactored to build pattern soon probably anyway.

pkg/flowkit/services/transactions.go
```diff
@@ -80,6 +77,56 @@ func (t *Transactions) GetStatus(
...
+	for _, auth := range signers {
+		if _, ok := seen[auth.Address()]; ok {
+			continue
+		}
+		authorizers = append(authorizers, auth.Address())
+	}
+
+	if len(authorizers) < authorizerCount {
+		return nil, fmt.Errorf("invalid number of authorizers, expected: %d", authorizerCount)
+	}
```
sideninja, line 105: wouldn't that be handled by https://github.com/bluesign/flow-cli/blob/c820e7b9125d75b12e85b28afbd9b796461d7a09/pkg/flowkit/transaction.go#L317

pkg/flowkit/services/transactions.go
```diff
@@ -80,6 +77,56 @@ func (t *Transactions) GetStatus(
 	return tx, result, err
 }
 
+// BuildWithSigners builds a transaction with a set of signers.
+func (t *Transactions) BuildWithSigners(
```
sideninja, line 81: not sure if I like this extra method on the API. It starts to make API grow for very specific cases.

pkg/flowkit/services/transactions.go
bluesign: yeah, I will move the check to the slice one line below.

pkg/flowkit/services/transactions.go
bluesign: This part is a bit tricky, because as we are getting a lot of data with assumption from signers ( last signer is payer, first one is proposer , first N ( N authorizer count in code ) is authorizers. It is hard to put them as a method of `transaction`. 

For sure it has to be either a NewTransaction constructor or some builder.

Another option is to break this function as utility functions. But the utility function is a bit complicated too, for example one of them needs to be something like 'GetUniqueAddressesFromAccounts`, but it feels so useless and not enough utility.

sideninja: I'm not sure I like this implicitness of the order. I feel we could in case you want to have multiple roles signers just support usage of the same flags as in build (where multiple roles are allowed) to keep the consistency of flag purpose. So I feel if you want to use multiple signers you have to specify them using the flags like `--proposer alice --authorizer alice --payer bob`. What do you think? And because it's a send command that then replaces the single signer where all the roles are done by one account. I feel that also makes the implementation clearer. We can also just use the build then as it's already used that way.

sideninja: Also @bluesign now that I think more about it I feel the `transactions.Send` function itself should be refactored to support passing proposer, payer, authorizers. By default that will all just be same but still.

bluesign: > Also @bluesign now that I think more about it I feel the `transactions.Send` function itself should be refactored to support passing proposer, payer, authorizers. By default that will all just be same but still.

yeah, I think this would be the best solution. Let me shape it up a bit and we discuss.

bluesign: @sideninja one question here: if` --payer bob` is specified but `proposer` is not, should we assume proposer from payer ? I am thinking how we can move to that state from having single `--signer` with less problems

sideninja: I would keep signer for only one that has all roles as ti was before, but if you want to use different roles then you must specify all.

bluesign:
Yeah I was waiting on that, I think it will be much cleaner

bluesign:
sorry this waited too long, it should be ready now

pkg/flowkit/arguments.go
```diff
@@ -65,6 +66,79 @@ func ParseArgumentsJSON(input string) ([]cadence.Value, error) {
 	return cadenceArgs, nil
 }
 
+func ParseArgumentsCommaSplit(input []string) ([]cadence.Value, error) {
```
sideninja, line 69: let's just remove support for this, it was deprecated for a while now.

docs/send-transactions.md
```diff
@@ -114,6 +114,27 @@ Specify fields to exclude from the result output. Applies only to the text outpu
...
+
+- Flag: `--payer`
+- Valid inputs: the name of an account defined in the configuration (`flow.json`)
+
+Specify the name of the account that will be used as payer in the transaction.
+
+### Authorizer
+
+- Flag: `--authorizer`
+- Valid inputs: the name of an account defined in the configuration (`flow.json`)
```
sideninja, line 134:
```suggestion
- Valid inputs: the name of a single or multiple comma-separated accounts defined in the configuration (`flow.json`)
```

internal/transactions/send.go
```diff
@@ -21,23 +21,24 @@ package transactions
 import (
 	"fmt"
 
-	"github.com/onflow/flow-cli/pkg/flowkit/config"
```
sideninja, line 24: fix lint

internal/transactions/sign.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 type flagsSign struct {
-	Signer        []string `default:"emulator-account" flag:"signer" info:"name of the account used to sign"`
+	Signer        []string `default:"emulator-account" flag:"signer" info:"name of the account(s) used to sign"`
```
sideninja, line 35:
```suggestion
	Signer        []string `default:"emulator-account" flag:"signer" info:"name of a single or multiple comma-separated accounts used to sign"`
```

internal/transactions/send.go
```diff
@@ -21,23 +21,24 @@ package transactions
...
-	Signer   string   `default:"emulator-account" flag:"signer" info:"Account name from configuration used to sign the transaction"`
-	GasLimit uint64   `default:"1000" flag:"gas-limit" info:"transaction gas limit"`
-	Include  []string `default:"" flag:"include" info:"Fields to include in the output"`
-	Exclude  []string `default:"" flag:"exclude" info:"Fields to exclude from the output (events)"`
+	ArgsJSON  string   `default:"" flag:"args-json" info:"arguments in JSON-Cadence format"`
+	Arg       []string `default:"" flag:"arg" info:"⚠️  Deprecated: use command arguments"`
+	Signer    string   `default:"" flag:"signer" info:"Account name from configuration used to sign the transaction as proposer, payer and suthorizer"`
+	Proposer  string   `default:"" flag:"proposer" info:"Account name from configuration used as proposer"`
+	Payer     string   `default:"" flag:"payer" info:"Account name from configuration used as payer"`
+	Autorizer []string `default:"" flag:"authorizer" info:"Account name(s) from configuration used as authorizer(s)"`
```
sideninja, line 39:
```suggestion
	Autorizer []string `default:"" flag:"authorizer" info:"Name of a single or multiple comma-separated accounts used as authorizers from configuration"`
```

sideninja (commented):
Left some comments

sideninja (approved):
Awesome! Thank you

PR #678, opened 2022-10-17
Write errors to stderr

### LINKED ISSUE #669: Minor: Write errors to stderr instead of stdout
### Problem

I noticed that this error doesn't get printed to stderr:

```
invalid private key for account: emulator-account
```

After some investigation, I saw that most of the errors caught at the top level (e.g. when triggering the emulator start command) are written to stdout instead of stderr, because this utility is used:

https://github.com/onflow/flow-cli/blob/30e91b0a05d7bbd2ce911ba0e4a1c763eb6e113a/pkg/flowkit/util/cli.go#L32-L35

I think that it would be nice if we write all error messages to stderr as a general practice. Some benefits of this:
- some dev tools (like [Flowser](https://github.com/onflowser/flowser)) can automatically detect errors and provide better feedback to the user
- users running the `flow emulator` command, can filter error logs (e.g. `flow emulator > emulator.log 2> emulator-error.log`)

### Steps to Reproduce

1. Run the emulator
2. Trigger an error (e.g. provide an invalid key in flow.json)

### Acceptance Criteria

Flow-cli error logs are written to STDERR.

### Context

I'm working on [Flowser tool](https://github.com/onflowser/flowser), which automatically detects errors (those logs emitted to stderr) and provides a visual feedback to the user.

bluesign:
Closes #669 

## Description

This should be discussed as it is a breaking change and can break someone's workflow. 

______

For contributor use:

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"runtime/debug"
 	"strings"
 	"time"
@@ -302,7 +303,7 @@ func initCrashReporting() {
 		},
 	})
 	if err != nil {
-		fmt.Println(err) // safest output method at this point
+		fmt.Fprintln(os.Stderr, err) // safest output method at this point
 	}
 }
 
--- a/internal/command/flags.go
+++ b/internal/command/flags.go
@@ -20,6 +20,7 @@ package command
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/psiemens/sconfig"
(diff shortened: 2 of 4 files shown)


sideninja:
This is good, thank you. Just wondering how to make it so it doesn't break for someone that was piping outputs etc. I feel if we keep both for some time won't help much as anyhow we can't detect usage of each stream. Maybe just an appropriate note in release notes could be enough. Any thoughts/ideas?

bluesign:
@sideninja that is my concern too actually, maybe we announce and then merge in few releases ahead?

sideninja (approved):

PR #680, opened 2022-10-20
Transaction API refactor

sideninja:
## Description
This refactors transaction API with a couple of improvements:
- less arguments and better defined types
- surface prompts to the command
- allow further extension of multiple roles for send
______

For contributor use:


pkg/flowkit/services/transactions.go
```diff
@@ -79,40 +79,92 @@ func (t *Transactions) GetStatus(
...
 }
 
+// TransactionAccounts define all the accounts for different transaction roles.
+//
+// If all roles are defined by the same account you can only set the payer.
+// You can read more about roles here: https://developers.flow.com/learn/concepts/accounts-and-keys
+type TransactionAccounts struct {
+	Proposer    *flowkit.Account
+	Authorizers []*flowkit.Account
+	Payer       *flowkit.Account
```
bluesign, line 89: I think here can be more than one payer ( for multisign )

sideninja: good idea

sideninja: but I will let the be part of another PR, we just need to add a method like `NewMultiSigTransactionAccounts`

pkg/flowkit/services/transactions.go
```diff
@@ -223,60 +288,52 @@ func (t *Transactions) SendSigned(
...
 
-	sentTx, err := t.gateway.SendSignedTransaction(signed)
+	sentTx, err := t.gateway.SendSignedTransaction(tx)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	t.logger.StopProgress()
 	t.logger.StartProgress("Waiting for transaction to be sealed...")
+	defer t.logger.StopProgress()
```
janezpodhostnik, line 333: how about:
```
func withLoggerProgress(logger, msg string ,f func() ) {
   logger.StartProgress(msg)
   defer logger.StopProgress()
   f()
}
```

internal/transactions/sign.go
```diff
@@ -105,10 +105,15 @@ func sign(
...
 	for _, signer := range signers {
-		signed, err = services.Transactions.Sign(signer, payload, globalFlags.Yes)
+		signed, err = services.Transactions.Sign(signer, payload)
 		if err != nil {
 			return nil, err
 		}
+
+		if !globalFlags.Yes && !output.ApproveTransactionForSigningPrompt(tx) {
+			return nil, fmt.Errorf("transaction was not approved for signing")
+		}
```
janezpodhostnik, line 115: shouldn't this happen before `services.Transactions.Sign`?

pkg/flowkit/services/transactions.go
```diff
@@ -79,40 +79,133 @@ func (t *Transactions) GetStatus(
 	return tx, result, err
 }
 
-// Build builds a transaction with specified payer, proposer and authorizer.
-func (t *Transactions) Build(
+// NewTransactionAccounts defines transaction roles by accounts.
+//
+// You can read more about roles here: https://developers.flow.com/learn/concepts/accounts-and-keys
+func NewTransactionAccounts(
```
janezpodhostnik, line 85: minor: the naming isn't clicking with me...

How about something like: `NewTransactionAccountRoles` and `NewTransactionAccountRolesFromSingleAccount` (ok, that is a bit long), `transactionAccountRoles`, `transactionAddressRoles`...

pkg/flowkit/services/transactions.go
janezpodhostnik: > I think here can be more than one proposer & payer ( for multisign )

I think there can be only one payer/proposer address, but it can have multiple required keys, so needs multi-sign.

pkg/flowkit/services/transactions.go
sideninja: Yeah not great naming, couldn't think of others. NewTransactionAccountRoles this one I like.

sideninja:
https://github.com/onflow/flow-cli/pull/680#discussion_r1004633959
We will add that once we support it in the command, but yeah I agree.

pkg/flowkit/services/transactions.go
bluesign: named return values are an option, but kills the readability. Maybe something like scoped progress with finalizer can be useful.

sideninja: > scoped progress with finalizer can be useful.

Can you elaborate? meaning returning a func that stops the progress?

bluesign: hmm nvm, I thought a bit like cpp.

internal/transactions/sign.go
```diff
@@ -105,10 +105,15 @@ func sign(
 	})
 
 	for _, signer := range signers {
-		signed, err = services.Transactions.Sign(signer, payload, globalFlags.Yes)
+		if !globalFlags.Yes && !output.ApproveTransactionForSigningPrompt(tx) {
```
DylanTinianov, line 108: What is globalFlags.Yes ?

sideninja: :D there's a `-Y` global flag that allows you to accept any prompts, that's an established pattern in CLIs.

DylanTinianov (approved):

PR #688, opened 2022-11-06
P256 key generation with SLIP10

bluesign:
Closes #???

## Description

Hao from lilico alerter on discord about SLIP10 [0] [1]; when I was developing BIP44 I developed with BIP32 ( which is using same private key generation for P256 as sec256k1 )

But seems https://github.com/satoshilabs has some SLIP for this, and it is even mentioned in our FLIP about hd wallet generation.  

So I fixed our implementation to use SLIP10 instead of BIP32. 

[0] https://discord.com/channels/613813861610684416/709804784089301002/1038105907806081064
[1] https://github.com/satoshilabs/slips/blob/master/slip-0010.md
______

For contributor use:

### DIFF
--- a/pkg/flowkit/services/keys_test.go
+++ b/pkg/flowkit/services/keys_test.go
@@ -53,10 +53,10 @@ func TestKeys(t *testing.T) {
 		t.Parallel()
 
 		_, s, _ := setup()
-		key, err := s.Keys.DerivePrivateKeyFromMnemonic("isolate visa defy link gate ordinary notice desert punch zero please pistol", crypto.ECDSA_P256, "")
+		key, err := s.Keys.DerivePrivateKeyFromMnemonic("normal dune pole key case cradle unfold require tornado mercy hospital buyer", crypto.ECDSA_P256, "")
 
 		assert.NoError(t, err)
-		assert.Equal(t, key.String(), "0x04430b77d604b52815494a8e453e1ffa2483644c063a0b7a46eb5effb3dc7b0b")
+		assert.Equal(t, key.String(), "0x638dc9ad0eee91d09249f0fd7c5323a11600e20d5b9105b66b782a96236e74cf")
 	})
 
 	//https://github.com/onflow/ledger-app-flow/blob/dc61213a9c3d73152b78b7391d04165d07f1ad89/tests_speculos/test-basic-show-address-expert.js#L28
--- a/pkg/flowkit/services/keys_test.go
+++ b/pkg/flowkit/services/keys_test.go
@@ -22,9 +22,6 @@ import (
 	"encoding/hex"
 	"testing"
 
-	goeth "github.com/ethereum/go-ethereum/accounts"
-	slip10 "github.com/lmars/go-slip10"
-
 	"github.com/onflow/flow-go-sdk/crypto"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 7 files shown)


sideninja:
@bluesign can you please add a description of why this PR is an improvement?

sideninja (approved):

pkg/flowkit/services/keys_test.go
```diff
@@ -53,10 +53,10 @@ func TestKeys(t *testing.T) {
 		t.Parallel()
 
 		_, s, _ := setup()
-		key, err := s.Keys.DerivePrivateKeyFromMnemonic("isolate visa defy link gate ordinary notice desert punch zero please pistol", crypto.ECDSA_P256, "")
+		key, err := s.Keys.DerivePrivateKeyFromMnemonic("normal dune pole key case cradle unfold require tornado mercy hospital buyer", crypto.ECDSA_P256, "")
 
 		assert.NoError(t, err)
-		assert.Equal(t, key.String(), "0x04430b77d604b52815494a8e453e1ffa2483644c063a0b7a46eb5effb3dc7b0b")
+		assert.Equal(t, key.String(), "0x638dc9ad0eee91d09249f0fd7c5323a11600e20d5b9105b66b782a96236e74cf")
```
tarakby, line 59: I'm wondering how the expected value was computed? Is there a "source of truth" tool that can be referenced?

tarakby (changes requested):
The P-256 curve key-derivation process is indeed described by SLIP10 and not BIP32. The multi-account FLIP also [mentions that](https://github.com/onflow/flips/blob/main/flips/20201125-bip-44-multi-account.md#path-levels).

The PR implementation is based on the package `github.com/lmars/go-slip10`. I had a look at its code base and I noticed that its tests only cover secp256k1 curve (unless I read wrong). As far as I can see, there are no tests for P-256 curve. 
One particular [function](https://github.com/lmars/go-slip10/blob/master/utils.go#L148) caught my attention and it doesn't seem correct in the P-256 case (it assumes `a=0` while that is only true for secp256k1). That would result in incorrect results, and that's why I'm not confident about using the package. 

SLIP10 [provides](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vectors) plenty of test vectors that we should be using to validate our implementation. I suggest we integrate those to our tests. If P-256 tests do not pass, we could restrict the CLI support to secp256k1 only for now, and work on a solution for P-256.

bluesign:
@tarakby thanks for looking into. I actually tested with slip10 test vectors, but eventually I forgot to add tests here, I will add now.

tarakby:
Thank you @bluesign for updating the tests! 
Sorry if I wasn't clear. What I meant is to add tests for the current repo implementation, specifically for `newBip44AccountKey(..).PrivateKey()` function. I think using `github.com/lmars/go-slip10` is an internal implementation detail and that our tests should pass regardless of the internal method we use. If we were to use another package in the future, our implementation-agnostic test files would still be relevant. 
Since our implementation only works with the lower level of the derivation path (we do not output the intermediate keys), we could test our implementation only against the lower key in the test vector. I think there are 4 keys in total to test against in the test vector (2 for each curve). 
What do you think?

I can add a quick PR against yours if I'm bad at explaining with words 😄

bluesign:
@tarakby, that would be great; ( probably I am the one bad with understanding, I am not good at testing in general )

sideninja:
Just a note here but I feel this should be at some later time moved to Go SDK.

bluesign:
@tarakby that looks great, now I understand what you meant, I will merge them asap.

bluesign:
merged @tarakby changes and rebased.

sideninja:
@tarakby are we good now with this PR?

sideninja:
I believe @tarakby is OOO so we will have to wait for his approval.

tarakby (approved):
Very Sorry for blocking this PR for so long! I missed approving it early and had limited online access

tarakby:
@bluesign @sideninja please feel free to merge at your convenience

sideninja:
> Very Sorry for blocking this PR for so long! I missed approving it early and had limited online access

No worries, I hope you had a nice time recharging.

Issue #733, opened 2022-12-13
Resolving imports with the new schema


sideninja:
Epic: #710 

Resolve the imports with the new schema. 
An example would be:
```
// Foo.cdc
pub contract Foo {}

// Bar.cdc
import Foo

pub contract Bar{}
```
The Bar.cdc would resolve the import of `import Foo` to the Foo contract in the state previously build by the dev command.

Issue #732, opened 2022-12-09
Implementing file watcher and build state


sideninja:
EPIC: #710 

Implement a file watcher to track files in contracts, transactions, and scripts and build the state based on files observed. The state should also be synced with every restart and change in folders.

PR #734, opened 2022-12-13
Dev supercommand

### LINKED ISSUE #710: [EPIC] CLI dev command
Flow dev `flow dev` would start up all the tools (emulator, dev-wallet, flowser) and continuously sync all contracts on the emulator. If there are changes in the contract it would watch in that folder and redeploy to keep things up to date. If there are problems in the source code it would communicate those issues in a clear way.

Adding and importing in contracts would also be improved as explained in the bellow section. Developers will be able to import contracts without any additional configuration or understanding or any context switches. 

An example of flow would be: a developer creates a `Foo.cdc` contract and then creates the `Bar.cdc` that imports the first Foo contract. 

```
// Foo.cdc
pub contract Foo {}

// Bar.cdc
import Foo

pub contract Bar{}
```

That would automatically find possible `./contracts/Foo.cdc` and add it to flow.json and deploy without much overhead and configuration. If it wouldn’t find it it would warn the user during deployment to provide a source for that contract. File watchers could detect any issues with the structures/content/configuration and warn the developer when they are doing something that's not suggested.

Tests should also be run automatically when changes are made.

- [x] #732 
- [x] #733

sideninja:
Closes #710 #733 #732 

## Description
This PR implements the `flow dev` command which is a project monitor that keeps your project in-sync with the source code on the emulator. It continuously watches for changes and redeploys everytime as well as creates all the accounts it needs. It also supports the new import syntax `import "Identifier"` which will resolve the import by the contract name in flow.json which the dev command also keeps in-sync.  
______

For contributor use:


internal/super/files.go
```diff
@@ -0,0 +1,231 @@
...
+			select {
+			case event := <-f.watcher.Event:
+				rel, err := f.relProjectPath(event.Path)
+				if err != nil { // skip if failed
+					continue
+				}
+
+				name, containsAccount := accountFromPath(rel)
+				if event.IsDir() && containsAccount {
+					// todo handle rename and move
```
DylanTinianov, line 150: Separate PR for this later?

internal/super/output.go
```diff
@@ -0,0 +1,162 @@
...
+		return
+	}
+
+	fmt.Println(okBanner())
+	fmt.Println(successfulDeployment(deployed))
+}
+
+func successfulDeployment(deployed []*flowkitProject.Contract) string {
+	var out bytes.Buffer
+	okFaces := []string{"😎", "😲", "😱", "😜"}
```
DylanTinianov, line 55: Maybe consider adding cowboy dude 🤠

DylanTinianov (approved):
Looks amazing, let's do some additional testing once merged.

internal/super/files.go
sideninja: yeah, this is a bit edge case I feel, we can cover later

PR #758, opened 2023-01-05
specify compatible environments

MrDSGC:
## Description

update docs to match actual behavior: 

- "no longer able to remove contracts from accounts"
- added disclaimer that removing contracts is only possible in testnet/emulator

### DIFF
--- a/docs/account-remove-contract.md
+++ b/docs/account-remove-contract.md
@@ -2,6 +2,7 @@
 title: Remove a Contract with the Flow CLI
 sidebar_title: Remove a Contract
 ---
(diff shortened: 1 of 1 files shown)


docs/account-remove-contract.md
```diff
@@ -2,6 +2,7 @@
 title: Remove a Contract with the Flow CLI
 sidebar_title: Remove a Contract
 ---
+_This feature is only found in Emulator and Testnet environments. You **cannot** remove a contract on Mainnet._
```
turbolent, line 5:
```suggestion
_This feature is only found in Emulator and Testnet environments. You **cannot** remove a contract on Testnet or Mainnet._
```

turbolent (approved):

PR #763, opened 2023-01-06
Fix project deploy edge case with account key sequence

### LINKED ISSUE #754: Account key sequence number not being correctly set
There are cases where during the deployment of a project incorrect sequence for a key is set on a transaction. I believe this might be coming from the fact that there's one transaction that is not yet finalized but the key is at that point fetched and before sending the next transaction with same sequence key the previous one is finalized. Making the latter one invalid. Look into how we obtain the key sequence in case of no diff here https://github.com/onflow/flow-cli/blob/master/pkg/flowkit/services/project.go#L331

sideninja:
Closes #754 

## Description
Waits for a transaction to be sealed before getting account info, otherwise account key sequence could not yet be updated.

______

For contributor use:


DylanTinianov (approved):
Nice changes!

PR #776, opened 2023-01-17
Super commands help improvements

sideninja:
## Description
Add improved template on help screen to guide the user to use the super commands as well as deprecate some that might not be relevant anymore with the new commands.

______

For contributor use:


pkg/flowkit/services/services.go
```diff
@@ -61,3 +61,17 @@ func NewServices(
 		Tests:        NewTests(state, logger),
 	}
 }
+
+func (s *Services) SetLogger(logger output.Logger) {
```
janezpodhostnik, line 65: This seems slightly fragile. Is there a different way to do this?

Maybe all of the "components" (s.Transactions, s.Scripts, ...) should have a common interface and be in a list, over which you can iterate and set the log.

pkg/flowkit/services/scripts_test.go
```diff
@@ -19,6 +19,7 @@
 package services
 
 import (
+	"github.com/onflow/flow-cli/pkg/flowkit"
```
janezpodhostnik, line 22: potential linting issue.

internal/super/output.go
```diff
@@ -0,0 +1,152 @@
...
+		return
+	}
+
+	fmt.Println(okBanner())
+	fmt.Println(successfulDeployment(deployed))
+}
+
+func successfulDeployment(deployed []*flowkitProject.Contract) string {
+	var out bytes.Buffer
+	okFaces := []string{"😎", "😲", "😱", "😜"}
```
janezpodhostnik, line 54: This is cool :). But why is the third one considered an ok-face :)?

Also can `GetRandomOKFace` be its own function somewhere. Sounds useful.

internal/super/project.go
```diff
@@ -0,0 +1,358 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package super
+
+import (
+	"fmt"
```
janezpodhostnik, line 22: potential linting issue. imports not grouped.

internal/super/setup.go
```diff
@@ -0,0 +1,181 @@
...
+		Short:   "Start a new Flow project",
+		Example: "flow setup my-project",
+		Args:    cobra.ExactArgs(1),
+		GroupID: "super",
+	},
+	Flags: &setupFlags,
+	Run:   create,
+}
+
+const scaffoldListURL = "https://raw.githubusercontent.com/onflow/flow-scaffold-list/main/scaffold-list.json"
```
janezpodhostnik, line 39: I'm a bit worried to take this from `/main/` is there any way it can be from a tag instead?

pkg/flowkit/config/deployment.go
```diff
@@ -97,3 +97,32 @@ func (d *Deployments) Remove(account string, network string) error {
 
 	return nil
 }
+
+// AddContract to deployment list on the account name and network name.
+//
+// Deployment needs to be previously created by calling AddOrUpdate otherwise the contract is not added.
+func (d *Deployments) AddContract(account string, network string, contract ContractDeployment) {
+	for i, deploy := range *d {
+		if deploy.Network == network && deploy.Account == account {
```
janezpodhostnik, line 106: Invert condition to reduce nesting, same below.

janezpodhostnik (commented):
This review was very long :sweat_smile:. It feels like it could have been multiple PRs

I might have missed something, I went fast through some parts, but looks good to me, just had some minor comments.

pkg/flowkit/services/services.go
sideninja: I plan to refactor and do just that, didn't want to also do that in this PR tho, but there's an upcoming refactor happening soon improving bunch of ugly stuff like this.

internal/super/output.go
sideninja: haha yeah I might need to update emojis to be more consistently OK

sideninja: Probably the first time when emojis are discussed in PR review :D

internal/super/setup.go
sideninja: I can see what you mean, but my thinking is: main is protected, so any changes to that list mush be approved, and that is not the main reason, which is, I want to make this accessible for anyone to add to, and if we also require a tag then they will have to first make a PR which gets merged and then ask us to tag it and then make another PR to update the version here, so a bit of extra friction. The thing tho, which is here not visible is that I plan to close that repo and just have the list in the CLI repo. I plan to fix that today.

DylanTinianov (approved):

PR #799, opened 2023-01-26
Deploy ignore non-existing contract

sideninja:
## Description
Ignore failure to remove during deploy on the emulator. This ignores non-existing contract error which is ok to skip to deploy a new contract.
______

For contributor use:

### DIFF
--- a/pkg/flowkit/services/project.go
+++ b/pkg/flowkit/services/project.go
@@ -250,11 +250,7 @@ func (p *Project) Deploy(network string, update bool) ([]*project.Contract, erro
 		// special case for emulator updates, where we remove and add a contract because it allows us to have more freedom in changes.
 		// Updating contracts is limited as described in https://developers.flow.com/cadence/language/contract-updatability
 		if update && network == config.DefaultEmulatorNetwork().Name {
-			_, err = accounts.RemoveContract(targetAccount, contract.Name)
-			if err != nil {
-				deployErr.add(contract, err, fmt.Sprintf("failed to remove the contract %s before the update", contract.Name))
-				continue
-			}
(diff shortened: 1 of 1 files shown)


pkg/flowkit/services/project.go
```diff
@@ -250,11 +250,7 @@ func (p *Project) Deploy(network string, update bool) ([]*project.Contract, erro
 		// special case for emulator updates, where we remove and add a contract because it allows us to have more freedom in changes.
 		// Updating contracts is limited as described in https://developers.flow.com/cadence/language/contract-updatability
 		if update && network == config.DefaultEmulatorNetwork().Name {
-			_, err = accounts.RemoveContract(targetAccount, contract.Name)
-			if err != nil {
-				deployErr.add(contract, err, fmt.Sprintf("failed to remove the contract %s before the update", contract.Name))
-				continue
-			}
+			_, _ = accounts.RemoveContract(targetAccount, contract.Name) // ignore failure as it's meant to be best-effort
```
janezpodhostnik, line 253: Maybe it would be useful to still log it under debug level or something, in case someone needs to figure out what happened.

janezpodhostnik (approved):

pkg/flowkit/services/project.go
sideninja: Actually this only serves one specific case where we intentionally need to skip errors, this is because on emulator we want to remove the contract instead of updating as it gives us more freedom to change it, and if contract is non-existing it returns an error which is fine in this case. I could compare the error if it's the non-existing one but it just adds complexity then that I don't really care about, even if it fails to remove it should still work, so this is a bit weird case.

PR #853, opened 2023-02-15
Account creation using APIs

### LINKED ISSUE #597: Support creation of accounts on testnet and mainnet using APIs

sideninja:
Closes #597 

## Description
This PR implements APIs for creating the account on testnet and mainnet. It also writes the account keys to a private key file and use the newest schema for specifying the key in flow.json which works with FCL.

______

For contributor use:


internal/accounts/create-interactive.go
```diff
@@ -0,0 +1,266 @@
...
+		http.MethodPost,
+		fmt.Sprintf("https://openapi.lilico.org/v1/address%s", apiNetwork),
+		bytes.NewReader(data),
+	)
+	request.Header.Add("Content-Type", "application/json; charset=UTF-8")
+	request.Header.Add("Authorization", lilicoToken)
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // lilico api doesn't yet have a valid cert, todo reevaluate
```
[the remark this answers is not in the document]
sideninja: in this case we are forced to since the API doesn't have a valid cert

internal/accounts/create-interactive.go
```diff
@@ -0,0 +1,268 @@
...
+	return flowkit.NewAccount(name).SetAddress(networkAccount.Address).SetKey(
+		flowkit.NewHexAccountKeyFromPrivateKey(0, crypto.SHA3_256, key),
+	), nil
+}
+
+func getAccountCreationResult(services *services.Services, id flow.Identifier) (*flow.TransactionResult, error) {
+	_, result, err := services.Transactions.GetStatus(id, true)
+	if err != nil {
+		if status.Code(err) == codes.NotFound { // if transaction not yet propagated, wait for it
+			time.Sleep(1 * time.Second)
```
DylanTinianov, line 187: Would be nice to use a wait group or future/ channel to know when a transaction has been propagated. But if this doesn't come up much then it's okay.

internal/accounts/create.go
```diff
@@ -60,18 +54,15 @@ var CreateCommand = &command.Command{
...
 ) (command.Result, error) {
 	// if user doesn't provide any flags go into interactive mode
 	if len(createFlags.Keys) == 0 {
-		_, err := createInteractive(state, loader)
-		if err != nil {
-			return nil, err
-		}
-		return nil, nil
+		err := createInteractive(state)
+		return nil, err
```
DylanTinianov, line 65:
```suggestion
		return nil, createInteractive(state)
```

sideninja: nice

DylanTinianov (approved):
Couple small comments, but nice work 👍

internal/accounts/create-interactive.go
sideninja: hmm, what do you mean? 
So the issue here is that when we use the account creation API they send us back the transaction ID immediately since that can be calculated offchain and we don't know when it will actually be propagated to network, that we don't have actual control over since we are not the ones who submit it, so it wouldn't be possible to have like a back channel to indicate that. A good idea in that case. The only thing we could is to monitor events and then as soon as the event with that account comes we know it was created but that has different downsides and in this case it only maybe takes one or two seconds to get there.

PR #869, opened 2023-02-22
Use API for account creation

sideninja:
## Description
Use an API for account creation instead of internally creating the accounts.

______

For contributor use:


Makefile
```diff
@@ -13,6 +13,9 @@ PATH := $(PATH):$(GOPATH)/bin
 # OS
 UNAME := $(shell uname)
 
+MIXPANEL_PROJECT_TOKEN := 3fae49de272be1ceb8cf34119f747073
+ACCOUNT_TOKEN := lilico:sF60s3wughJBmNh2
```
bluesign, line 17: is this for mainnet ?

sideninja: yes

sideninja: don't even get me started... apparently, there's no other way if use homebrew... since homebrew builts its own binaries and hence can not access the secrets we use in CD.

internal/accounts/create-interactive.go
```diff
@@ -80,145 +102,161 @@ func createInteractive(state *flowkit.State) error {
...
-	address, err := sendCreateAccountTransaction(signer, key, network, service)
+	err = state.ReaderWriter().WriteFile(privateFile, []byte(key.String()), os.FileMode(0644))
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed saving private key: %w", err)
 	}
 
-	return flowkit.NewAccount(name).SetAddress(*address).SetKey(accountKey), nil
+	return flowkit.NewAccount(name).SetAddress(*address[0]).SetKey(
+		flowkit.NewFileAccountKey(privateFile, 0, crypto.ECDSA_P256, crypto.SHA3_256),
```
DylanTinianov, line 151: Might be nice to define the crypto as constants since it's used everywhere
```
const defaultSignAlg = crypto.ECDSA_P256
const defaultHashAlg = crypto.SHA3_256
```

DylanTinianov (approved):
Small comment, but looks good 👍

internal/accounts/create-interactive.go
sideninja: @DylanTinianov addressed

PR #891, opened 2023-03-03
remove logrus logger and add zerolog Logger to flowkit backend config

bjartek:
in commit #a03e87a03 the emulator was updated without adding a way for thirdparties to set the logger. 

This pr removes logrus and adds zerolog to the EmulatorGateway so that it is possible again.


bjartek:
It also looks like WithEmulatorLogger is not used anymore. It is the logger sent into the backend that is used for both logging meters and logging from emulator.

pkg/flowkit/gateway/emulator.go
```diff
@@ -57,21 +56,21 @@ func NewEmulatorGatewayWithOpts(serviceAccount *flowkit.Account, opts ...func(*E
...
+		logger:          &zerolog.Logger{},
 		emulatorOptions: []emulator.Option{},
 	}
 	for _, opt := range opts {
 		opt(gateway)
 	}
 
 	gateway.emulator = newEmulator(serviceAccount, gateway.emulatorOptions...)
-	gateway.backend = backend.New(&zerolog.Logger{}, gateway.emulator)
+	gateway.backend = backend.New(gateway.logger, gateway.emulator)
```
janezpodhostnik, line 67: Unrelated: 
  Hm... Interesting that zerolog is passed by reference. I'm not sure it should be. Something to look into...

pkg/flowkit/gateway/emulator.go
```diff
@@ -57,21 +56,21 @@ func NewEmulatorGatewayWithOpts(serviceAccount *flowkit.Account, opts ...func(*E
 
 	gateway := &EmulatorGateway{
 		ctx:             context.Background(),
-		logger:          logrus.New(),
+		logger:          &zerolog.Logger{},
```
janezpodhostnik, line 59:
```suggestion
		logger:          &zerolog.Nop(),
```

pkg/flowkit/gateway/emulator.go
bluesign: hmm I was thinking on this when moving from logrus to zerolog, but I felt reference passing is correct. But I maybe totally missing the point. Why do you think copy is better @janezpodhostnik  ?

sideninja: I think the issue might be that the receiver can have an influence on the logging of the other components.

sideninja: In any case I don't think this is the subject of this PR

sideninja (approved):

pkg/flowkit/gateway/emulator.go
bluesign: I think in that case pass by reference is still correct. Sender always have option to send a new logger, but on the other case, sender loses influence on the component. ( e.g. change logging level in runtime )

janezpodhostnik (approved):

bjartek:
dammit what did I do to this pr now, i just merged in master, is this ok or should i clean it up?

bjartek:
i tried to update master but it just created a mess so I will leave it like it is.

sideninja:
@bjartek could you update changes from master so we can get this in.

bjartek:
@sideninja master merged. feel free to squash into main

sideninja:
@bjartek the tests fail, can you check what is the issue please.

PR #965, opened 2023-04-06
Persist developer created accounts

### LINKED ISSUE #949: Maintain Account Key in fresh Flow Dev Runs
### Issue To Be Solved

The objective is for the account objects that are in the flow.json file to maintain the key specified there when running the `flow dev` command to set up the project in a fresh emulator run.

Currently when we want to deploy a contract with the specified account in the flow.json through the flow dev command, a new key is generated and replaces the existing. It causes incompatibility with programs that use that key alongside the emulator within the development cycle as it always changes.

Not a technical solution, but hopefully there is a way to maintain those keys similar to what we can do to the service account key if we specify it.

If there are multiple solutions, please present each one separately. Save comparisons for the very end.)

### (Optional): Context

When simulating a wallet and the contracts deployed by it, there is no control over the key that's used to deploy it, even though the account address does remain the same. It makes it hard to use `flow dev` and any outside program to interact with them

chasefleming:
Closes #949 

## Description

Do not remove accounts added by developer. Only remove those generated to super commands.
______

For contributor use:

### DIFF
--- a/internal/super/project.go
+++ b/internal/super/project.go
@@ -113,6 +113,10 @@ func (p *project) deploy() {
 	printDeployment(deployed, err, p.pathNameLookup)
 }
 
+func (p *project) AddSuperGeneratedAccount(name string) {
+	p.superGeneratedAccounts = append(p.superGeneratedAccounts, name)
+}
+
 func (p *project) isSuperGeneratedAccount(name string) bool {
 	for _, a := range p.superGeneratedAccounts {
 		if a == name {
@@ -231,7 +235,7 @@ func (p *project) addAccount(name string) error {
 		SetAddress(flowAcc.Address).
 		SetKey(flowkit.NewHexAccountKeyFromPrivateKey(0, crypto.SHA3_256, *privateKey))
 
-	p.superGeneratedAccounts = append(p.superGeneratedAccounts, account.Name())
+	p.AddSuperGeneratedAccount(account.Name())
 
 	p.state.Accounts().AddOrUpdate(account)
 	p.state.Deployments().AddOrUpdate(config.Deployment{ // init empty deployment
--- a/internal/super/project.go
+++ b/internal/super/project.go
@@ -58,12 +58,13 @@ func newProject(
 }
 
 type project struct {
-	service        *flowkit.Account
-	services       *services.Services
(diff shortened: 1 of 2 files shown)


internal/super/project.go
```diff
@@ -58,12 +58,13 @@ func newProject(
...
-	readerWriter   flowkit.ReaderWriter
-	projectFiles   *projectFiles
-	pathNameLookup map[string]string
+	service                *flowkit.Account
+	services               *services.Services
+	state                  *flowkit.State
+	readerWriter           flowkit.ReaderWriter
+	projectFiles           *projectFiles
+	pathNameLookup         map[string]string
+	superGeneratedAccounts []string // accounts created by "flow dev" super commands and not developer
```
sideninja, line 67: Since this is only accessible withing the dev command we can avoid adding super in the name, otherwise we should put it on all other vars, so I just suggest changing to something like createdAccounts, I would even say you can just make it list of accounts types, but that's more subjective.
```suggestion
	createdAccounts []*flowkit.Account 
```

internal/super/project.go
```diff
@@ -112,6 +113,19 @@ func (p *project) deploy() {
 	printDeployment(deployed, err, p.pathNameLookup)
 }
 
+func (p *project) AddSuperGeneratedAccount(name string) {
```
sideninja, line 116: This function is a bit light, I would avoid it and just use append directly where you need to.

internal/super/project.go
```diff
@@ -112,6 +113,19 @@ func (p *project) deploy() {
...
+	p.superGeneratedAccounts = append(p.superGeneratedAccounts, name)
+}
+
+func (p *project) isSuperGeneratedAccount(name string) bool {
```
sideninja, line 120: Same as above, I would just use `slices.Contains(p.createdAccounts, name)`

sideninja (commented):
Left some code-style comments, but also make sure to test this manually, I think you currently have a bug if accounts already exist in flow.json from either previous runs of flow dev or from manual creation of accounts before running flow dev.

sideninja (approved):
Thank you. Just make sure to manually test this as well.

PR #996, opened 2023-04-25
Optional argument support 

bluesign:
Added optional argument support for transaction and script arguments parsing.

______

For contributor use:

### DIFF
--- a/flowkit/arguments/arguments.go
+++ b/flowkit/arguments/arguments.go
@@ -123,11 +123,13 @@ func ParseWithoutType(args []string, code []byte, fileName string) (scriptArgs [
 	for index, argumentString := range args {
 		astType := parameterList[index].TypeAnnotation.Type
 		semaType := checker.ConvertType(astType)
+		isOptional := false
 
 		for {
 			switch v := semaType.(type) {
 			case *sema.OptionalType:
 				semaType = v.Type
+				isOptional = true
 				continue
 
 			case *sema.SimpleType:
@@ -145,14 +147,21 @@ func ParseWithoutType(args []string, code []byte, fileName string) (scriptArgs [
 			break
 		}
 
-		var value, err = runtime.ParseLiteral(argumentString, semaType, inter)
-		if err != nil {
-			return nil, fmt.Errorf(
-				"argument `%s` is not expected type `%s`",
-				parameterList[index].Identifier,
-				semaType.QualifiedString(),
-			)
+		var value cadence.Value
+
+		if isOptional && argumentString == "nil" {
+			value = cadence.NewOptional(nil)
+		} else {
+			value, err = runtime.ParseLiteral(argumentString, semaType, inter)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"argument `%s` is not expected type `%s`",
+					parameterList[index].Identifier,
+					semaType.QualifiedString(),
+				)
+			}
 		}
+
 		resultArgs = append(resultArgs, value)
 	}
 	return resultArgs, nil


sideninja (commented):
Nice 👍 
Could you add a test for optional in arguments_test

flowkit/arguments/arguments.go
```diff
@@ -145,14 +147,21 @@ func ParseWithoutType(args []string, code []byte, fileName string) (scriptArgs [
...
-		var value, err = runtime.ParseLiteral(argumentString, semaType, inter)
-		if err != nil {
-			return nil, fmt.Errorf(
-				"argument `%s` is not expected type `%s`",
-				parameterList[index].Identifier,
-				semaType.QualifiedString(),
-			)
+		var value cadence.Value
+
+		if isOptional && argumentString == "nil" {
```
janezpodhostnik, line 152: can we add this to the doc? I will always be unsure if I need to use `nil` or `null` or something else.

janezpodhostnik (approved):

flowkit/arguments/arguments.go
```diff
@@ -145,14 +147,21 @@ func ParseWithoutType(args []string, code []byte, fileName string) (scriptArgs [
...
+			value = cadence.NewOptional(nil)
```
turbolent, line 153: The value must be wrapped in as many additional optionals as the type indicates. For example:
- `T?`: `cadence.NewOptional(nil)`
- `T??`: `cadence.NewOptional(cadence.NewOptional(nil))`

flowkit/arguments/arguments.go
```diff
@@ -145,14 +147,21 @@ func ParseWithoutType(args []string, code []byte, fileName string) (scriptArgs [
 			break
 		}
 
-		var value, err = runtime.ParseLiteral(argumentString, semaType, inter)
```
turbolent, line 148: Maybe instead/also report this upstream in https://github.com/onflow/cadence.
`ParseLiteral` should support `nil` and optional types, I'm surprised it does not at the moment.

turbolent (commented):
Thank you for opening a fix so quickly!

flowkit/arguments/arguments.go
turbolent: From what I can see, Cadence already supports parsing nil/optionals, here's a test case  for it: https://github.com/onflow/cadence/blob/5c53da527bdd3b814382bb9c87cd99e5351616e4/runtime/literal_test.go#L69-L80. What is being passed as arguments to `ParseLiteral` that makes it return an error?

flowkit/arguments/arguments.go
bluesign: thanks I will fix asap.

flowkit/arguments/arguments.go
bluesign: I think sematype check is erroring then, let me try to fix that way

turbolent: The bug is likely the unwrapping of the optional type here: https://github.com/onflow/flow-cli/pull/996/files#diff-116113b82ac2b212970f78016025c6aa67609801d3ae0d98aa3cc00fafbb7e29R131

flowkit/arguments/arguments.go
bluesign: fixed with correct parseLiteral usage.

flowkit/arguments/arguments.go
bluesign: fixed in https://github.com/onflow/flow-cli/pull/996/commits/83a5d514508c8623a655291f5477978a60c62a5c

flowkit/arguments/arguments.go
bluesign: sure I will add to this PR

flowkit/arguments/arguments.go
bluesign: @turbolent hmm how can I check if it is Address or String, I need for that part only to unwrap. Actually commit from you would be the best if not much hassle

turbolent: Refactored the function and improved the tests in c3a5c074293aa93c9f5f4d7d14898e92fbeb4722

Issue #1000, opened 2023-04-25
Regression: show diff flag no longer working after 1.0.0


chasefleming:
### Problem

After 1.0.0, `--show-diff` no longer works and fails to let you continue on because of log.

sideninja:
@chasefleming I can't reproduce the issue, here's trying to deploy a project:

```
./main project deploy --show-diff --update

Deploying 2 contracts for accounts: emulator-account

Bar -> 0xf8d6e0586b0a20c7 [skipping, no changes found]
import Bar from 0xf8d6e0586b0a20c7
import Zoo from 0xf8d6e0586b0a20c7

pub contract Foo {
	pub var X: String

	pub init() {
		self.X = "MOO"
	}

	pub fun bye() {
		Bar.bye()
	}
}

? Do you wish to deploy this contract?? [y/N] █
```

sideninja:
If you mean it doesn't work when adding to add contract then that's not a regression since that feature doesn't exist yet.

chasefleming:
<img width="687" alt="Screenshot 2023-04-26 at 9 50 48 AM" src="https://user-images.githubusercontent.com/1666730/234647164-4bf7ed18-644e-48a3-a264-c74e52366d5f.png">

@sideninja definitely happening for me on master. When you try to input an answer with the logger running there it is erased and you can't continue.

PR #1001, opened 2023-04-25
Add show diff flag to update-contract command and fix regression

### LINKED ISSUE #962: Show contract update diff on update command
The contract update diff should also be accessible on the `flow accounts update-contract` command.

chasefleming:
Closes #962, #1000 

## Description

- Fixes regression after 1.0.0 that broke showing diff on `flow project deploy` (logger runs while waiting for user input)
- Adds `--show-diff` options to `flow accounts update-contract`
______

For contributor use:


flowkit/flowkit.go
```diff
@@ -307,8 +307,6 @@ func (f *Flowkit) AddContract(
 		return flow.EmptyID, false, err
 	}
 
-	f.logger.StartProgress(fmt.Sprintf("Checking contract '%s' on account '%s'...", name, account.Address))
```
sideninja, line 310: I don't like this being removed, it provides valuable progress to the user while the update is happening.

sideninja: What I believe it should be done is for the diff prompt to use the same logger, we should never use fmt.Println to output data, just because of this kind of issues. So make sure to use the logger here https://github.com/onflow/flow-cli/blob/master/internal/util/prompt.go#L257 instead of fmt.Println. That way the logger will hide the progress log.

chasefleming: Changed

internal/util/prompt.go
```diff
@@ -250,23 +250,25 @@ func addAnotherContractToDeploymentPrompt() bool {
 // ShowContractDiffPrompt shows a diff between the new contract and the existing contract
 // and asks the user if they wish to continue with the deployment
 // returns true if the user wishes to continue with the deployment and false otherwise
-func ShowContractDiffPrompt(newContract []byte, existingContract []byte) bool {
-	dmp := diffmatchpatch.New()
-	diffs := dmp.DiffMain(string(newContract), string(existingContract), false)
-	diffString := dmp.DiffPrettyText(diffs)
-	fmt.Println(diffString)
+func ShowContractDiffPrompt(logger *output.Logger) func([]byte, []byte) bool {
```
nvdtf, line 253: `output.Logger` is an interface so we can drop the pointer, This will result in a simpler syntax.

nvdtf (approved):
✨

DylanTinianov (approved):
Look good!

PR #1011, opened 2023-04-27
Add contract to flow.json when adding through accounts add-contract

### LINKED ISSUE #922: Adding a contract doesn't add to flow.json
When we use the `flow accounts add-contract` command we don't alter the flow.json reflecting the deployment. Which is a problem when we are doing imports as they won't get resolved due to contracts not being configured in flow.json.
This might be something we could change, but we need to think if there are any downsides.
cc @DylanTinianov

chasefleming:
Closes #922 

______

For contributor use:


flowkit/flowkit.go
```diff
@@ -365,6 +365,22 @@ func (f *Flowkit) AddContract(
...
+		d.AddContract(config.ContractDeployment{
+			Name: name,
+		})
+	}
+	state.Contracts().AddOrUpdate(config.Contract{
+		Name:     name,
+		Location: contract.Location,
+	})
+
+	err = state.SaveDefault()
```
sideninja, line 379: I'm not sure about this, the problem I see is with all the software that uses flowkit, they won't like if the library starts to write some files to their system, especially so out of their control. I feel the save should be inside the internal as it is specific to CLI.

bjartek: I think having this behind a flag would be nice, I can see where frameworks would want both behaviors.

sideninja: IMHO I think calling the save separately is better than introducing a new function argument (it will make APIs messier). I think flowkit shouldn't be saving the state at any point, I think persisting the state is the responsibility of the consumer.

bjartek: Should we then expose the save function so that the consumer can easily get do it if he did not directly create the flowkit instance?

sideninja (approved):

PR #1013, opened 2023-04-28
Improve coverage report statistics for integration tests

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/132

## Description

The coverage report generated from integration tests, includes all the system contracts (+ some already provided, with the `--contracts` flag), due to the underlying backend implementation using the Flow Emulator. This not only creates "noise", but it also skews the metrics, since these contracts are not really of interest in the context of integration tests.

With the proposed changes, our goal is to:
- Exclude all the system contracts from coverage report (this is currently supported from the `CoverageReport` API, by using the `ExcludeLocation` method)
- Exclude all scripts and transactions from coverage report (this is currently being introduced [upstream](https://github.com/onflow/cadence/pull/2474), in the Cadence repository)

______

For contributor use:

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -113,6 +113,38 @@ func testCode(
 	runner := cdcTests.NewTestRunner()
 	if coverageEnabled {
 		coverageReport = runtime.NewCoverageReport()
+		contracts := map[string]string{
+			"FlowToken":             "0x0ae53cb6e3f42a79",
+			"FlowFees":              "0xe5a8b7f23e8b548f",
+			"FungibleToken":         "0xee82856bf20e2aa6",
+			"FlowClusterQC":         "0xf8d6e0586b0a20c7",
+			"FlowDKG":               "0xf8d6e0586b0a20c7",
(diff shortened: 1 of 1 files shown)


internal/test/test.go
```diff
@@ -113,6 +113,38 @@ func testCode(
 	runner := cdcTests.NewTestRunner()
 	if coverageEnabled {
 		coverageReport = runtime.NewCoverageReport()
+		contracts := map[string]string{
```
m-Peter, line 116: **NOTE:** I am sure there's a way to get these mapping dynamically, I have just hardcoded it here to illustrate my point. We'll update it to dynamically fetch all the available system contracts, as new ones we'll probably be introduced in the future.

m-Peter: This might be better suitable for the `cadence-tools/test` repository.

m-Peter: Update: Opened https://github.com/onflow/cadence-tools/pull/110 in favor of this.

PR #1015, opened 2023-04-28
Add block_id and block_height to transaction result response

### LINKED ISSUE #1014: Include BlockID/Height in transaction response
### Issue To Be Solved

When reviewing a transaction result, it's helpful to know which block included the tx. 

Please add fields for Block ID and Height to the response from `flow transaction get`:
https://github.com/onflow/flow-cli/blob/000fb981daddd1c4a01e763a42f2a73d27e45af3/internal/transactions/transactions.go

The fields should already be available in [`flow.TransactionResult`](https://github.com/onflow/flow-go-sdk/blob/aa03c00bca3ff7d9b1da4bd5cd735c5ec1d5a727/transaction.go#L611-L618), so I _think_ it's just a matter of adding them to the output.

peterargue:
Closes #1014 

## Description

Update the transaction result response to add the ID and height of the block which included the tx.

______

For contributor use:

### DIFF
--- a/internal/transactions/transactions.go
+++ b/internal/transactions/transactions.go
@@ -63,6 +63,8 @@ func (r *transactionResult) JSON() any {
 	result["payer"] = r.tx.Payer.String()
 
 	if r.result != nil {
+		result["block_id"] = r.result.BlockID.String()
+		result["block_height"] = r.result.BlockHeight
 		result["status"] = r.result.Status.String()
 
 		txEvents := make([]any, 0, len(r.result.Events))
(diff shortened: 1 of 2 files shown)


janezpodhostnik:
does this automatically work with "--exclude"?

peterargue:
> does this automatically work with "--exclude"?

from the help output, it looks like the only field that can currently be excluded is events:
```
Flags:
      --exclude strings   Fields to exclude from the output. Valid values: events.
```

These are pretty small fields. Do you think that's something users would want?

janezpodhostnik:
I was just wondering if `--exclude` could be used by anyone that is bothered by this extra information. But, I don't see why they would be really...

janezpodhostnik (approved):

sideninja (approved):
Perfect, thank you

PR #1030, opened 2023-05-18
Allow importing helper/utility files in test files

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/148

## Description

> Able to write modular test cases means a developer can import the utility functions from a cadence file to the test file. to avoid repetitive code.

(taken from https://github.com/onflow/developer-grants/issues/148#issuecomment-1549064834)

**Note:** By convention, we consider files that contain the `_helper` substring, to be helper/utility files. Of course, this is just an initial implementation detail, feel free to suggest an alternative. The main difficulty here is that we can't distinguish from a `common.StringLocation`, if we are dealing with a contract, or with a simple script containing Cadence code. The `importResolver` function, has business logic which rejects any imports that are not declared in the `flow.json` config file.

______

For contributor use:

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -133,8 +133,17 @@ func importResolver(scriptPath string, state *flowkit.State) cdcTests.ImportReso
 		if !isFileImport {
 			return "", fmt.Errorf("cannot import from %s", location)
 		}
-
 		relativePath := stringLocation.String()
+
+		if strings.Contains(relativePath, "_helper") {
+			importedScriptFilePath := absolutePath(scriptPath, relativePath)
+			scriptCode, err := state.ReadFile(importedScriptFilePath)
+			if err != nil {
+				return "", nil
+			}
+			return string(scriptCode), nil
+		}
+
 		contractFound := false
 		for _, contract := range *state.Contracts() {
 			if strings.Contains(relativePath, contract.Location) {
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -37,6 +37,10 @@ import (
 	"github.com/onflow/flow-cli/internal/util"
 )
 
+// Import statements with a path that contain this substring,
+// are considered to be helper/utility scripts for test files.
+const helperScriptSubstr = "_helper"
+
 type flagsTests struct {
 	Cover        bool   `default:"false" flag:"cover" info:"Use the cover flag to calculate coverage report"`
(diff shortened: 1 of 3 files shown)


internal/test/test_test.go
```diff
@@ -124,6 +124,21 @@ func TestExecutingTests(t *testing.T) {
...
 	})
 
+	t.Run("with helper script import", func(t *testing.T) {
+		t.Parallel()
+		_, state, _ := util.TestMocks(t)
+
+		// Execute script
+		script := tests.TestScriptWithHelperImport
+		testFiles := make(map[string][]byte, 0)
+		testFiles[script.Filename] = script.Source
```
turbolent, line 134:
```suggestion
		testFiles := map[string][]byte{
			script.Filename: script.Source,
		}
```

internal/test/test.go
```diff
@@ -133,8 +133,17 @@ func importResolver(scriptPath string, state *flowkit.State) cdcTests.ImportReso
 		if !isFileImport {
 			return "", fmt.Errorf("cannot import from %s", location)
 		}
-
 		relativePath := stringLocation.String()
+
+		if strings.Contains(relativePath, "_helper") {
```
turbolent, line 138: Maybe strip the file extension (`path/filepath` `Ext()`) and check the plain filename contains this as a suffix. Or is the idea that this allows both `_helper` and e.g. `_helpers` like in the two test files?

Also, maybe refactor this magic string into a constant and add a comment explaining it.

turbolent (commented):
Nice

sideninja (approved):
Looks good after Bastian comments.

internal/test/test_test.go
m-Peter: Fixed in https://github.com/onflow/flow-cli/pull/1030/commits/6c6d17453e5a5167fa7846e590bede5a9b78fdfa

internal/test/test.go
m-Peter: > Or is the idea that this allows both _helper and e.g. _helpers like in the two test files?

Yes, exactly :+1: It might be the case that developers will put everything in a single helper file, or split them in several files. So I wanted to allow some flexibility in both plural & singular names.

> Also, maybe refactor this magic string into a constant and add a comment explaining it.

Fixed in https://github.com/onflow/flow-cli/pull/1030/commits/e920d18d486172b5d9dc225ed5a2fb1db92080d8

sideninja:
@m-Peter can I merge and release?

PR #1074, opened 2023-06-11
Add flag for skipping code coverage inspection on scripts and transactions

m-Peter:
## Description

The locations for scripts and transactions cannot be easily attributed to their source files, so we add a new flag allowing the developer to skip them. The coverage for contracts is much more important. E.g:

```json
"coverage": {
    "S.FooContract": {
      "line_hits": {
        "14": 1,
        "18": 9,
        "19": 1,
        "20": 8,
        "21": 1,
        "22": 7,
        "23": 1,
        "24": 6,
        "25": 1,
        "26": 5,
        "27": 0,
        "30": 5,
        "31": 4,
        "34": 1,
        "6": 1
      },
      "missed_lines": [
        27
      ],
      "statements": 15,
      "percentage": "93.3%"
    },
    "s.12a2004c5d590a2c62d462918790dbdae271fbfc605bdc6d4789922be701595d": {
      "line_hits": {
        "3": 1
      },
      "missed_lines": [],
      "statements": 1,
      "percentage": "100.0%"
    },
    "t.146c6706c5cde1bfac08e0c69a3f75073cf5c1be7039511547769766e4483b2b": {
      "line_hits": {
        "4": 1,
        "5": 1
      },
      "missed_lines": [],
      "statements": 2,
      "percentage": "100.0%"
    }
  }
```

______

For contributor use:


chasefleming:
I'm not sure if skipping them for everyone is the right solve here. Maybe it's better to have a flag to only run coverage on certain types of code, e.g. contracts? Can you tell me more about what you're trying to achieve with this? I don't see an issue for this.

m-Peter:
The main problem here is that scripts & transactions cannot be traced back to their respective source files. And quite often, certain scripts/transactions are even dynamic, they may be executed from `flow-go` or `flow-emulator`, without a file in the local filesystem. For example, running an integration test, will include in the coverage report 17+ transactions from the bootstrapping of the Flow Emulator, which are quite irrelevant to the integration tests, and they may skew the coverage metrics.However, you are right that we shouldn't exclude them universally, without a flag. In many cases, developers will also want code coverage for their scripts & transactions, which comprise a dApp, along with their contracts. But even with the flag, I have to find a way to give them a human-friendly identifier.

m-Peter:
@chasefleming Your point was quite valid, so I have addressed it here: https://github.com/onflow/flow-cli/pull/1074/commits/d29d846623ae479e40c2a4095fac992989980c68. Can you take another look? Thanks :pray:

internal/test/test_test.go
```diff
@@ -209,11 +209,95 @@ func TestExecutingTests(t *testing.T) {
...
+		})
+
+		// Execute script
+		script := tests.TestScriptWithCoverage
+		testFiles := map[string][]byte{
+			script.Filename: script.Source,
+		}
+		flags := flagsTests{
+			Cover:     true,
+			CoverCode: contractsCoverCode,
```
kozlovb, line 292: A question in the struct description there is '"default:"all" flag:"covercode"' for CoverCode string.
But const contractsCoverCode is equal to "contracts". So "covercode" is what a user has to specify and thens somewhere it is converted to contractsCoverCode value ?

m-Peter: The idea is that you add a default value to a flag, for the most used scenario. But developers can provide a different value to the flag. For example:

```bash
flow test --cover --covercode="contracts" tests/test_approval_voting.cdc
```

The `const contractsCoverCode` is added just as a best practice, to have magic strings defined in a single place, to avoid typos and having to update multiple places. So, if I want to update its value to something else, I can do it in a single place, in the `const` declaration.

m-Peter: Valid point about listing the available values for this flag :ok_hand: 
Updated in https://github.com/onflow/flow-cli/pull/1074/commits/3debe6f2a669c967c3a6e246b1cb4526a559d972

chasefleming:
@m-Peter is there a PR to add new flags to the docs repo here? https://github.com/onflow/docs

m-Peter:
@chasefleming Good point :+1: I haven't made one yet, because it's not yet included in a release, however I will get to it shortly.

chasefleming (approved):

m-Peter:
@chasefleming I have opened a [PR](https://github.com/onflow/docs/pull/160) for updating the relevant documentation of the `flow test` command and its supported flags. Can you please take a look, so we can get these PRs merged? :pray:

chasefleming:
@m-Peter what was the reason for closing this?

PR #1118, opened 2023-07-04
if results too large get each result one at a time

### LINKED ISSUE #1107: When fetching Transasctions and Results for a block that is very large it fails
### Instructions

Using flowkit try to GetTransactionsByBlockID() block 55179362

You will get the error message `getting transactions: getting transaction results: client: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (26319669 vs. 20971520)`

### Problem

Flowkit fetches all Transactions in a block (small size) and then TransactionResults (large size) in one got. 

### Steps to Reproduce

Using flowkit try to GetTransactionsByBlockID() block 55179362

### Acceptance Criteria

Guard against the error and manually run GetTransactionResults in a loop for every Transaction in the block.

bjartek:
Could not find a test for the original method so i did not add one here. 

Closes #1107

### DIFF
--- a/flowkit/flowkit.go
+++ b/flowkit/flowkit.go
@@ -871,6 +871,17 @@ func (f *Flowkit) GetTransactionsByBlockID(
 
 	txRes, err := f.gateway.GetTransactionResultsByBlockID(blockID)
 	if err != nil {
+		if strings.Contains(err.Error(), "received message larger than max") {
+			txRes := []*flow.TransactionResult{}
+			for _, transaction := range tx {
+				txr, err := f.gateway.GetTransactionResult(transaction.ID(), true)
+				if err != nil {
+					return nil, nil, err
+				}
+				txRes = append(txRes, txr)
+			}
+			return tx, txRes, nil
+		}
 		return nil, nil, err
 	}
 	return tx, txRes, nil


nvdtf (approved):

flowkit/flowkit.go
```diff
@@ -871,6 +871,17 @@ func (f *Flowkit) GetTransactionsByBlockID(
 
 	txRes, err := f.gateway.GetTransactionResultsByBlockID(blockID)
 	if err != nil {
```
janezpodhostnik, line 873: I would suggest inverting the condition to make the code easier to read.

if no error return else try fetching one by one.

janezpodhostnik (approved):
lgtm, just one minor comment

flowkit/flowkit.go
bjartek: Totally agree.

bjartek:
Question: will this not hit rate limit really fast? The blocks where this hits have hundreds of transactions and that will result in a lot of hits against the grpc endpoint.

janezpodhostnik:
The gateway should take care of handling rate limits I think. So basically outgoing calls would automatically be delayed if they are getting close to the rate limit. I did something like that here previously https://github.com/onflow/flow-batch-scan/blob/main/client/interceptors/uci_rate_limit.go

bjartek:
I just figured out that this will not work for some scenarios. Basically the only way atm to get SystemChunkTransactionResults is through the method that hits the limits.

bjartek:
Block id 55179362 is a block that has systemChunkTransactions and it is not possible to get all transactionResults for that block since it is too large, how do i then get the systemChunkEvents?

bjartek:
Ammended the pr with using the new GetTransactionByHeight that i exposed in another PR.

flowkit/go.mod
```diff
@@ -180,3 +180,5 @@ require (
 	modernc.org/memory v1.5.0 // indirect
 	modernc.org/sqlite v1.21.1 // indirect
 )
+
+replace github.com/onflow/flow-go-sdk => github.com/findonflow/flow-go-sdk v0.40.1-0.20230717122814-47abbbe6e6b0
```
sideninja, line 184: why is this needed?

flowkit/flowkit.go
```diff
@@ -864,16 +864,29 @@ func (f *Flowkit) GetTransactionsByBlockID(
...
-
 	txRes, err := f.gateway.GetTransactionResultsByBlockID(blockID)
-	if err != nil {
-		return nil, nil, err
+	if err == nil {
+		return tx, txRes, nil
+	}
+	errorMessage := err.Error()
+
+	if strings.Contains(errorMessage, "received message larger than max") || strings.Contains(errorMessage, "trying to send message larger than max") {
```
sideninja, line 878: do we have access to these errors? so we could import them by type instead of hardcoding error message?

sideninja (commented):
Good idea to handle this, left some comments

flowkit/go.mod
bjartek: because at this time the pr needed a new version of flow-go-sdk since it added a new method to it

bjartek: https://github.com/onflow/flow-go-sdk/pull/427 this pr is still not merged

flowkit/flowkit.go
bjartek: they look like standard GRPC error messages to me, so i doubt it.

go.mod
```diff
@@ -230,5 +228,3 @@ require (
 	modernc.org/sqlite v1.21.1 // indirect
 	nhooyr.io/websocket v1.8.6 // indirect
 )
-
-replace github.com/onflow/flow-cli/flowkit => ./flowkit
```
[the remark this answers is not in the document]
bjartek: will add it back again

flowkit/go.mod
sideninja: merged the PR and released a new version https://github.com/onflow/flow-go-sdk/releases/tag/v0.42.0

bjartek: fixed

bjartek:
I pushed changes but they do not appear here

bjartek:
We have decided to not do this, it should be done on another level.

PR #1125, opened 2023-07-13
Execute FLIX scripts/transactions

### LINKED ISSUE #984: Execute FLIX transactions/scripts
We like to be able to run FLIX interactions from the Flow CLI directly. 

### Approach

Let's consider the normal transaction send flow:

```
flow transactions send <path> ...
```

We can extend `<path>` into:
* Can be a filesystem `path`:
  * If a Cadence file is passed, the current behaviour is expected.
  * If a FLIX interaction is passed locally, Flow CLI will parse and run it.
* Can be a `URL`: 
  * If a Cadence file is fetched, Flow CLI will run it.
  * If a FLIX interaction is fetched, Flow CLI will parse and run it.
* Can be `flix:<name>`:
  * The Flow FLIX server is checked for named interactions with the passed `<name>`

If a FLIX interaction is passed, Flow CLI will first show a Yes/No confirmation screen with interaction template's:
* Title
* Description
* Audits

### References
https://github.com/onflow/flips/blob/main/flips/20220503-interaction-templates.md
https://github.com/bjartek/overflow/blob/flix2/flix.go

chasefleming:
Closes #984  

## Description

TODO:
- [ ] Execute FLIX scripts by name
- [ ] Execute FLIX scripts by ID
- [ ] Execute local FLIX scripts
- [ ] Execute FLIX transactions by name
- [ ] Execute FLIX transactions by ID
- [ ] Execute local FLIX transactions
______

For contributor use:


bjartek:
I think using cobra and commands and flags would make this api better. It is also a lot easier to generate good help that way. flow-cli is modeled that way, any reason to change that?

chasefleming:
@bjartek fair point. I'm still getting familiar with Cobra. Is it possible to handle commands like the `flix:id:123` or `flix:name:transfer-flow` in a command like `flow transactions send flix:name:transfer-flow ...` or `flow scripts execute flix:id:123 ...`?

chasefleming:
@bjartek I like the simplicity of the third option. The others are a bit verbose. I'd be for switching to the third, `flow flix transfer-flow`, since -- as you mentioned -- you can identify if it's a script or tx from the flix field.

chasefleming:
The first too are a little verbose. I like the simplicity of option three, `flow flix transfer-flow` since -- as you mentioned -- you can determine if it's a script or tx from the flix type field. Good suggestions. Thanks. I'll implement it.

chasefleming:
@bjartek one more note: yeah we'll actually need to probably add `id`, `name`, or `path` in there as well -- like `flow flix name transfer-flow` -- otherwise it won't know how to query flix...unless you have some really reliable way to identify an ID and then we could do it programmatically.

sideninja (commented):
I think defining this command to use flags would be better. Example: `flow flix --name hello` or `flow flix --id 123`. The usage of `flow flix id 123` or `flow flix name hello` is nonconsistent with other commands. CLI never uses arguments to define the type of the next argument value. Also https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#arguments

internal/flix/flix.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+var flags = Flags{}
+
+var Cmd = &cobra.Command{
+	Use:   "flix",
+	Short: "Commands for the Flix functionality",
+}
+
+var idCommand = &command.Command{
```
sideninja, line 39: by using flags you could merge all three functions into one. And then just handle based on the flag present.

internal/flix/flix.go
```diff
@@ -0,0 +1,181 @@
...
+	Cmd: &cobra.Command{
+		Use:     "id <id>",
+		Short:   "Execute flix operation with given id",
+		Args:    cobra.MinimumNArgs(1),
+		Example: `flow flix id 123`,
+	},
+	Flags: &flags,
+	RunS: func(args []string, _ command.GlobalFlags, _ output.Logger, flow flowkit.Services, state *flowkit.State) (result command.Result, err error) {
```
sideninja, line 47: It's good idea to extract the function out, so we can later test it, example: https://github.com/onflow/flow-cli/blob/master/internal/blocks/get.go#L50

bjartek:
I like the new command syntax.

bjartek:
What should happen if you try to execute a flix that do not have updated pins? So say your flix relies on an old version of a contract pinned and you try to run it after it has been updated. 

should there be a prompt that tells the user "hey this flix is outdated" or "this flix is not verified"

The same can be said if you want your flix to be verified by a validator.

internal/flix/flix.go
chasefleming: Addressed

internal/flix/flix.go
chasefleming: Addressed

chasefleming:
UPDATE: I wound up detecting if it's a path, name, or id and removed requirement to provided it as a command or as a flag (cc @sideninja )

Also, @bjartek I really like the prompt idea for the out of sync pins. Do you think that's fine to do in a separate issue? Doesn't seem required for this PR, right? Would be nice to close this one and start that separately since it would also require us adding the pinning logic to flixkit. Is Overflow's pin compute functions working fully or it requires flix to be updated to the new version that is more reliable for hashing with the array format?

bjartek:
Lets get this merged and we can sync on pins. Flowdiver has the entire contrsct tree on flow so we can compute pins therr and expose.

bjartek (approved):

cmd/flow/main.go
```diff
@@ -111,6 +116,10 @@ func main() {
 		ID:    "security",
 		Title: "🔒 Flow Security",
 	})
+	cmd.AddGroup(&cobra.Group{
```
sideninja, line 119: I would add them under super commands.

go.mod
```diff
@@ -232,3 +233,4 @@ require (
 )
 
 replace github.com/onflow/flow-cli/flowkit => ./flowkit
+replace github.com/onflow/flixkit-go => ../flixkit-go
```
sideninja, line 236: remove this

internal/flix/flix.go
```diff
@@ -0,0 +1,142 @@
...
+	Exclude     []string `default:"" flag:"exclude" info:"Fields to exclude from the output (events)"`
+	GasLimit    uint64   `default:"1000" flag:"gas-limit" info:"transaction gas limit"`
+}
+
+var flags = flixFlags{}
+
+var FlixCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "flix <id | name | path>",
+		Short:   "Execute flix with given id, name or path",
```
sideninja, line 40:
```suggestion
		Short:   "Execute FLIX template with a given id, name, or local filename",
```

internal/flix/flix.go
```diff
@@ -0,0 +1,142 @@
...
+		Example: "flow flix multiply 2 3 --network testnet",
```
sideninja, line 41: NIT If we don't need to provide a network let's not so it's not complicating example.
```suggestion
		Example: "flow flix multiply 2 3",
```

internal/flix/flix.go
```diff
@@ -0,0 +1,142 @@
...
+
+type flixQueryTypes string
+
+const (
+	name flixQueryTypes = "name"
+	path flixQueryTypes = "path"
+	id   flixQueryTypes = "id"
+)
+
+func isHex(s string) bool {
```
sideninja, line 57: You can use: `_, err := hex.DecodeString(str)` if no error then all good.

internal/flix/flix.go
```diff
@@ -0,0 +1,142 @@
...
+	match, _ := regexp.MatchString("^[a-fA-F0-9]+$", s)
+	return match
+}
+
+func isPath(path string) bool {
+	return strings.HasPrefix(path, "./")
```
sideninja, line 63: I would just check if file exists, you can use `_, err := os.Stat(path)`

internal/flix/flix.go
```diff
@@ -0,0 +1,142 @@
...
+	flow flowkit.Services,
+	state *flowkit.State,
+) (result command.Result, err error) {
+	flixService := flixkit.NewFlixService(&flixkit.Config{})
+	ctx := context.Background()
+	var template *flixkit.FlowInteractionTemplate
+	flixQuery := args[0]
+	flixQueryType := getType(flixQuery)
+
+	if flixQueryType == id {
```
sideninja, line 90: better to use switch type:
```
switch getType(flixQuery) {
    case id: 
}
```

sideninja: and make default case to return an error because it couldn't match a type

internal/flix/flix.go
sideninja: But also assert length of the string to be fixed to whatever the ID length should be, because otherwise some names would be valid IDs

internal/scripts/execute.go
```diff
@@ -32,13 +32,13 @@ import (
 	"github.com/onflow/flow-cli/internal/command"
 )
 
-type flagsScripts struct {
```
sideninja, line 35: why this changed?

internal/scripts/execute.go
```diff
@@ -100,3 +98,12 @@ func execute(
 
 	return &scriptResult{value}, nil
 }
+
+func executeLocalScript(args []string, filename string, readerWriter flowkit.ReaderWriter, flow flowkit.Services) (command.Result, error) {
```
sideninja, line 102: do you need this function?

internal/transactions/send.go
```diff
@@ -64,8 +64,19 @@ func send(
 	flow flowkit.Services,
 	state *flowkit.State,
 ) (result command.Result, err error) {
-	codeFilename := args[0]
+	return executeLocalTransaction(args, args[0], flow, state)
+}
+
+func executeLocalTransaction(args []string, filename string, flow flowkit.Services, state *flowkit.State) (result command.Result, err error) {
```
sideninja, line 70: same Q as above

sideninja (changes requested):
Left comments, looks good but some things to address, also can you link a documentation PR to this one.

chasefleming:
@sideninja Here is the docs PR: https://github.com/onflow/docs/pull/182

internal/scripts/execute.go
chasefleming: Good catch. I had these separated before it was moved to the new commands and needs to be removed now.

internal/scripts/execute.go
chasefleming: These need to be exported to use in the other package

sideninja:
@chasefleming let me know when you want another review

internal/super/flix.go
```diff
@@ -0,0 +1,169 @@
...
+
+func isHex(str string) bool {
+	if len(str) != 64 {
+		return false
+	}
+	_, err := hex.DecodeString(str)
+	if err != nil {
+		return false
+	}
+	return true
```
sideninja, line 82:
```suggestion
	return err == nil
```

internal/super/flix.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func isPath(path string) bool {
+	_, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+	return true
```
sideninja, line 90: same as above

cmd/flow/main.go
```diff
@@ -111,6 +112,10 @@ func main() {
...
```
sideninja, line 115: not sure if you now still need this group if it's added under super commands.

sideninja (approved):
Looks good, left couple of small comments

Issue #1151, opened 2023-08-02
Improve events display


sideninja:
We could improve the display of events in the CLI to separate the Flow fee events at the bottom and maybe also visually separate with a divider. 
Currently, any transaction events are stuffed with a lot of fee related events which can make things confusing. Mostly fee realted events are not relevant when you are looking for events and could be hidden by default or separated. 

Here's an example of current events section for a random tx:
```

Events:		 
    Index	0
    Type	A.0ae53cb6e3f42a79.FlowToken.TokensWithdrawn
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
		- amount (UFix64): 0.00100000 
		- from (Address?): 0x01cf0e2f2f715450 

    Index	1
    Type	A.0ae53cb6e3f42a79.FlowToken.TokensWithdrawn
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
		- amount (UFix64): 0.00100000 
		- from (Never?): nil 

    Index	2
    Type	A.0ae53cb6e3f42a79.FlowToken.TokensDeposited
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
		- amount (UFix64): 0.00000000 
		- to (Address?): 0xe5a8b7f23e8b548f 

    Index	3
    Type	A.e5a8b7f23e8b548f.FlowFees.TokensDeposited
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
		- amount (UFix64): 0.00000000 

    Index	4
    Type	A.0ae53cb6e3f42a79.FlowToken.TokensDeposited
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
		- amount (UFix64): 0.00100000 
		- to (Address?): 0x045a1763c93006ca 

    Index	5
    Type	flow.AccountCreated
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
		- address (Address): 0x045a1763c93006ca 

    Index	6
    Type	flow.AccountKeyAdded
    Tx ID	db93f1da001a8fbc65d1e287bef140a9afcd564de802a3e6827feaa15077691e
    Values
[...]

PR #1188, opened 2023-09-06
Filter fee events

bthaile:
Closes: https://github.com/onflow/flow-cli/issues/1151

## Description
 - filter out "FeeEvent" and "FlowToken" events by default.
 - Add new included `fee-events` flag so users can get all events including the fee events.
 - Added two new tests to test filtering events
 - Updated transactions get command include flags help
______

For contributor use:


bjartek:
Will this not remove all FlowToken events?
The fee events are always the 3 latest in a tx. Just remove them?
So the withdraw/deposit/fee
I think fee and effort should be returned in TransactionResult personally. Those 3 events are different the the others cause you did not emit them. The system did.

bthaile:
> Will this not remove all FlowToken events? The fee events are always the 3 latest in a tx. Just remove them? So the withdraw/deposit/fee I think fee and effort should be returned in TransactionResult personally. Those 3 events are different the the others cause you did not emit them. The system did.

 I was going off the github issue linked. Spot checking some transactions, I see what you are saying the last three events are what need to be filtered out. I'll update the PR. 
 Thanks

bjartek (approved):

internal/transactions/transactions.go
```diff
@@ -153,6 +153,13 @@ func (r *transactionResult) String() string {
 			Events: r.result.Events,
 		}
 
+		if r.result != nil && !command.ContainsFlag(r.include, "fee-events") {
+			// last 3 events are fee events
+			if e.Events != nil && len(e.Events) >= 3 {
```
janezpodhostnik, line 158: Can you put this 3 in a constant. It will be easier to find if we change the amount of events the fee deduction emits.

janezpodhostnik (changes requested):
In certain setups (fees are set to 0) there will be no fee events. I would make the check more complex:

If a FeesDeducted event is present (should be the last one) then the last 3 events are fee events.

internal/transactions/transactions.go
bthaile: replace magic number with constant.

internal/transactions/transactions.go
```diff
@@ -153,6 +156,22 @@ func (r *transactionResult) String() string {
...
+					if strings.Contains(event.Type, feeDeductedEvent) {
+						found = true
+						break
+					}
+				}
+				if found {
+					e.Events = e.Events[:len(e.Events)-feeEventsCountAppended]
+				}
+			}
+		}
```
janezpodhostnik, line 173:
```suggestion
		if r.result != nil && e.Events != nil  && !command.ContainsFlag(r.include, "fee-events") {
			for _, event := range e.Events {
				if strings.Contains(event.Type, feeDeductedEvent) {
					// if fee event are present remove them
					e.Events = e.Events[:len(e.Events)-feeEventsCountAppended]
					break
				}
			}
		}
```

janezpodhostnik (approved):

chasefleming:
@bthaile Can we make sure we update the docs wherever necessary with this new flag?

bthaile:
> @bthaile Can we make sure we update the docs wherever necessary with this new flag?

Updated here, internal/transactions/get.go in the PR. Are their other docs? 

good call, I can update developer docs, here https://developers.flow.com/tools/flow-cli/transactions/get-transactions

chasefleming (approved):

chasefleming:
Thanks @bthaile !

PR #1190, opened 2023-09-08
Windows path fix

### LINKED ISSUE #1183: `flow project deploy` broken on windows
When I deploy my contracts in flow.json on linux/mac, it works totally fine. But windows seems to be broken.

image
Something to do with not reading the imports in contracts correctly. Maybe due to wrongful file paths

jribbink:
Closes #1183 

## Description

This fixes bugs related to Windows path formatting.  We were using the [`"path"`](https://pkg.go.dev/path) package in a few places that should have been using[ `"path/filepath"`](https://pkg.go.dev/path/filepath).

Some issues with this were more theoretical and didn't seem to surface yet (absolute paths like `C:/foo/bar` would not be recognized as absolute).  But most critically the mixing of `"path"` and `"path/filepath"` meant that when doing comparisons/dictionary lookups to see if paths were equivalent the different format would result in two paths not matching (this is what broke deployment).

The PR also ensures that all paths written to `flow.json` are in UNIX format for cross-platform compatibility.  This is the pattern adopted by NPM in the `package.json` file.

______

For contributor use:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,22 @@
+# Set OS-related environment variables
+ifeq ($(OS),Windows_NT)
+	SHELL := cmd.exe
+	GOPATH ?= $(USERPROFILE)\go
+	PATH := $(PATH);$(GOPATH)\bin
+	MKDIR_GOPATH := if not exist "$(GOPATH)" md "$(GOPATH)"
+	SET_GOPATH := go env -w GOPATH="$(GOPATH)"
+	SET_GO111MODULE := go env -w GO111MODULE=on
+	SET_CGO_DISABLED := go env -w CGO_ENABLED=0
+else
+	SHELL := /bin/bash
+	GOPATH ?= $(HOME)/go
+	PATH := $(PATH):$(GOPATH)/bin
+	MKDIR_GOPATH := mkdir -p "$(GOPATH)"
+	SET_GOPATH := export GOPATH="$(GOPATH)"
+	SET_GO111MODULE := export GO111MODULE=on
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -24,17 +35,20 @@ binary: $(BINARY)
 
 .PHONY: install-tools
 install-tools:
-	cd ${GOPATH}; \
```
sideninja, line 27: can we please extract this as a separate PR

flowkit/state_test.go
```diff
@@ -343,7 +344,7 @@ func generateAliasesComplexProject() State {
```
sideninja, line 344: could we add some windows style paths to test

internal/super/files_test.go
```diff
@@ -20,19 +20,19 @@ package super
...
 )
 
 func Test_AccountFromPath(t *testing.T) {
 	paths := [][]string{ // first is path, second is account name
-		{"cadence/contracts/alice/foo.cdc", "alice"},
-		{"cadence/contracts/alice", "alice"},
-		{"cadence/contracts/alice/boo/foo.cdc", ""},
-		{"cadence/contracts/foo.cdc", ""},
-		{"cadence/contracts/foo/bar/goo/foo", ""},
+		{filepath.FromSlash("cadence/contracts/alice/foo.cdc"), "alice"},
```
sideninja, line 31: you can just call this function inside the for loop once

chasefleming (approved):

PR #1191, opened 2023-09-08
Add Windows tests

jribbink:
Runs the test suite in Windows as well as Ubuntu.  Hopefully will catch Windows-specific bugs like #1183

### DIFF
--- a/flowkit/cmd/flow-schema/flow-schema.go
+++ b/flowkit/cmd/flow-schema/flow-schema.go
@@ -23,6 +23,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


flowkit/cmd/flow-schema/flow-schema.go
```diff
@@ -54,11 +55,15 @@ func main() {
 			os.Exit(1)
 		}
 
-		if string(fileContents) != string(json) {
+		if normalizeNewlines(string(fileContents)) != normalizeNewlines(string(json)) {
```
jribbink, line 58: JSON schema generator checks for changes, we need to normalize CRLF to LF otherwise Windows machine will show changes compared to UNIX-generated schema.

chasefleming (approved):

PR #1222, opened 2023-09-29
add support for js code generation for flix

bthaile:
Depends on flixkit PR  https://github.com/onflow/flixkit-go/pull/12

## Description

Add support for generating binding files based on flix. 
new flags 
`--generate js` and `--save <filename>`

image

______

For contributor use:

### DIFF
--- a/internal/super/flix.go
+++ b/internal/super/flix.go
@@ -26,12 +26,11 @@ import (
 
 	"github.com/onflow/flixkit-go"
 
+	"github.com/onflow/flow-cli/flowkit"
 	"github.com/onflow/flow-cli/flowkit/output"
 	"github.com/onflow/flow-cli/internal/command"
 	"github.com/onflow/flow-cli/internal/scripts"
 	"github.com/onflow/flow-cli/internal/transactions"
-
-	"github.com/onflow/flow-cli/flowkit"
 
 	"github.com/spf13/cobra"
 )
@@ -47,7 +46,14 @@ type flixFlags struct {
 	Include     []string `default:"" flag:"include" info:"Fields to include in the output"`
 	Exclude     []string `default:"" flag:"exclude" info:"Fields to exclude from the output (events)"`
 	GasLimit    uint64   `default:"1000" flag:"gas-limit" info:"transaction gas limit"`
+	Generate 	string	 `default:"" flag:"generate" info:"generate code for the given language"`
+	Save 		string	 `default:"" flag:"output" info:"output file for generated code"`
 }
+
+type flixResult struct {
+	output string
+	save string
+} 
 
 var flags = flixFlags{}
 
@@ -106,6 +112,7 @@ func execute(
 	ctx := context.Background()
 	var template *flixkit.FlowInteractionTemplate
 	flixQuery := args[0]
+	isLocal := false
 
 	switch getType(flixQuery) {
 	case flixId:
@@ -121,6 +128,7 @@ func execute(
 		}
 
 	case flixPath:
+		isLocal = true
 		file, err := os.ReadFile(flixQuery)
 		if err != nil {
 			return nil, fmt.Errorf("could not read flix file %s: %w", flixQuery, err)
@@ -133,6 +141,22 @@ func execute(
 	default:
 		return nil, fmt.Errorf("invalid flix query type: %s", flixQuery)
 	}
+
+	if flags.Generate != "" {
+		jsGen := flixkit.JavascriptGenerator{}
+		out, err := jsGen.Generate(template, flixQuery, isLocal)
+
+		if flags.Save != "" {
+			err = os.WriteFile(flags.Save, []byte(out), 0644)
+			if err != nil {
+				return nil, fmt.Errorf("could not write to file %s: %w", flags.Save, err)
+			}
+		}
+		return &flixResult{
+			save: flags.Save,
+			output: out,
+		}, err
+	} 
 
 	cadenceWithImportsReplaced, err := template.GetAndReplaceCadenceImports(flow.Network().Name)
 	if err != nil {
@@ -160,4 +184,26 @@ func execute(
 		GasLimit:    flags.GasLimit,
 	}
 	return transactions.SendTransaction([]byte(cadenceWithImportsReplaced), args[1:], "", flow, state, transactionFlags)
+}
+
+func (fr *flixResult) JSON() any {
+	result := make(map[string]any)
+	result["output"] = fr.output
+	result["save"] = fr.save
+	return result
+}
+
+func (fr *flixResult) String() string {
+	return isSaved(fr)
+}
+
+func (fr *flixResult) Oneliner() string {
+	return isSaved(fr)
+}
+
+func isSaved(fr *flixResult) string {
+	if fr.save != "" {
+		return fmt.Sprintf("Generated code saved to %s", fr.save)
+	}
+	return fr.output
 }


internal/super/flix.go
```diff
@@ -161,3 +185,25 @@ func execute(
...
+	return isSaved(fr)
+}
+
+func (fr *flixResult) Oneliner() string {
+	return isSaved(fr)
+}
+
+func isSaved(fr *flixResult) string {
+	if fr.save != "" {
+		return fmt.Sprintf("Generated code saved to %s", fr.save)
```
chasefleming, line 206: Use the logger instead of fmt

chasefleming (changes requested):
I think the generate flag type needs to be `fcl`, not `js` since this is specifically for fcl and there could be other js libs.

Can you also link to a larger issue (probably on flixkit) which shows what the expected output is and describes the use case please?

chasefleming:
Can you also create a CLI issue as well please? Thanks.

internal/super/flix.go
bthaile: the isSaved method returns a string. It's a convenience method for `String()` and `Oneliner()` methods. it's not a log out statement specifically.

bthaile:
> Can you also create a CLI issue as well please and link it? Thanks.

added https://github.com/onflow/flow-cli/issues/1232 and referenced it in the PR description

chasefleming (changes requested):
@bthaile I'm getting `unknown flag: --generate` when running `go run cmd/flow/main.go flix multiply.template.json --generate fcl-js --output multiply.template.js --network testnet`. 

I'm also wondering if commands like `generate` should be as flags `--generate`. Those are for things like args. `generate` feels like a command say `flow flix generate...` or `flow flix execute...`. Now that we are getting lots of flix items it feel like they should be subcommands otherwise you can run `flow flix --bindings --execute --generate` and what is it supposed to do?

internal/super/flix.go
```diff
@@ -47,6 +47,13 @@ type flixFlags struct {
 	Include     []string `default:"" flag:"include" info:"Fields to include in the output"`
 	Exclude     []string `default:"" flag:"exclude" info:"Fields to exclude from the output (events)"`
 	GasLimit    uint64   `default:"1000" flag:"gas-limit" info:"transaction gas limit"`
+	Bindings    string   `default:"" flag:"bindings" info:"generate binding code for the given language, fcl-js' is supported language"`
+	Save        string   `default:"" flag:"output" info:"output file for generated binding code"`
```
bjartek, line 51: there is a global save flag on all flow commands, can we not use that one here?

bjartek (changes requested):

internal/super/flix.go
bjartek: There is also a global --output flag and this would clash with that.

jribbink:
> I'm also wondering if commands like `generate` should be as flags `--generate`. Those are for things like args. `generate` feels like a command say `flow flix generate...` or `flow flix execute...`. Now that we are getting lots of flix items it feel like they should be subcommands otherwise you can run `flow flix --bindings --execute --generate` and what is it supposed to do?

+1 on this point.  In my opinion, these are distinct commands without significant functionality overlap, so they should probably be namespaced accordingly instead of distinguishing using flags.

internal/super/flix.go
```diff
@@ -47,6 +47,13 @@ type flixFlags struct {
 	Include     []string `default:"" flag:"include" info:"Fields to include in the output"`
 	Exclude     []string `default:"" flag:"exclude" info:"Fields to exclude from the output (events)"`
 	GasLimit    uint64   `default:"1000" flag:"gas-limit" info:"transaction gas limit"`
+	Bindings    string   `default:"" flag:"bindings" info:"generate binding code for the given language, fcl-js' is supported language"`
```
jribbink, line 50:
```suggestion
	Bindings    string   `default:"" flag:"bindings" info:"generate binding code for the given language, fcl-js is supported language"`
```

bjartek:
> @bthaile I'm getting `unknown flag: --generate` when running `go run cmd/flow/main.go flix multiply.template.json --generate fcl-js --output multiply.template.js --network testnet`.
> 
> I'm also wondering if commands like `generate` should be as flags `--generate`. Those are for things like args. `generate` feels like a command say `flow flix generate...` or `flow flix execute...`. Now that we are getting lots of flix items it feel like they should be subcommands otherwise you can run `flow flix --bindings --execute --generate` and what is it supposed to do?

generate should be a subcommand of flix, not a flag.

internal/super/flix.go
```diff
@@ -121,6 +129,7 @@ func execute(
 		}
 
 	case flixPath:
+		isLocal = true
 		file, err := os.ReadFile(flixQuery)
```
bjartek, line 133: use state.ReadFile here since it uses the underlying ReaderWrite. 

(yes i know this line is not in this PR but it should be used)

bjartek (changes requested):

internal/super/flix.go
bjartek, line 133:
```suggestion
		file, err := state.ReadFile(flixQuery)
```

bjartek (changes requested):

internal/super/flix.go
bjartek: most other places in flow-cli where file reading/writing is done it is done on the ReaderWriter.

internal/super/flix.go
bthaile: good call, I'll update to use the flag `--save` also pushed a change to reorganize flix verbs. I'll update PR description.

bthaile:
> @bthaile I'm getting `unknown flag: --generate` when running `go run cmd/flow/main.go flix multiply.template.json --generate fcl-js --output multiply.template.js --network testnet`.
> 
> I'm also wondering if commands like `generate` should be as flags `--generate`. Those are for things like args. `generate` feels like a command say `flow flix generate...` or `flow flix execute...`. Now that we are getting lots of flix items it feel like they should be subcommands otherwise you can run `flow flix --bindings --execute --generate` and what is it supposed to do?

Updated description and PR.  It makes more sense to break out verbs than add flags.

internal/super/flix.go
bthaile: made changes to use state to read and write files. Also dried up the code a bit.

chasefleming:
Cool @bthaile can you make an issue to remember to update the docs with this new syntax? We'll also need to add the bindings stuff to the docs.

chasefleming:
@bthaile There isn't much to test here but as a best practice does it make sense to at least add a test file maybe mocking the library was called so that its in place for whoever touches the code next? Apologies I should have added that when I originally put in the execute code originally. I don't think it matters that much right now since most of the logic is in the flixkit library, but if you are feeling like it wouldn't hurt.

bthaile:
> @bthaile There isn't much to test here but as a best practice does it make sense to at least add a test file maybe mocking the library was called so that its in place for whoever touches the code next? Apologies I should have added that when I originally put in the execute code originally. I don't think it matters that much right now since most of the logic is in the flixkit library, but if you are feeling like it wouldn't hurt.

I don't think it's needed at this time. I could add tests when adding flix generate.

chasefleming:
Sounds good fine with that. Is this ready for me to manually test it again?

internal/super/flix.go
```diff
@@ -49,18 +49,44 @@ type flixFlags struct {
...
+		Args:    cobra.MinimumNArgs(1),
 	},
 	Flags: &flags,
-	RunS:  execute,
+	RunS:  executeCmd,
+}
+
+var bindingCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "bindings <id | name | path>",
```
bjartek, line 78: bindings is not a term i like here, but not sure if generate is a lot better and generate-client is very long.

chasefleming: Yeah I don't love`bindings` either here tbh

chasefleming:
@bthaile I ran `go run cmd/flow/main.go flix bindings multiply.template.json` and I'm getting this error:

`❌ Command Error: could not find flix with name bindings: unexpected end of JSON input`

chasefleming (commented):
Screenshot 2023-10-26 at 1 11 34 PM

I think we need a little more text and explanation here. e.g. saved what to there?

chasefleming (approved):

bjartek:
One thing i was thinking about here is that the usecase we talked about in the developer office hours yesterday.

 - could a client be generated for more then 1 flix?
 - should a client have a name or other information. Like if we generate a go file it needs a package. If we generate jvm it needs a artifactId and a name

internal/super/flix.go
bthaile: what would be a better word? "integrate" or "bind" or "wrap" or "gen-wrap" or "gen-bind", "craft" "conjure", ...

bthaile: good suggestion "package" good suggestion @chasefleming

chasefleming:
True @bjartek would be nice if it could grab a whole directory

bthaile:
> One thing i was thinking about here is that the usecase we talked about in the developer office hours yesterday.
> 
> * could a client be generated for more then 1 flix?
> * should a client have a name or other information. Like if we generate a go file it needs a package. If we generate jvm it needs a artifactId and a name

very simple for a developer to write a script (bash, nodejs, ...) to generate package files per flix. no need to add that complexity to flow-cli. 
when we get to golang generation we can figure out module name to add, don't need for js, imo. and dev can call the filename what ever. I really want to keep this simple, more features isn't better. imo

Issue #1022, opened 2023-05-04
Progress loggers issues


sideninja:
Progress reporting should be avoided where possible (in some cases it provides value to CLI users seeing the progress). 

There's a possible bug not stopping the progress https://github.com/onflow/flow-cli/blob/93b7336efceb00bf20b3039ca3694cec492ebff9/flowkit/flowkit.go#L310 
All the progress reports should be revisited and make sure they can run nicely with multiple go routines. 

The logger should support integration with popular go logging libraries (logrus, zap, zerolog).

bjartek:
That makes sense to me.

PR #1249, opened 2023-10-26
fix: ensure that stopProgress is deferred on all code paths where Sta…

bjartek:
## Description

Does not fix all that @sideninja suggest in #1022 but it does ensure that the loggers are deferred stopped on all codepaths. 

The reason this is important is because if you use flowkit in a multithreaded enivronment and stopProgress is not called it will cause all sorts of havoc.

______

For contributor use:

### DIFF
--- a/flowkit/flowkit.go
+++ b/flowkit/flowkit.go
@@ -220,7 +220,6 @@ func (f *Flowkit) prepareTransaction(
 	tx *transactions.Transaction,
 	account *accounts.Account,
 ) (*transactions.Transaction, error) {
-
 	block, err := f.gateway.GetLatestBlock()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


bjartek:
One could argue that doing this in internal is not that important as that code is  not exposed in flowkit. 

I do not think StopProgress should be in flowkit at all, I would prefer for it to use slog or another semantic logging solution like zerolog og zap.

chasefleming:
Yeah I agree I'm not sure how important the setting of a custom logger is for developers. Although would love to hear from them if they find it important. I'd be for options from the standard library like slog. That ticket I think was made before slog was released so probably worth revisiting the thinking.

chasefleming (approved):

Issue #1248, opened 2023-10-25
Support Flixkit-go FLIX generation


bthaile:
Support flixkit-go generate for creating FLIX version 1.0.0 given cadence transaction, script or interface.

PR #1258, opened 2023-11-07
Add `generate` to flix comamnd 

bthaile:
Closes: https://github.com/onflow/flow-cli/issues/1248

## Description

```
flow flix
execute, generate, package

Usage:
  flow flix [command]

Available Commands:
  execute     execute FLIX template with a given id, name, or local filename
  generate    generate FLIX json template given local filename
  package     package file for FLIX template fcl-js is default
```

generate FLIX version 1.0.0 json
When version 1.1.0 is added it will be the default version and user will have to pass in a flag for version. If that is the direction we want to go in. 

DAFT: Depends on merge in flixkit-go PR, https://github.com/onflow/flixkit-go/pull/20

______

For contributor use:


internal/super/flix.go
```diff
@@ -234,9 +318,67 @@ func getTemplate(state *flowkit.State, flixService flixkit.FlixService, flixQuer
 		if err != nil {
 			return nil, fmt.Errorf("could not parse flix from file %s: %w", flixQuery, err)
 		}
+	case flixUrl:
+		resp, err := http.Get(flixQuery)
```
chasefleming, line 322: This logic already occurs in flixkit. Why are we duplicating it here? We should be using what flixkit already has

chasefleming (changes requested):

internal/super/flix.go
```diff
@@ -193,6 +227,56 @@ func packageCmd(
...
+		return nil, fmt.Errorf("no cadence code found")
+	}
+
+	code, err := state.ReadFile(cadenceFile)
+	if err != nil {
+		return nil, fmt.Errorf("could not read cadence file %s: %w", cadenceFile, err)
+	}
+
+	depContracts := GetDeployedContracts(state)
+	generator, err := flixkitv1_0_0.NewGenerator(depContracts, nil, logger)
```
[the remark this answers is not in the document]
bthaile: FLIX v1.1.0 uses placeholder syntax so the Cadence is normalized to string import then converted to placeholder. So generation can take both import syntax or even a mix.

internal/super/flix.go
bthaile: now that you mention it, I see that flixkit has `FetchFlixWithContext` which isn't part of the Flix server implementation. Sure I can use `FetchFlixWithContext` to reduce code duplication.

internal/super/flix.go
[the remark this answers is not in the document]
chasefleming: How does it get core contracts like `FungibleToken` if it is looking in the deployments section of `flow.json` with `DeploymentsContractByNetwork` and core contracts are already deployed and not listed there?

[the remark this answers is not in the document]
bthaile: user can  pass them in as depContracts or use the default core contracts auto populated in `NewGenerator` constructor.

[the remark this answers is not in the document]
chasefleming: Oh I see it now, interesting: https://github.com/onflow/flixkit-go/blob/0e81dc829b9c8610ba6ddc6b4df55cdd84b2ece5/generator/core_contracts.go#L7

There's quite a few missing. Might want to make a ticket to add them on flixkit. But I see the auto-population now.

internal/super/flix.go
```diff
@@ -234,9 +316,56 @@ func getTemplate(state *flowkit.State, flixService flixkit.FlixService, flixQuer
...
+
+	// If the file is in the same directory and doesn't start with "./", prepend it.
+	if !filepath.IsAbs(relPath) && relPath[0] != '.' {
+		relPath = "./" + relPath
+	}
+
+	return relPath, nil
+}
+
+func GetDeployedContracts(state *flowkit.State) []flixkit.Contracts {
```
jribbink, line 350: This function will only get contracts in the `deployments` section of flow.json.  Do you need aliased contracts as well?

i.e.
```
{
"contracts": {
    "Flowvatar": {
        aliases: {
            testnet: "0x1234"
        }
    }
}
```

internal/super/flix.go
```diff
@@ -234,9 +316,56 @@ func getTemplate(state *flowkit.State, flixService flixkit.FlixService, flixQuer
...
 
 	default:
 		return nil, fmt.Errorf("invalid flix query type: %s", flixQuery)
 	}
 	return template, nil
 }
+
+// GetRelativePath computes the relative path from target to source.
+func GetRelativePath(source, target string) (string, error) {
+	relPath, err := filepath.Rel(filepath.Dir(target), source)
```
jribbink, line 337: this is backwards https://pkg.go.dev/path/filepath#Rel.

internal/super/flix.go
```diff
@@ -177,14 +203,20 @@ func packageCmd(
 	if err != nil {
 		return nil, err
 	}
-
 	isLocal := false
 	if getType(flixQuery) == flixPath {
 		isLocal = true
+		if flags.Save != "" {
+			// resolve template file location to relative path to be used by binding file
+			flixQuery, err = GetRelativePath(flixQuery, flags.Save)
```
jribbink, line 211: should be swapped also

jribbink, line 211: Also, this function is using `path/filepath` so all outputted paths will be in the format of the OS.

That means, on windows paths will be backslash separated.  At some point, `filepath.ToSlash` should be called on this path so that it is normalized for the OS.  Since it tends to be best to apply this transformation right before importing/exporting a path I'd say it probably belongs in flixkit itself here https://github.com/onflow/flixkit-go/blob/0e81dc829b9c8610ba6ddc6b4df55cdd84b2ece5/bindings/fcl-js.go#L47-L68

jribbink (changes requested):
few comments

internal/super/flix.go
bthaile: noice, I'll update. thanks

bthaile: It's correct. I'll update the method params so it's clear what is going on.

internal/super/flix.go
bthaile: it's correct, I updated the method comment and params to be more clear.

internal/super/flix.go
bthaile: I was only thinking of user's deployed contracts. Good call I can add contract aliases.

internal/super/flix.go
bthaile: filepath.Rel ends with calling `Clean`, which I didn't know. so the slashes will be correct depending on the OS the user is using. I see your point about flixkit being relative path aware, that would end up pushing IO logic into flixkit. I wanted it to only create the binding file and not have a IO dependency (interacting with the file system). 

https://pkg.go.dev/path/filepath#Clean
```
On Windows, Clean does not modify the volume name other than to replace occurrences of "/" with `\`. For example, Clean("//host/share/../x") returns `\\host\share\x`.
```

Great point! I don't think any changes are needed.

jribbink: I don't think there's any IO in the path module, just string operations.  The point I'm trying to make is that we need to make sure the path that ends up in the bindings is in Unix format.

i.e.
```
import "some\template\here.json"
```

Won't be valid javascript.  But beyond that even if the backslashes were escaped properly it templates generated by Windows OS wouldn't be compatible with Unix systems when they share it with other developers using Unix-based OS.  If it's normalized to Unix format it's supported by Windows & Unix, but not vice versa

internal/super/flix.go
bthaile: You know what the flow-cli command is to add an alias?

internal/super/flix.go
bthaile: gotcha, that is a good point. I'll make that change

bthaile: Updated PR

internal/super/flix.go
bthaile: Update PR to add alias contracts.

internal/super/flix.go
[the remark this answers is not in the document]
bthaile: I'll add these in v1.1 implementation.

internal/super/flix.go
```diff
@@ -84,9 +90,21 @@ var packageCommand = &command.Command{
 	RunS:  packageCmd,
 }
 
+var generateCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "generate cadence.cdc",
```
chasefleming, line 95: Should this be in `<>`?

chasefleming (approved):

internal/super/flix.go
bthaile: Yes, good one, updated

jribbink (approved):

PR #1285, opened 2023-11-29
Allow running individual test by name

### LINKED ISSUE #1273: Allow specifying individual specific tests to run with `flow test`
### Issue To Be Solved

It would be useful to have the ability to specify an individual test to run when using the `flow test` command.

I.e. for
```cadence
access(all) fun testFoobar() {
...
}

access(all) fun testBarFoo() {
...
}
```

It would be nice to be able to just run one of these tests instead of all of them.

Add a flag like `--run` to `flow test`.  From here a user could specify a regular expression to match against test functions & filter the tests which they wish to run.

### (Optional): Context

This is important for any IDE/external tooling supporting of the Cadence test framework.  Specifically, to integrate with the VSCode Testing API, it's important to have the ability to target individual test runs beyond a file/folder level.

jribbink:
Closes #1273 

## Description

Adds `--name` flag to run a specific test by name

______

For contributor use:

- x] Code follows the [standards mentioned here](https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#styleguides)

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -65,6 +65,7 @@ type flagsTests struct {
 	CoverCode    string `default:"all" flag:"covercode" info:"Use the covercode flag to calculate coverage report only for certain types of code. Available values are \"all\" & \"contracts\""`
 	Random       bool   `default:"false" flag:"random" info:"Use the random flag to execute test cases randomly"`
 	Seed         int64  `default:"0" flag:"seed" info:"Use the seed flag to manipulate random execution of test cases"`
+	Name         string `default:"" flag:"name" info:"Use the name flag to run only tests that match the given name"`
 }
 
 var testFlags = flagsTests{}
@@ -191,13 +192,22 @@ func testCode(
 			WithFileResolver(fileResolver(scriptPath, state)).
 			WithContracts(contracts)
 
-		results, err := runner.RunTests(string(code))
-		if err != nil {
-			return nil, err
+		if flags.Name != "" {
+			result, err := runner.RunTest(string(code), flags.Name)
+			if err != nil {
+				return nil, err
+			}
+			testResults[scriptPath] = []cdcTests.Result{*result}
+		} else {
+			results, err := runner.RunTests(string(code))
+			if err != nil {
+				return nil, err
+			}
+			testResults[scriptPath] = results
 		}
-		testResults[scriptPath] = results
 
-		for _, result := range results {
+
+		for _, result := range testResults[scriptPath] {
 			if result.Error != nil {
 				status = 1
 				break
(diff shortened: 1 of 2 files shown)


jribbink:
This is blocked by https://github.com/onflow/cadence-tools/issues/252

chasefleming (approved):

jribbink:
@m-Peter these changes look good on your end?

internal/test/test.go
```diff
@@ -191,13 +192,35 @@ func testCode(
...
+			}
+			includesTest := false
+			for _, testFunction := range testFunctions {
+				if testFunction == flags.Name {
+					includesTest = true
+					break
+				}
+			}
+
+			if includesTest {
```
m-Peter, line 208: I believe we can get rid of this `if includesTest` condition, by simply moving its body in the `if testFunction == flags.Name`.

m-Peter: You could also reverse the `if testFunction == flags.Name`, e.g.:
```go
if testFunction != flags.Name {
	continue
}

result, err := runner.RunTest(string(code), flags.Name)
if err != nil {
	return nil, err
}
testResults[scriptPath] = []cdcTests.Result{*result}
```

m-Peter:
@jribbink Nice work :clap: 
Just a minor comment on the code itself, and a couple more for the test cases.

internal/test/test.go
jribbink: This was a workaround for running `flow test` across multiple files.  e.g. `flow test test/**/*.cdc`.

If we don't check for the existence of the test, each file contained in the test run will receive an error from the `runner.RunTest` function, so the idea was to only run the test if it exists

i.e.

{
    "test/foo.cdc": {
        "testFoo": "Error: Test not found......"
    },
    "test/bar.cdc": {
        "testFoo": "Error: Test not found......"
    },
    "test/baz.cdc": {
        "testFoo": "PASS"
    }
}

m-Peter: I understand that :sweat_smile: I am not suggesting to alter the logic of it, it was merely a suggestion to eliminate the `bool` variable.

internal/test/test_test.go
```diff
@@ -585,4 +585,81 @@ Seed: 1521
...
+			script.Filename: script.Source,
+		}
+		flags := flagsTests{
+			Name: "doesNotExist",
+		}
+
+		result, err := testCode(testFiles, state, flags)
+
+		assert.NoError(t, err)
+		assert.Len(t, result.Results, 0)
```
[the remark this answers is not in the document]
jribbink: Yeah, I can add some additional messaging that the test isn't found 👍

internal/test/test.go
jribbink: Ah sorry, read that too quickly, I misunderstood :).  Yup I can do that for sure 👍

m-Peter (approved):
This is quite a neat idea :clap: 
Well done :100:

PR #1294, opened 2023-12-06
do not force the use of a noop logger, it can be changed in the builder

bjartek:
This was overlooked in a refactoring, the functional builder loop can change the logger and the zerolog.Nop() line reverts changes to the logger.


bluesign (commented):
thanks for the fix

bluesign (approved):

PR #1300, opened 2023-12-12
add scaffold id in as a flag

### LINKED ISSUE #1271: Scaffold a specific project template with `--scaffold={id}`
### Issue To Be Solved

Currently, you can only scaffold a project in "interactive mode" with `--scaffold` flag.

This isn't ideal for these reasons:
- makes it harder to provide instructions in the documentation (e.g. start by scaffolding X project)
- makes it hard to reuse `flow setup --scaffold` by 3rd party tools
- nit: more work if you already know which project to scaffold (in case you are doing that repeatedly when playing around, etc...)

For example, the bellow instructions from [hybrid custody scaffold repo](https://github.com/onflow/hybrid-custody-scaffold#-start-developing) could be shortened to just one step: `flow setup --scaffold=hybrid-custody`.

Screenshot 2023-11-15 at 10 38 22

### Suggest A Solution

Make the `--scaffold` flag optionally accept a scaffold ID argument. If that argument is present, immediately scaffold that project instead of launching the interactive scaffold flow.

This would also require adding an `id` field to the scaffolds json: https://github.com/onflow/flow-cli/blob/master/scaffolds.json

### Context

We are thinking of better streamlining project creation within Flowser. A part of that is also enabling users to scaffold new projects easily within Flowser itself. We could implement our scaffolding logic, but it could be even better to just reuse the Flow CLI `setup` command for this. And other tools or users could also benefit from this change.

ianthpun:
Closes #1271 
## Description

______

For contributor use:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,7 +168,6 @@ with the bellow content.
 
 ```json
 {
-  "id": "unique id of the scaffold",
   "name": "name of the scaffold",
   "repo": "HTTP URL to the repository containing the scaffold",
   "commit": "Hash of the (latest) commit in your scaffold repository",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,6 +168,7 @@ with the bellow content.
 
 ```json
 {
+  "id": "unique id of the scaffold",
   "name": "name of the scaffold",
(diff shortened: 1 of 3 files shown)


chasefleming (approved):
Looks good thanks for adding this!

chasefleming:
@ianthpun should probably also update the docs with this

PR #1319, opened 2023-12-21
`add` and `install` from live networks with dependency manager

### LINKED ISSUE #944: Install contract dependencies from live Flow networks
## Goal

`flow contracts install` should check and fetch contracts defined in `flow.json` from live Flow networks into the `imports` folder.
- The simple syntax is `NETWORK/0xADDRESS/CONTRACT_NAME` e.g. `mainnet/0x0123123123/TopShot`
- Only applies to missing dependencies based on what's inside the `imports` folder.
- Will be repeated recursively for every other imported contract from the original. The sub-dependencies are checked out into subfolders inside `imports`.
- `imports` should be added to `.gitignore`

## Example
https://github.com/jrkhan/cadence-import

chasefleming:
Closes #944 

### Add Command

To run the add command, you can run:

```
flow dependencies add testnet://0afe396ebc8eee65.FLOAT
```

or

```
flow deps add testnet://0afe396ebc8eee65.FLOAT
```

### Install Command

You can also run `flow dependencies install` if you add the dependencies to your `flow.json` manually.
______

For contributor use:


bjartek:
In this flix-wg we have spoken about a contract to store interactions in a contracts account on different networks. I have stated that i think it would give tons of value for that registry/or something else to know about "where am i, on other networks". I think that should be taken into consideration in this work.

internal/dependencymanager/add.go
```diff
@@ -0,0 +1,67 @@
...
+	args []string,
+	_ command.GlobalFlags,
+	logger output.Logger,
+	flow flowkit.Services,
+	state *flowkit.State,
+) (result command.Result, err error) {
+	logger.Info(fmt.Sprintf("🔄 Installing dependencies for %s...", args[0]))
+
+	dep := args[0]
+
```
ianthpun, line 57: we should log any new depencies added, similiar to how `go mod` works :)

internal/dependencymanager/dependencyinstaller.go
```diff
@@ -0,0 +1,249 @@
...
+)
+
+type DependencyInstaller struct {
+	Gateways map[string]gateway.Gateway
+	Logger   output.Logger
+	State    *flowkit.State
+	Mutex    sync.Mutex
+}
+
+func NewDepdencyInstaller(logger output.Logger, state *flowkit.State) *DependencyInstaller {
```
ianthpun:
```suggestion
func NewDependencyInstaller(logger output.Logger, state *flowkit.State) *DependencyInstaller {
```

internal/dependencymanager/dependencyinstaller.go
```diff
@@ -0,0 +1,249 @@
...
+	emulatorGateway, err := gateway.NewGrpcGateway(config.EmulatorNetwork)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error creating emulator gateway: %v", err))
```
ianthpun, line 52: we are logging the error but still contiuning down the chain of the funciton, is that intended?

internal/dependencymanager/dependencyinstaller.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (ci *DependencyInstaller) add(depRemoteSource, customName string) error {
+	depNetwork, depAddress, depContractName, err := config.ParseRemoteSourceString(depRemoteSource)
+	if err != nil {
+		return fmt.Errorf("error parsing remote source: %w", err)
+	}
+
+	var name string
+
+	if customName != "" {
```
ianthpun, line 96:
```suggestion
       name := depContractName
	if customName != "" {
	    name = customName
	}
	...
```

internal/dependencymanager/dependencyinstaller.go
```diff
@@ -0,0 +1,249 @@
...
+	// If it is, ask if they want to update
+	// If no hash, ignore
+	if dependency != nil && dependency.Version != "" && dependency.Version != originalContractDataHash {
+		msg := fmt.Sprintf("The latest version of %s is different from the one you have locally. Do you want to update it?", contractName)
+		if !util.GenericBoolPrompt(msg) {
+			return nil
+		}
+	}
+
+	if !contractFileExists(contractAddr, contractName) {
```
ianthpun, line 215: as discussed offline, we can move the mutex call here to improve speed

flowkit/config/contract.go
```diff
@@ -106,3 +111,89 @@ func (c *Contracts) Remove(name string) error {
...
+	return nil
+}
+
+// AddDependencyAsContract adds a dependency as a contract if it doesn't already exist.
+func (c *Contracts) AddDependencyAsContract(dependency Dependency, networkName string) {
+	var aliases []Alias
+
+	// If core contract found by name and address matches, then use all core contract aliases across networks
+	if isCoreContract(networkName, dependency.RemoteSource.ContractName, dependency.RemoteSource.Address.String()) {
+		for _, networkStr := range []string{NetworkEmulator, NetworkTestnet, NetworkMainnet} {
```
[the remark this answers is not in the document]
bjartek: Add crescendo here? Or do we not think this will be out in time for that?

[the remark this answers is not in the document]
chasefleming: That's a good catch @bjartek . Do you know if it is supported in the core contract library yet? Also, adding this ticket to add Crescendo across the whole CLI: https://github.com/onflow/flow-cli/issues/1350

internal/dependencymanager/add.go
chasefleming: Added 👍

internal/dependencymanager/dependencyinstaller.go
chasefleming: Fixed in another commit

flowkit/CHANGELOG.md
```diff
@@ -87,9 +87,9 @@ GenerateMnemonicKey(context.Context, crypto.SignatureAlgorithm, string) (crypto.
 
 DerivePrivateKeyFromMnemonic(context.Context, string, crypto.SignatureAlgorithm, string) (crypto.PrivateKey, error)
 
-// DeployProject contracts to the Flow network or update if already exists and update is set to true.
+// DeployProject dependencymanager to the Flow network or update if already exists and update is set to true.
```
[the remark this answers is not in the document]
chasefleming: Good catch. This shouldn't be here.

flowkit/config/contract.go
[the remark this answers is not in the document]
bjartek: Nopes dont know

flowkit/config/contract.go
```diff
@@ -106,3 +111,96 @@ func (c *Contracts) Remove(name string) error {
...
+	if len(aliases) == 0 {
+		aliases = append(aliases, Alias{
+			Network: dependency.RemoteSource.NetworkName,
+			Address: dependency.RemoteSource.Address,
+		})
+	}
+
+	contract := Contract{
+		Name:         dependency.Name,
+		Location:     filepath.ToSlash(fmt.Sprintf("imports/%s/%s", dependency.RemoteSource.Address, dependency.RemoteSource.ContractName)),
```
nvdtf, line 188: minor, maybe `imports` can be a const

flowkit/config/dependency.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/onflow/flow-go-sdk"
+)
+
+func ParseRemoteSourceString(s string) (network, address, contractName string, err error) {
+	parts := strings.Split(s, "://")
+	if len(parts) != 2 {
+		return "", "", "", fmt.Errorf("invalid dependency source format")
```
nvdtf, line 31: When using `Errorf` it's best to add more context, like the `s` string that caused the error:
```go
fmt.Errorf("invalid dependency source format: %s", s)
```
If you don't have any args you can use `errors.New(str)` (https://pkg.go.dev/errors#New)

flowkit/config/dependency.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	network = parts[0]
+
+	subParts := strings.Split(parts[1], ".")
+	if len(subParts) != 2 {
+		return "", "", "", fmt.Errorf("invalid dependency source format")
```
nvdtf, line 37: Same as above

flowkit/config/dependency.go
```diff
@@ -0,0 +1,79 @@
...
+type RemoteSource struct {
+	NetworkName  string
+	Address      flow.Address
+	ContractName string
+}
+
+type Dependency struct {
+	Name         string
+	RemoteSource RemoteSource
+	Version      string
```
nvdtf, line 54: I noticed the hash is used as version, maybe we can call it `CodeHash` for clarity? What do you think?

flowkit/config/json/dependency_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	dependencyOne := dependencies.ByName("HelloWorld")
+	assert.NotNil(t, dependencyOne)
+
+	assert.NotNil(t, dependencyOne)
+}
+
+func Test_TransformDependenciesToJSON(t *testing.T) {
+	b := []byte(`{
+		"HelloWorld": "testnet://877931736ee77cff.HelloWorld"
```
nvdtf, line 50: This comment applies to a lot more than just this line. I think it's best to always write addresses with `0x` prefix because they are more clear to read. This applies to folders we create under `imports` as well. Anywhere in the new JSON structure when an address is written (like in simple format) it should have `0x` prefix. What do you think?

flowkit/project/program.go
```diff
@@ -42,18 +43,34 @@ func NewProgram(code []byte, args []cadence.Value, location string) (*Program, e
...
 	return &Program{
-		code:       code,
-		args:       args,
-		location:   location,
-		astProgram: astProgram,
+		code:            code,
+		args:            args,
+		location:        location,
+		astProgram:      astProgram,
+		developmentCode: code, // has converted import syntax e.g. 'import "Foo"'
```
nvdtf, line 50: `codeWithUnprocessedImports`? `unprocessedImportsCode` ? trying more clarity

nvdtf (commented):
Nice job @chasefleming 🏅  this is great! Left some minor comments

flowkit/config/json/dependency_test.go
bjartek: I suggest using identifier here because it is known already. So A.address.contract

bjartek:
I wrote somewhere that it could be sn idea to. Have a resource we write to the same storage as contracts when we deploy that has some information about related addresses. The most useful one here would be a self relation on another network.

bjartek:
It is possible to know what network an address is, so do the command really need to specify the network? Can it guess based on input?

bjartek:
What if i want to add all contracts on an address xan i do that easily?

cmd/flow/main.go
```diff
@@ -117,8 +119,13 @@ func main() {
 	cmd.AddCommand(snapshot.Cmd)
 	cmd.AddCommand(super.FlixCmd)
 	cmd.AddCommand(super.GenerateCommand)
+	cmd.AddCommand(dependencymanager.Cmd)
 
 	command.InitFlags(cmd)
+	cmd.AddGroup(&cobra.Group{
```
sideninja, line 125: Better to make sure the super commands are always on top. Contract manager should be after

internal/dependencymanager/add.go
```diff
@@ -0,0 +1,67 @@
...
+	name string `default:"" flag:"name" info:"Name of the dependency"`
+}
+
+var addFlags = addFlagsCollection{}
+
+var addCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:   "add",
+		Short: "Add a single contract and its dependencies.",
+		Args:  cobra.ExactArgs(1),
```
sideninja, line 41: add example doc

internal/dependencymanager/add.go
```diff
@@ -0,0 +1,67 @@
...
+
+type addFlagsCollection struct {
+	name string `default:"" flag:"name" info:"Name of the dependency"`
+}
+
+var addFlags = addFlagsCollection{}
+
+var addCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:   "add",
```
sideninja, line 39: add placeholder to tell what to provide for an argument, check other commands for syntax, I believe something like: 
```
add <name of the dependency>
```

sideninja:
I've noticed files in imports don't have .cdc extension, is there a reason?

sideninja:
If using add with invalid contract name it doesn't error out, example:
```
flow deps add mainnet://0x1d7e57aa55817448.NonFungibleToken123
```

sideninja:
What happens if I use a contract that doesn't have aliases for the emulator and then I want to deploy my project using that contract locally?
One example to reproduce would be:
```
flow deps add mainnet://0xb78ef7afa52ff906.SwapError
```

and then import that swap error contract in my test contract locally and try to deploy to the emulator. That doesn't work right now, I wonder whether it should.

chasefleming:
> What if i want to add all contracts on an address xan i do that easily?

Not yet @bjartek , do you see a use for that. Let's make a ticket for a command for that. There will be additional commands we'll need. This first mvp is just including `add` and `install`. Would be curious to know other commands you'd find useful as well.

chasefleming:
> I've noticed files in imports don't have .cdc extension, is there a reason?

Good catch. Meant to fix that. Updated.

flowkit/config/dependency.go
chasefleming: Updated

flowkit/config/json/dependency_test.go
chasefleming: I like `0x` but it may cause some inconsistencies with other pieces of the CLI. For example, the addresses created on `init` are this format, as well as addresses created by `flow dev`.

flowkit/config/dependency.go
chasefleming: I struggled on naming this one. Finding the right balance of being understandable and being specific about what it was was difficult. I don't really have an opinion. Other options are just `Hash`, `CheckSum`, `Id`

chasefleming:
> I wrote somewhere that it could be sn idea to. Have a resource we write to the same storage as contracts when we deploy that has some information about related addresses. The most useful one here would be a self relation on another network.

It will be helpful for populating aliases. When it exists, we can add it in.

chasefleming:
> If using add with invalid contract name it doesn't error out, example:
> 
> ```
> flow deps add mainnet://0x1d7e57aa55817448.NonFungibleToken123
> ```

Good catch. Fixed. Should log an error now.

chasefleming:
> What happens if I use a contract that doesn't have aliases for the emulator and then I want to deploy my project using that contract locally? One example to reproduce would be:
> 
> ```
> flow deps add mainnet://0xb78ef7afa52ff906.SwapError
> ```
> 
> and then import that swap error contract in my test contract locally and try to deploy to the emulator. That doesn't work right now, I wonder whether it should.

@sideninja Can expand on this? Why would the local version you pulled not deploy to the emulator if you set it in deployments?

flowkit/project/program.go
```diff
@@ -42,18 +43,34 @@ func NewProgram(code []byte, args []cadence.Value, location string) (*Program, e
...
+
+func (p *Program) HasAddressImports() bool {
+	return len(p.AddressImportDeclarations()) > 0
+}
+
+// Imports builds an array of all the import locations
 // It currently supports getting import locations as identifiers or as strings. Strings locations
 // can represent a file or an account name, whereas identifiers represent contract names.
-func (p *Program) imports() []string {
+func (p *Program) pathImports() []string {
```
jribbink, line 73: nit: I think the terminology that is used elsewhere for this type of import is "string import".  It encompasses both path imports & identifier imports (i.e. `"import X from "Y""` or `"import "X""`).

we should maybe try to stay consistent in the naming

chasefleming:
Docs PR: https://github.com/onflow/docs/pull/524

chasefleming:
> > What happens if I use a contract that doesn't have aliases for the emulator and then I want to deploy my project using that contract locally? One example to reproduce would be:
> > ```
> > flow deps add mainnet://0xb78ef7afa52ff906.SwapError
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then import that swap error contract in my test contract locally and try to deploy to the emulator. That doesn't work right now, I wonder whether it should.
> 
> @sideninja Can expand on this? Why would the local version you pulled not deploy to the emulator if you set it in deployments?

I understand what you're saying now. Yeah it's worth thinking if they should be automatically added to the `emulator-account` if they are not core contracts. Right now `flow config add deployment` will still work though. We could add an update later if we decided to add this functionality and someone would just need to run `install` again.

bjartek:
What about letting ppl chose whqt account to add them to on emulator

jribbink (approved):

chasefleming:
> What about letting ppl chose what account to add them to on emulator

Yep @bjartek  I'm hesitant to make assumptions, which is why `flow config add deployment` seems like a good option for now. None of the functionality differs from the contracts at the moment. We can add in some nice features to streamline things as people try it out and see how people are using it.

bjartek:
> > What about letting ppl chose what account to add them to on emulator
> 
> 
> 
> Yep @bjartek  I'm hesitant to make assumptions, which is why `flow config add deployment` seems like a good option for now. None of the functionality differs from the contracts at the moment. We can add in some nice features to streamline things as people try it out and see how people are using it.

Just a flag in install could work aswell --emulator-account or somethibg

chasefleming:
Yeah that would be cool @bjartek . I will open an issue to discuss this as an upcoming feature.

PR #1326, opened 2024-01-03
move create-interactive, support flow-cli testnet link with flow account fund

### LINKED ISSUE #1213: Add a command for funding an account on testnet
Having to manually add your address to fund an account from the [Faucet site](https://testnet-faucet.onflow.org/fund-account) while developing is a little tedious. It would be nice if you could run a command like the following to fund an account:

`flow fund testnet-acct`

This would take the account from your flow.json and send it 1000 like on the Faucet site. We could also add a `--amount` flag if the dev wanted to customize the amount but it could also be added later. Not sure the flag is necessary.

Note: should error with a message if you try to fund an account that isn't testnet.

ianthpun:
Closes #1213 

## Description

image

Also added the `flow account fund [ACCOUNT]`  command, which generates this and pops a browser window:

image

______

For contributor use:


internal/accounts/accounts.go
```diff
@@ -86,6 +86,13 @@ func (r *accountResult) String() string {
 	var b bytes.Buffer
 	writer := util.CreateTabWriter(&b)
 
+	if r.Address.IsValid(flow.Testnet) {
```
jribbink, line 89: Can we base this off of `selectedNetwork` instead? https://github.com/onflow/flow-cli/pull/1326/files#diff-8cfd65a8348c96dd493e6ed1bc0b2c91340dc16c9026b1e7d8d4c71d1d20de5eR53

If we only validate by the codeword, then crescendo and testnet will be the same, so probably best if we differentiate between them (I'm assuming crescendo will eventually be added to the CLI?).

bjartek (approved):

jribbink (changes requested):

jribbink (approved):

chasefleming:
On the issue there is the `fund` command. I think it might still be useful to have that command open up this pre populated url. Curious what others think.

jribbink:
> On the issue there is the `fund` command. I think it might still be useful to have that command open up this pre populated url. Curious what others think.

Yeah I can see a few situations in which a dev wants to later fund an account so no harm in keeping that door open IMO.  I think best structure is probably `flow accounts fund MY_ACCT`, that way we avoid bloating top-level commands as much as possible.  It's probably a really easy task so we might as well just do it

internal/accounts/accounts.go
```diff
@@ -86,6 +86,13 @@ func (r *accountResult) String() string {
...
+		_, _ = fmt.Fprintf(writer, fmt.Sprintf(
+			"If you would like to fund the account with 1000 FLOW tokens for testing,"+
+				" visit %s \n\n",
+			fmt.Sprintf("https://testnet-faucet.onflow.org/fund-account?address=%s", r.Address.String())))
```
turbolent, line 93: `Fprintf` already formats, so no need for extra formats inside:

```suggestion
		_, _ = fmt.Fprintf(
			writer,
			"If you would like to fund the account with 1000 FLOW tokens for testing,"+
				" visit https://testnet-faucet.onflow.org/fund-account?address=%s\n\n",
		 	r.Address.String(),
		)
```

turbolent (approved):

internal/accounts/accounts.go
ianthpun: updated and thank you!

ianthpun:
@chasefleming @jribbink ive now added the `flow accounts fund MY_ACCT` functionality  which would pop a browser window for you to complete the hcaptcha and get your account funded. It also has validation its only for testnet addresses. I'll make sure to update the docs on this next once this has been merged. THanks everyone!

Issue #1330, opened 2024-01-05
User address missing prefix generating FLIX for local environment


bthaile:
### Instructions

Generate a FLIX using emulator using contracts deployed by local user.

### Problem

emulator addresses are missing 0x prefix when dependencies are created to pass into flixkit-go for generating flix json

### Steps to Reproduce

deploy contracts and construct a transaction or script that uses a deployed contract. After generating a FLIX, notice that address for the emulator is missing the prefix of "0x"

PR #1331, opened 2024-01-05
add prefix to user address when processing flow.json

bthaile:
Closes: https://github.com/onflow/flow-cli/issues/1330

## Description

add prefix to user address when processing flow.json
______

For contributor use:


bjartek (approved):

chasefleming (changes requested):
There should be a Go SDK utility to do this. I think it's `HexToAddress`

bjartek (changes requested):
I think address has a Hex method that will do this.

chasefleming (approved):
Just a heads up, this file is getting close to needing some tests

chasefleming:
@bthaile looks like lint step didn't pass

bthaile:
> Just a heads up, this file is getting close to needing some tests

https://github.com/onflow/flow-cli/issues/1335
created this ticket to track adding unit tests. I've moved a lot of the complexity to flixkit-go itself, I do agree flix.go needs tests.

bthaile:
> I think address has a Hex method that will do this.

`Hex()` does not include "0x" that I see, maybe it's how I'm printing out the string fmt.Println or fmt.Sprintf doesn't print out a leading "0x"

bthaile:
> `Hex()` does not include "0x" that I see, maybe it's how I'm printing out the string fmt.Println or fmt.Sprintf doesn't print out a leading "0x"

Hex() does not include "0x" that I see, maybe it's how I'm printing out the string fmt.Println or fmt.Sprintf doesn't print out a leading "0x". I changed to use BytesToAddress to create a flow.Address `address := flow.BytesToAddress([]byte(addr))`

chasefleming:
@bthaile Can you create an issue in Go SDK if there is no method available for what you need so it's available in the future for others? You can also also make the PR to the Go SDK so we are avoiding hardcoding.

bthaile:
> @bthaile Can you create an issue in Go SDK if there is no method available for what you need so it's available in the future for others? You can also also make the PR to the Go SDK so we are avoiding hardcoding.

flow-go-sdk, added issue and PR.
https://github.com/onflow/flow-go-sdk/pull/554

Any other blockers on this PR?

bthaile:
> I think address has a Hex method that will do this.

nope `Hex()` does not add "0x"

bjartek (approved):

PR #1337, opened 2024-01-09
emulator private key file

### LINKED ISSUE #1325: emulator pkey
### Instructions

Please fill out the template below to the best of your ability and include a label indicating which tool/service you were working with when you encountered the problem.

### Problem

When you run `flow init`, it puts the emulator pkey inside the flow.json file. When you then take any action on testnet/mainnet using Flow CLI, it complains about the emulator pkey being in the flow.json file.

### Suggested Fix

When a user runs `flow init`, make a `emulator-account.pkey` file that holds the emulator private key so it no longer complains. Very similar to how `testnet-account.pkey` gets made when you run `flow accounts create`

ianthpun:
Closes #1325 

## Description
- create a `emulator_account.pkey` on `flow init` instead of using private key on `flow.json`
- Add readerWriter to PrivateKey() call in FileKey

______

For contributor use:

### DIFF
--- a/flowkit/accounts/account.go
+++ b/flowkit/accounts/account.go
@@ -19,6 +19,7 @@
 package accounts
 
 import (
+	"crypto/rand"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


chasefleming:
@ianthpun Can you add the correct label and make the title a little more descriptive? These are what will show in the release notes. Thanks.

chasefleming:
@ianthpun Tried it and it's working great. But one question: what do you think we should do about the scaffolds (especially the one used in `setup`)? Because it still uses the old way so we may want to make a PR to change it to follow this pattern, but if the emulator key is gitignored if you pull the scaffold it won't work. That may be one reason to not gitignore the emulator key used for the main service account.

Also, have you tested this with `flow dev` just to verify it is all working fine?

bjartek:
should there be a way to migrate an existing flow.json to this new format?

chasefleming:
@bjartek That's a cool feature idea. Probably a separate ticket though.

bjartek (approved):

jribbink (approved):

PR #1410, opened 2024-02-14
Create Cadence 1.0 pre-release installer

### LINKED ISSUE #1374: Stable Cadence Installation Workflow
The current workflow of supporting stable cadence overrides the existing `flow-cli` binary

https://forum.flow.com/t/update-on-cadence-1-0/5197/7

Until Stable cadence is offically out and we stop supporting anything before (v1.X deprecation), we will have a new workflow that can allow both binaries to exist at once.

A solution to that would be having a `flow` and `flow-sc` binary that is created, and that `flow-sc` is going to use the latest pre-release version of `flow-cli` at `v2.0.0.stable-cadence.X`, which will always have the latest changes on `master`, but with the breaking stable cadence changes

jribbink:
Closes #1374 

## Description

Creates a Cadence 1.0 pre-release installer.  This installer will iterate through all of the Flow CLI releases (99.99999% will be on first page anyways) and install the latest release containing `cadence-v1.0.0`.

 - We should update C1.0 install instructions to run the install script from `stable-cadence` branch instead of master
 - We should make sure that all C1.0 releases are tagged with `vX.X.X-cadence-v1.0.0.X`
 - We should make sure that we do not merge these changes into master as they are C1-specific and temporary

______

For contributor use:


install.ps1
```diff
@@ -39,16 +39,54 @@ $repo = "onflow/flow-cli"
...
+        [string]$searchTerm,
+        [string]$githubTokenHeader
+    )
+
+    $page = 1
+    $version = null
+
+    while (-not $version) {
+        if ($githubTokenHeader) {
+            $response = Invoke-WebRequest -Uri "https://api.github.com/repos/$repo/releases?per_page=100&page=$page" -Headers @{ 'Authorization' = $githubTokenHeader } -UseBasicParsing -ErrorAction SilentlyContinue
```
chasefleming, line 55: Just confirming we want `-ErrorAction SilentlyContinue` here

chasefleming, line 55: Is `$githubTokenHeader` exposed and is that fine?

install.ps1
```diff
@@ -39,16 +39,54 @@ $repo = "onflow/flow-cli"
...
+            Write-Output "Failed to get latest version from Github API, please manually specify a version to install as an argument to this script."
+            return $null
+        }
+
+        $jsonResponse = $response.Content | ConvertFrom-Json
+
+        foreach ($release in $jsonResponse) {
+            if ($release.tag_name -like "*$searchTerm*") {
+                $version = $release.tag_name
+                break
```
chasefleming, line 78: What if no tag is found?

chasefleming:
@jribbink what about uninstalling? Should we add support for that too?

install.ps1
jribbink: If no tag is found, it will reach the end of the outer loop, call `$page++`, and iterate again.  Once it runs out of pages (`$status -ne 200`), it will return null.

install.ps1
jribbink: I just changed this anyways so that we didn't have 2 different codepaths.  But yeah, it works now I just tested on a Windows machine.

jribbink: Yup we need it so that it doesn't throw an exception when it doesn't receive an OK response.  We catch these manually and decide whether to continue iterating pages.

jribbink:
> @jribbink what about uninstalling? Should we add support for that too?

I think this is probably a separate ticket.  It's not blocking to get this functionality out & we don't even provide a mechanism (outside of homebrew) to do this with our existing install script.  It's probably a question of providing an `uninstall.sh`/`uninstall.ps1` script IMO.

ianthpun (approved):
code looks fine, lets give it a shot

PR #1419, opened 2024-02-20
EVM Gateway command

sideninja:
Added a command that starts up the Flow EVM Gateway: https://github.com/onflow/flow-evm-gateway

There is only command for now:
```
  flow evm gateway [flags]

Examples:
flow evm gateway


Flags:
      --access-node-grpc-host string   host to the flow access node gRPC API (default "localhost:3569")
      --coa-address string             Flow address that holds COA account used for submitting transactions
      --coa-key string                 WARNING: do not use this flag in production! private key value for the COA address used for submitting transactions
      --coa-resource-create            auto-create the COA resource in the Flow COA account provided if one doesn't exist
      --coinbase string                coinbase address to use for fee collection
      --database-dir string            path to the directory for the database (default "./db")
      --gas-price string               static gas price used for EVM transactions (default "1")
      --init-height uint               init cadence block height from where the event ingestion will start. WARNING: you should only provide this if there are no existing values in the database, otherwise an error will be thrown
      --network-id string              EVM network ID (testnet, mainnet) (default "testnet")
      --rpc-host string                host for the RPC API server (default "localhost")
      --rpc-port int                   port for the RPC API server (default 3000)

```


bthaile (approved):
A few `evm ..` example commands would be beneficial. Format of the coa-address in particular.

bthaile:
Also, will there be a way to look up a transaction hash on EVM?

go.mod
```diff
@@ -16,6 +17,8 @@ require (
 	github.com/onflow/flixkit-go v1.1.1-0.20240214222351-03b90f7d32ef
 	github.com/onflow/flow-core-contracts/lib/go/templates v0.15.2-0.20240206003101-928bf99024d7
 	github.com/onflow/flow-emulator v1.0.0-M5
+	github.com/onflow/flow-evm-gateway v0.0.0-20240220202014-12947709fffd
+	github.com/onflow/flow-go v0.33.2-0.20240220142056-d1573c75f26c
```
turbolent, line 21: This seems to change the flow-go version. Is this intended?

internal/evm/gateway.go
```diff
@@ -0,0 +1,99 @@
...
+			return nil, fmt.Errorf("invalid COA address value")
+		}
+
+		pkey, err := crypto.DecodePrivateKeyHex(crypto.ECDSA_P256, flagGateway.COAKey)
+		if err != nil {
+			return nil, fmt.Errorf("invalid COA key: %w", err)
+		}
+		cfg.COAKey = pkey
+
+		switch flagGateway.NetworkID {
```
turbolent, line 88: Can the Emulator be supported?

turbolent:
@sideninja Any update here? It would be great to have developers just install one CLI pre-release, which has both Cadence 1.0 with EVM support and EVM gateway

turbolent:
Fixed conflicts in [5a1d7c7](https://github.com/onflow/flow-cli/pull/1419/commits/5a1d7c7527f2d01e1a4704da7672db66cdf5d328)

turbolent:
CLI wants Go 1.20, but the version of the EVM gateway used by this PR still requires Go 1.21. The version of the EVM gateway must be bumped to at least include https://github.com/onflow/flow-evm-gateway/pull/75, which adds support for Go 1.20

internal/evm/gateway.go
m-Peter: @turbolent Yeah, certainly. All this time we used the Emulator as an Access Node backend. But there's no dedicated chain ID for it, so we might have to use `emulator.FlowEVMTestnetChainID`, both `testnet` & `mainnet`.

turbolent:
@sideninja @m-Peter When you get a moment, could you please review [bf0cd0a](https://github.com/onflow/flow-cli/pull/1419/commits/bf0cd0ad34c44a7a93b23cae698acb7b9b779791)? I tried to align it with https://github.com/onflow/flow-evm-gateway/blob/v0.1.0/config/config.go and https://github.com/onflow/flow-evm-gateway/blob/v0.1.0/cmd/main/main.go

sideninja (commented):
LGTM

turbolent (approved):

PR #1423, opened 2024-02-21
Switch to goreleaser & add support for CGO cross-compilation

jribbink:
### Description

Switches go-release-action to go-releaser & adds now cross compiles with `CGO_ENABLED=1`

Windows builds are excluded currently, pending a new flow-go release with https://github.com/onflow/flow-go/pull/5430

______

For contributor use:

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,6 +1,7 @@
 before:
   hooks:
     - make generate
+    - export LD_FLAGS="-X github.com/onflow/flow-cli/build.commit={{ .Env.COMMIT }} -X github.com/onflow/flow-cli/build.semver={{ .Env.VERSION }} -X github.com/onflow/flow-cli/internal/command.mixpanelToken={{ .Env.MIXPANEL_PROJECT_TOKEN }} -X github.com/onflow/flow-cli/internal/accounts.accountToken={{ .Env.LILICO_TOKEN }}"
 
 builds:
   - id: darwin-amd64
@@ -15,7 +16,7 @@ builds:
     flags:
       - -mod=readonly
     ldflags:
-      - -X github.com/onflow/flow-cli/build.commit={{ .Env.COMMIT }} -X github.com/onflow/flow-cli/build.semver={{ .Env.VERSION }} -X github.com/onflow/flow-cli/internal/command.mixpanelToken={{ .Env.MIXPANEL_PROJECT_TOKEN }} -X github.com/onflow/flow-cli/internal/accounts.accountToken={{ .Env.LILICO_TOKEN }}
+      - {{ .Env.LD_FLAGS }}
 
   - id: darwin-arm64
     main: ./cmd/flow
@@ -29,7 +30,7 @@ builds:
     flags:
       - -mod=readonly
     ldflags:
-      - -X github.com/onflow/flow-cli/build.commit={{ .Env.COMMIT }} -X github.com/onflow/flow-cli/build.semver={{ .Env.VERSION }} -X github.com/onflow/flow-cli/internal/command.mixpanelToken={{ .Env.MIXPANEL_PROJECT_TOKEN }} -X github.com/onflow/flow-cli/internal/accounts.accountToken={{ .Env.LILICO_TOKEN }}
+      - {{ .Env.LD_FLAGS }}
 
   - id: linux-amd64
     main: ./cmd/flow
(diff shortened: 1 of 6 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - CC=aarch64-linux-gnu-gcc
+      - CXX=oa64-aarch64-linux-gnu-g++
+    flags:
+      - -mod=readonly
+    ldflags:
+      - -X github.com/onflow/flow-cli/build.commit={{ .Env.COMMIT }} -X github.com/onflow/flow-cli/build.semver={{ .Env.VERSION }} -X github.com/onflow/flow-cli/internal/command.mixpanelToken={{ .Env.MIXPANEL_PROJECT_TOKEN }} -X github.com/onflow/flow-cli/internal/accounts.accountToken={{ .Env.LILICO_TOKEN }}
+
+archives:
+  - id: golang-cross
+    builds:
```
tarakby, line 64: we could also add another `windows-amd64` build above to get it ready while all this work is fresh, but not call it under `archives` for now (till flow-go is released).

`CC` would be `x86_64-w64-mingw32-gcc`, assuming the goreleaser image has it too.

tarakby (approved):
Nice, thanks for updating the release process! 

I left a few comments, all of them are optional and not necessary to merge.

.goreleaser.yaml
jribbink: Sounds good, I've added everything needed for the Windows builds and just left it out in the list of `GOOS` targets

.goreleaser.yaml
```diff
@@ -0,0 +1,81 @@
...
+builds:
+  - id: darwin-amd64
+    main: ./cmd/flow
+    goos:
+      - darwin
+    goarch:
+      - amd64
+    env:
+      - CC=o64-clang
+      - CXX=o64-clang++
```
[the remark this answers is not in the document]
jribbink: I was just planning to leave it since we get it shipping in the docker image to us anyways even though we don't use it, but I can remove it if wanted

jribbink:
Dummy release run here showing it works: https://github.com/onflow/flow-cli/actions/runs/8009149116/job/21877169848

bthaile (approved):
I would think we need to get this merged up so we can test it in full production process.

.goreleaser.yaml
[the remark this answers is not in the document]
tarakby: You're right the docker image comes with all the c++ tool chains. The `go build` command looks at the `CXX` variable only when the project has some cpp files, but I believe we don't have any. 
But I said there is nothing wrong with keeping them if it's not extra work for you.

.goreleaser.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - CXX_darwin_arm64=oa64-clang++
+      - CC_linux_amd64=x86_64-linux-gnu-gcc
+      - CXX_linux_amd64=x86_64-linux-gnu-g++
+      - CC_linux_arm64=aarch64-linux-gnu-gcc
+      - CXX_linux_arm64=aarch64-linux-gnu-g++
+      - CC_windows_amd64=x86_64-w64-mingw32-gcc
+      - CXX_windows_amd64=x86_64-w64-mingw32-g++
+      - CC_windows_arm64=aarch64-w64-mingw32-gcc
+      - CXX_windows_arm64=aarch64-w64-mingw32-g++
+      - 'CC={{ index .Env (print "CC_" .Os "_" .Arch) }}'
```
tarakby, line 29: Nice solution!
I'm curious where do `Os` and `Arch` come from? are they equal to `goos` and `goarch` from above?

.goreleaser.yaml
```diff
@@ -0,0 +1,54 @@
...
+    main: ./cmd/flow
+    goos:
+      - darwin
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    env:
+      - CGO_FLAGS="-O2 -D__BLST_PORTABLE__"
+      - CGO_ENABLED=1
```
tarakby, line 16: 👌🏼

.goreleaser.yaml
jribbink: Yeah it's something given to us from goreleaser coming from the goos/goarch above 👍

chasefleming (approved):

PR #1448, opened 2024-03-12
Add `flow cadence lint` command

### LINKED ISSUE #1395: Add `flow cadence lint` command
## Instructions

Please fill out the template below to the best of your ability.

### Issue To Be Solved

Currently, the [Cadence linter](https://github.com/onflow/cadence-tools/tree/master/lint) is only really surfaced through the language server & adjacent tools (VSCode extension, playground).  It is a valuable static-analysis tool which could be a part of automation processes and developers may want to run it independently (outside of integration with the aforementioned tooling).

We should surface this as a command in the Flow CLI so that developers can enforce better standards within their Cadence code and catch errors earlier in their development cycle.

Add a `flow cadence lint <files>` command which runs all linting analyzers as and surfaces these to the output.  The command should also attempt to surface any parsing/checking errors in a structured format.

### (Optional): Context

This relates to efforts related to Cadence 1.0 and helps developers understand if problems exist in their code, what these problems are, and how they can attempt to mediate them.

related: https://github.com/onflow/cadence-tools/pull/285

jribbink:
Closes #1395 

## Description

- Adds `flow cadence lint [files]` to the CLI which mimics diagnostics currently available in LS
- Syntax/semantic errors are captured as well & while non-terminal, will cause the linter to exit with code 1 when finished

_Notes:_

This PR has quite a bit of overlap with the Cadence Language Server.  Unfortunately, it's tricky to refactor this logic into a shared library, and the implementation has to be fairly nuanced.  I tried really hard to use any existing APIs for linting/analysis, but none of them fit the needs here - for similar reasons as the LS so this is something for further exploration.

In order to meet product deadlines & surface this feature to users ASAP, I suggest that we start by inserting the logic into the CLI.  We can work on refactoring this & the LS into a common library (probably inside of the cadence-tools/lint pacakge itself) in a subsequent ticket, carefully considering what this shared interface should actually look like (& try to keep the implementation generalized enough for other users as well).

______

For contributor use:


internal/cadence/cadence.go
```diff
@@ -44,4 +44,5 @@ var Cmd = &cobra.Command{
 
 func init() {
 	Cmd.AddCommand(languageserver.Cmd)
+	lintCommand.AddToParent(Cmd)
```
ianthpun, line 47: is this the correct place to put it? Seems odd that its the only one in this file

jribbink:
CI/lint failing, not sure why it's not matching up locallly... will try to fix

internal/cadence/lint.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"github.com/onflow/cadence/tools/analysis"
+	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flowkit/v2"
+	"github.com/onflow/flowkit/v2/output"
+)
+
+type lintFlagsCollection struct{}
+
+type lintResult struct {
```
ianthpun, line 38: may i suggest we call this something like `lintReport`? so it'll look like
```suggestion
type lintReport struct {
...
...

type lintResults struct {
    Results []lintReport
    ExitCode int

}
```

internal/cadence/cadence.go
jribbink: Yup, this is the same as every other command.  The difference between this command and the `languageserver.Cmd` is that the latter isn't actually using the `command.Command` wrapper that we use for other commands so it looks different

internal/cadence/linter.go
```diff
@@ -0,0 +1,334 @@
...
+
+func maybeProcessConvertableError(
+	err error,
+	location common.Location,
+) (
+	[]analysis.Diagnostic,
+	error,
+) {
+	diagnostics := make([]analysis.Diagnostic, 0)
+	if parentErr, ok := err.(errors.ParentError); ok {
```
jribbink, line 264: don't swallow

internal/cadence/linter.go
```diff
@@ -0,0 +1,334 @@
...
+)
+
+type linter struct {
+	checkers              map[common.Location]*sema.Checker
+	state                 *flowkit.State
+	checkerStandardConfig *sema.Config
+	checkerScriptConfig   *sema.Config
+}
+
+type convertibleError interface {
```
jribbink, line 46: positionedErrror

internal/cadence/linter.go
```diff
@@ -0,0 +1,334 @@
...
+func (l *linter) handleImport(
+	checker *sema.Checker,
+	importedLocation common.Location,
+	_ ast.Range,
+) (
+	sema.Import,
+	error,
+) {
+	switch importedLocation {
+	case stdlib.CryptoCheckerLocation:
```
jribbink, line 166: delete this

internal/cadence/linter.go
```diff
@@ -0,0 +1,334 @@
...
+		}, nil
+	case cdcTests.BlockchainHelpersLocation:
+		helpersChecker := cdcTests.BlockchainHelpersChecker()
+		return sema.ElaborationImport{
+			Elaboration: helpersChecker.Elaboration,
+		}, nil
+	default:
+		importedChecker, ok := l.checkers[importedLocation]
+		if !ok {
+			filepath, err := l.resolveImportFilepath(importedLocation, checker.Location)
```
jribbink, line 184: move to above importedChecker

internal/cadence/stdlib.go
```diff
@@ -0,0 +1,302 @@
+/*
```
jribbink, line 1: add comment to c&p from LS

internal/cadence/linter.go
```diff
@@ -0,0 +1,334 @@
...
+		diagnostics = append(diagnostics, diagnostic)
+	}
+	analysisProgram.Run(analyzers, report)
+
+	return diagnostics, nil
+}
+
+func (l *linter) newCheckerConfig(lib standardLibrary) *sema.Config {
+	return &sema.Config{
+		BaseValueActivationHandler: func(_ common.Location) *sema.VariableActivation {
```
jribbink, line 138: document more

internal/cadence/linter.go
```diff
@@ -0,0 +1,351 @@
...
+
+func getDiagnosticsForParsingCheckingError(
+	err error,
+	location common.Location,
+	code string,
+) (
+	[]analysis.Diagnostic,
+	error,
+) {
+	parentErr, ok := err.(errors.ParentError)
```
ianthpun, line 271: lets use `errors.As` if we can as that is preferred. https://pkg.go.dev/errors#section-documentation
```
var perr *fs.PathError
if errors.As(err, &perr) {
	fmt.Println(perr.Path)
}
is preferable to

if perr, ok := err.(*fs.PathError); ok {
	fmt.Println(perr.Path)
}
because the former will succeed if err wraps an [*io/fs.PathError](https://pkg.go.dev/io/fs#PathError).
```

internal/cadence/linter.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	return checkerDiagnostics, nil
+}
+
+func getDiagnosticsForParentError(err errors.ParentError, location common.Location, code string) ([]analysis.Diagnostic, error) {
+	diagnostics := make([]analysis.Diagnostic, 0)
+
+	for _, childErr := range err.ChildErrors() {
+		convertibleErr, ok := childErr.(positionedError)
```
ianthpun, line 289: ,same as above, also rename `convertibleErr` to `positionedErr`

internal/cadence/linter.go
```diff
@@ -0,0 +1,351 @@
...
+	code string,
+) *analysis.Diagnostic {
+	startPosition := err.StartPosition()
+	endPosition := err.EndPosition(nil)
+
+	var message string
+	var secondaryMessage string
+
+	message = err.Error()
+	if secondaryError, ok := err.(errors.SecondaryError); ok {
```
ianthpun, line 316: same as above

internal/cadence/linter.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+
+		diagnostics = append(diagnostics, errorDiagnostics...)
+	}
+
+	// If the program is nil, nothing can be checked & analyzed so return early
+	if program == nil {
+		return diagnostics, nil
+	}
+
```
ianthpun, line 103: lets make a comment that we are doing both a Parse and also a checker(?), and just briefly describe how both of them are part of what this "linter" function is

internal/cadence/linter.go
```diff
@@ -0,0 +1,351 @@
...
+// Choose the checker config based on the assumed type of the program
+func (l *linter) decideCheckerConfig(program *ast.Program) *sema.Config {
+	if program.SoleTransactionDeclaration() != nil || program.SoleContractDeclaration() != nil {
+		return l.checkerStandardConfig
+	}
+
+	return l.checkerScriptConfig
+}
+
+func (l *linter) handleImport(
```
ianthpun, line 163: whats this function do? Comment would be good

internal/cadence/linter.go
```diff
@@ -0,0 +1,351 @@
...
+		resolvedPath := filepath.Join(filepath.Dir(parentPath), location.String())
+		return resolvedPath, nil
+	default:
+		return "", fmt.Errorf("unsupported location: %T", location)
+	}
+}
+
+// helpers
+
+func getDiagnosticsForParsingCheckingError(
```
ianthpun, line 263: lets keep the function name simple. Nothing about `ParentError` in the implementation tells me its related to Parsing or Checking, so better just to call it `getDiagnostics` or `getDiagnosticsFromError`, but since theres already an `err` arugment, i propose dropping that from the func name

```suggestion
func getDiagnostics(
```

internal/cadence/lint.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+var _ command.ResultWithExitCode = &lintResult{}
+
+var lintFlags = lintFlagsCollection{}
+
+var lintCommand = &command.Command{
+	Cmd: &cobra.Command{
+		Use:     "lint [files]",
+		Short:   "Lint Cadence code",
```
chasefleming, line 56: Could this be slightly more descriptive?

internal/cadence/linter.go
```diff
@@ -0,0 +1,352 @@
...
+package cadence
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"errors"
+
+	cadenceLint "github.com/onflow/cadence-tools/lint"
```
chasefleming, line 28: nit: probably just lowercase these

internal/cadence/lint.go
```diff
@@ -0,0 +1,215 @@
...
+
+		numProblems += len(result.Diagnostics)
+
+		for _, diagnostic := range result.Diagnostics {
+			sb.WriteString(fmt.Sprintf("%s\n\n", renderDiagnostic(diagnostic)))
+		}
+	}
+
+	if numProblems == 0 {
+		sb.WriteString("No problems found")
```
chasefleming, line 188: Should this be green and say "All __ passing" or something more positive than "no problems."

internal/cadence/lint.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (r *lintResult) Oneliner() string {
+	problems := 0
+	for _, result := range r.Results {
+		problems += len(result.Diagnostics)
+	}
+
+	if problems == 0 {
+		return "No problems found"
```
chasefleming, line 207: same here

chasefleming (approved):
Approved, but left a few comments. Your call.

internal/cadence/lint.go
jribbink: I just removed it entirely.  This seems to be more in line with what other tools do, lmk what you think

PR #1451, opened 2024-03-12
 Generate contract test file along with contract boilerplate

### LINKED ISSUE #1394: Generate tests from `flow generate contract`
`flow generate contract` should automatically generate boilerplated test files to help developers test their code.  A flag could be added (`--skip-tests`) to avoid this feature if it is unwanted.

This fits in with https://github.com/onflow/flow-cli/issues/1390 -> the tests could live in `cadence/tests`

chasefleming:
Closes #1394 

## Description

Generate contract test file along with contract boilerplate

______

For contributor use:

### DIFF
--- a/internal/super/generate.go
+++ b/internal/super/generate.go
@@ -237,7 +237,7 @@ fun main() {
 
 		if generateFlags.SkipTests != true {
 			aliases = config.Aliases{{
-				Network: "testing",
+				Network: config.TestingNetwork.Name,
 				Address: flowsdk.HexToAddress("0x0000000000000007"),
 			}}
 		}
--- a/internal/super/generate.go
+++ b/internal/super/generate.go
@@ -20,7 +20,7 @@ package super
 
 import (
 	"fmt"
-	"github.com/onflow/flow-go-sdk"
+	flowsdk "github.com/onflow/flow-go-sdk"
 	"path/filepath"
 	"strings"
 
@@ -237,7 +237,7 @@ fun main() {
 		if generateFlags.SkipTests != true {
 			aliases = config.Aliases{{
 				Network: "testing",
-				Address: flow.HexToAddress("0x0000000000000007"),
+				Address: flowsdk.HexToAddress("0x0000000000000007"),
 			}}
 		}
 
--- a/internal/super/generate.go
+++ b/internal/super/generate.go
@@ -146,7 +146,6 @@ func generateNew(
 	var fileToWrite string
 	var testFileToWrite string
 	var basePath string
-	var testsBasePath = "cadence/tests"
 
 	if generateFlags.Directory != "" {
 		basePath = generateFlags.Directory
@@ -198,7 +197,6 @@ fun main() {
 	}
 
 	filenameWithBasePath := filepath.Join(basePath, filename)
-	testFilenameWithBasePath := filepath.Join(testsBasePath, addCDCExtension(fmt.Sprintf("%s_test", name)))
 
 	// Check file existence
 	if _, err := state.ReaderWriter().ReadFile(filenameWithBasePath); err == nil {
@@ -219,6 +217,9 @@ fun main() {
 	logger.Info(fmt.Sprintf("Generated new %s: %s at %s", templateType, name, filenameWithBasePath))
 
 	if generateFlags.SkipTests != true && templateType == "contract" {
+		testsBasePath := "cadence/tests"
+		testFilenameWithBasePath := filepath.Join(testsBasePath, addCDCExtension(fmt.Sprintf("%s_test", name)))
+
 		if _, err := state.ReaderWriter().ReadFile(testFilenameWithBasePath); err == nil {
 			return nil, fmt.Errorf("file already exists: %s", testFilenameWithBasePath)
 		}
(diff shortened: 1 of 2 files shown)


internal/super/generate.go
```diff
@@ -161,6 +166,19 @@ access(all)
 contract %s {
     init() {}
 }`, name)
+		testFileToWrite = fmt.Sprintf(`import Test
```
ianthpun, line 169: there seems to be some indentation error here? 

also maybe better we have this string as a constant somewhere outside of this function with a comment on what its supposed to do

internal/super/generate.go
```diff
@@ -188,15 +207,47 @@ fun main() {
...
+	// Write files
 	err = state.ReaderWriter().WriteFile(filenameWithBasePath, []byte(fileToWrite), 0644)
 	if err != nil {
 		return nil, fmt.Errorf("error writing file: %w", err)
 	}
 
 	logger.Info(fmt.Sprintf("Generated new %s: %s at %s", templateType, name, filenameWithBasePath))
 
+	if generateFlags.SkipTests != true {
+		if _, err := state.ReaderWriter().ReadFile(testFilenameWithBasePath); err == nil {
```
ianthpun, line 219: should we instantiate thes `testFilenameWithBasePath` and `testBasePath` bits inside here instead? Maybe as  seperate function that this calls so we don't bloat this conditional?

internal/super/generate_test.go
```diff
@@ -40,13 +40,33 @@ func TestGenerateNewContract(t *testing.T) {
...
+	assert.NotNil(t, testContent)
+
 	// Check content is correct
 	expectedContent := `
 access(all)
 contract TestContract {
     init() {}
 }`
+
+	expectedTestContent := `import Test
```
ianthpun, line 54: is this the same as `testFileToWrite`?

internal/super/generate.go
```diff
@@ -188,15 +207,47 @@ fun main() {
...
 
+	// Write files
 	err = state.ReaderWriter().WriteFile(filenameWithBasePath, []byte(fileToWrite), 0644)
 	if err != nil {
 		return nil, fmt.Errorf("error writing file: %w", err)
 	}
 
 	logger.Info(fmt.Sprintf("Generated new %s: %s at %s", templateType, name, filenameWithBasePath))
 
+	if generateFlags.SkipTests != true {
```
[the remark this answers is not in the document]
chasefleming: True although only contracts supports test

internal/super/generate.go
```diff
@@ -188,15 +207,47 @@ fun main() {
...
+		logger.Info(fmt.Sprintf("Generated new test file: %s at %s", name, testFilenameWithBasePath))
+	}
+
 	if templateType == "contract" {
-		state.Contracts().AddOrUpdate(config.Contract{Name: name, Location: filenameWithBasePath})
+		var aliases config.Aliases
+
+		if generateFlags.SkipTests != true {
+			aliases = config.Aliases{{
+				Network: "testing",
```
jribbink, line 240:
```suggestion
				Network: config.TestingNetwork.Name,
```

internal/super/generate.go
```diff
@@ -127,17 +137,12 @@ func generateNew(
...
 	name := args[0]
-	var filename string
-
-	// Don't add .cdc extension if it's already there
-	if strings.HasSuffix(name, ".cdc") {
-		filename = name
-	} else {
-		filename = fmt.Sprintf("%s.cdc", name)
-	}
+	var filename = addCDCExtension(name)
```
jribbink, line 140: nit: It seems like to me that the generate command is meant to accept both names with & without exentions.  (i.e. extension is only added if not already there)

However, the variable `name` is used throughout this function for templating, flow.json filling, etc.  Do we need to strip `.cdc` from `name` as well as add it to `filename`?

i.e. When I run `flow generate contract Foobar.cdc` what happens?  If this is intended to be valid syntax, can we add a test case?

internal/super/generate.go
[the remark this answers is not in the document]
jribbink: gotcha, yeah so I think you need something like this (or other) then:
```suggestion
	if generateFlags.SkipTests != true && templateType == "contract" {
```

internal/super/generate_test.go
chasefleming: Yeah, do you think I should make a function and just use it in the tests?

internal/super/generate.go
chasefleming: It actually needs to be like this. If I add a line break the boilerplate file generated will have a line break. But I can move it to a reusable function.

jribbink (approved):

internal/super/generate_test.go
ianthpun: ya since this is some sort of constant, i would recommend putting it somewhere that can be resused both in the tests and inside `generate`. Then you can also comment on what its used for for better clarity, cause right now its hard to tell what its doing

chasefleming: If it's used in the tests though is it really testing that it was created correctly though? Because if alter it to incorrect code in the file it would change in the test too and pass incorrectly.

internal/super/generate.go
chasefleming: Fixed good catch

PR #1469, opened 2024-03-21
Add Cadence V0.42->V1.0 Update Checker to staging

### LINKED ISSUE #1453: Add Cadence 1.0 Update Checker to Stage Command
## Instructions

Please fill out the template below to the best of your ability.

### Issue To Be Solved

We should add the [cadence 1.0 update checker](https://github.com/onflow/cadence/blob/eecd39315097b1db6ce0bbd62e62ad42a9add7d7/runtime/stdlib/cadence_v0.42_to_v1_contract_upgrade_validator.go#L44) to the staging contracts command.

This could be a preflight check to ensure that there is some basic level of compliance with Cadence V1.0 update rules before somebody stages to the network.  It could prevent friction where developers stage contracts, later find out from external analysis that this is invalid, have to repeat process.  The idea is to catch obvious errors where somebody makes an incompatible update.

(see https://discordapp.com/channels/613813861610684416/1108479699732152503/1217869683030622352)

Add this check but make it optional.  It should not block users from staging entirely as outlined in Discord.

jribbink:
Closes #1453 

## Description

Validates your proposed contract update against other staged contracts on chain.

Screen Shot 2024-03-27 at 4 48 17 PM

______

For contributor use:


internal/migrate/staging_validator.go
```diff
@@ -0,0 +1,427 @@
...
+	contractNames := make([]string, len(e.MissingContracts))
+	for i, location := range e.MissingContracts {
+		contractNames[i] = location.Name
+	}
+	return fmt.Sprintf("the following staged contract dependencies could not be found (have they been staged yet?): %v", contractNames)
+}
+
+var _ error = &missingDependenciesError{}
+
+var chainIdMap = map[string]flow.ChainID{
```
chasefleming, line 79: I've had to do this several times as well. I wonder if there is a way we could break this out to be reusable.

internal/migrate/staging_validator.go
```diff
@@ -0,0 +1,427 @@
...
+func newStagingValidator(flow flowkit.Services, state *flowkit.State) *stagingValidator {
+	return &stagingValidator{
+		flow:         flow,
+		state:        state,
+		contracts:    make(map[common.Location][]byte),
+		elaborations: make(map[common.Location]*sema.Elaboration),
+	}
+}
+
+func (v *stagingValidator) ValidateContractUpdate(
```
chasefleming, line 93: This is quite long. Thoughts on breaking this up for readability?

internal/migrate/staging_validator.go
jribbink: Yeah, just a flowkit helper for converting `flowkit` chain id to flow-go chainid would be good enough here

chasefleming (approved):

PR #1476, opened 2024-03-27
Support previewnet for `accounts create`

### LINKED ISSUE #1474: `flow accounts create` does not work in previewnet
The Lilico API is used to create accounts in the Flow CLI.  However, there are two issues here:

1. Lilico api does not seem to support previewnet.
2. CLI doesn't call a different API endpoint for previewnet (currently, there's only an `if` statement for testnet - so previewnet behaviour defaults to the same as mainnet)

This feature was missed when making the changes needed to support previewnet.

chasefleming:
Closes #1474 

______

For contributor use:

### DIFF
--- a/internal/accounts/create-interactive.go
+++ b/internal/accounts/create-interactive.go
@@ -263,6 +263,7 @@ func (l *lilicoAccount) create(network string) (flowsdk.Identifier, error) {
 	request.Header.Add("Authorization", accountToken)
 
 	client := &http.Client{
+		Timeout: time.Second * 20,
 		Transport: &http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // lilico api doesn't yet have a valid cert, todo reevaluate
 		},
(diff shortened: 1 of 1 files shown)


nvdtf (approved):

jribbink (approved):

internal/accounts/create-interactive.go
```diff
@@ -259,6 +263,7 @@ func (l *lilicoAccount) create(network string) (flowsdk.Identifier, error) {
 	request.Header.Add("Authorization", accountToken)
 
 	client := &http.Client{
+		Timeout: time.Second * 20,
```
jribbink, line 266: Should we add a timeout on `getAccountCreationResult` also?

jribbink (approved):

internal/accounts/create-interactive.go
chasefleming: Is that in flowkit @jribbink ?

jribbink: Yeah not sure.  If it's not easy, don't bother... it's not that important

PR #1488, opened 2024-04-01
Dependency Manager: Prompt to add an alias for other networks on install

### LINKED ISSUE #1363: Allow adding alias when adding a contract with Dependency Manager
When adding a contract with Dependency Manager it would be nice to add a alias for a dependency. Maybe something like this:

`flow deps add testnet://35717efbbce11c74.Find --alias mainnet://097bafa4e0b48eef`

Additionally, it should determine the network from the address and allow you to leave the prefix off of the alias argument.

chasefleming:
Closes #1363 

______

For contributor use:

### DIFF
--- a/internal/dependencymanager/add.go
+++ b/internal/dependencymanager/add.go
@@ -32,6 +32,7 @@ import (
 type addFlagsCollection struct {
 	name            string `default:"" flag:"name" info:"Name of the dependency"`
 	skipDeployments bool   `default:"false" flag:"skip-deployments" info:"Skip adding the dependency to deployments"`
+	skipAlias       bool   `default:"false" flag:"skip-alias" info:"Skip prompting for an alias"`
 }
 
 var addFlags = addFlagsCollection{}
@@ -58,7 +59,7 @@ func add(
 
 	dep := args[0]
 
-	installer, err := NewDependencyInstaller(logger, state, addFlags.skipDeployments)
+	installer, err := NewDependencyInstaller(logger, state, addFlags.skipDeployments, addFlags.skipAlias)
 	if err != nil {
 		logger.Error(fmt.Sprintf("Error: %v", err))
 		return nil, err
--- a/internal/dependencymanager/dependencyinstaller.go
+++ b/internal/dependencymanager/dependencyinstaller.go
@@ -48,10 +48,11 @@ type DependencyInstaller struct {
 	Logger          output.Logger
 	State           *flowkit.State
 	SkipDeployments bool
+	SkipAlias       bool
 }
 
 // NewDependencyInstaller creates a new instance of DependencyInstaller
-func NewDependencyInstaller(logger output.Logger, state *flowkit.State, skipDeployments bool) (*DependencyInstaller, error) {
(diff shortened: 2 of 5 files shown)


internal/dependencymanager/add.go
```diff
@@ -58,7 +59,7 @@ func add(
 
 	dep := args[0]
 
-	installer, err := NewDependencyInstaller(logger, state, addFlags.skipDeployments)
+	installer, err := NewDependencyInstaller(logger, state, addFlags.skipDeployments, addFlags.skipAlias)
```
ianthpun, line 62: minor, but should we just have `NewDependencyInstaller` take `addFlagsCollection` as a whole instead? That way we don't have to interrupt the method parameters each time if we want to add/remove params related to flags

internal/dependencymanager/dependencyinstaller.go
```diff
@@ -275,6 +277,14 @@ func (di *DependencyInstaller) handleFoundContract(networkName, contractAddr, as
 		}
 	}
```
ianthpun, line 279: quick comment here would be good, double negatives can be tricky to read

internal/dependencymanager/dependencyinstaller.go
```diff
@@ -308,6 +318,30 @@ func (di *DependencyInstaller) updateDependencyDeployment(contractName string) e
...
+	label := fmt.Sprintf("Enter an alias address for %s on %s if you have one, otherwise leave blank", contractName, missingNetwork)
+	raw := util.AddressPromptOrEmpty(label, "Invalid alias address")
+
+	if raw != "" {
+		contract, err := di.State.Contracts().ByName(contractName)
+		if err != nil {
+			return err
+		}
+
+		contract.Aliases.Add(missingNetwork, flowsdk.HexToAddress(raw))
```
jribbink, line 339: Does this alias get saved?  It looks like to me we save in `updateDependencyState` & `updateDependencyDeployment`, so I don't think the state gets saved from what I can see after this alias is added.

Also on that note -- should we be saving at every step, or just once at the end.  IMO we should probably not be saving at intermediate steps because there is no guarantee that the user will not abort/error will not be encountered at a later point.  I think the expected behaviour if the command does not successfully complete is probably that the user's `flow.json` configuration did not change

internal/dependencymanager/add.go
chasefleming: Can do that currently @ianthpun . Only time that could become a problem is if the install flags and add flags start to deviate in anyway. Thoughts?

ianthpun: i guess the option in that case would be to have a "dependencyManageConfig" struct, which add/install would fill in. Might be overkill, but something to consider. I'd prefer that over using the existing add/installFlag structs

internal/dependencymanager/dependencyinstaller.go
chasefleming: Added a comment

ianthpun (approved):

ianthpun (approved):

jribbink (approved):

PR #1518, opened 2024-04-15
Update EVM Gateway to latest version & add Previewnet chain ID

jribbink:
## Description

Some cleanup from the release last week where we had to use a commit hash to update the gateway to be compatible with `flow-go` & adds previewnet EVM chain ID.

______

For contributor use:

### DIFF
--- a/internal/evm/gateway.go
+++ b/internal/evm/gateway.go
@@ -56,7 +56,7 @@ type gatewayFlag struct {
 	RPCPort            int    `flag:"rpc-port" default:"3000" info:"port for the RPC API server"`
 	AccessNodeGRPCHost string `flag:"access-node-grpc-host" default:"localhost:3569" info:"host to the flow access node gRPC API"`
 	InitCadenceHeight  uint64 `flag:"init-cadence-height" default:"0" info:"init cadence block height from where the event ingestion will start. WARNING: you should only provide this if there are no existing values in the database, otherwise an error will be thrown"`
-	EVMNetworkID       string `flag:"evm-network-id" default:"testnet" info:"EVM network ID (testnet, mainnet)"`
+	EVMNetworkID       string `flag:"evm-network-id" default:"testnet" info:"EVM network ID (testnet, mainnet, previewnet)"`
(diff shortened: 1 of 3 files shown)


internal/evm/gateway.go
```diff
@@ -56,7 +56,7 @@ type gatewayFlag struct {
 	RPCPort            int    `flag:"rpc-port" default:"3000" info:"port for the RPC API server"`
 	AccessNodeGRPCHost string `flag:"access-node-grpc-host" default:"localhost:3569" info:"host to the flow access node gRPC API"`
 	InitCadenceHeight  uint64 `flag:"init-cadence-height" default:"0" info:"init cadence block height from where the event ingestion will start. WARNING: you should only provide this if there are no existing values in the database, otherwise an error will be thrown"`
-	EVMNetworkID       string `flag:"evm-network-id" default:"testnet" info:"EVM network ID (testnet, mainnet)"`
+	EVMNetworkID       string `flag:"evm-network-id" default:"testnet" info:"EVM network ID (testnet, mainnet, previewnet)"`
```
jribbink, line 59: Is this default still what we want?  I get that previewnet is probably what we want _today_, but in the long run it seems like testnet is probably going to be the suitable default?  Thoughts?

chasefleming (approved):

internal/evm/gateway.go
chasefleming: I'd think testnet long term, but I'm not sure. @sideninja any thoughts?

jribbink: @sideninja

ianthpun (approved):

internal/evm/gateway.go
jribbink: I'm thinking maybe we actually default it to previewnet.  Considering that `emulator` is used as default for everywhere else & this is probably the equivalent that somebody might use if they were on the emulator, I actually think this may make more sense.

Also, see: https://github.com/onflow/flow-go/commit/896c38ebd1e9c15ffc1933893474ee294147bb2c#diff-828e33688886f2aab971192b9c67e40f9b8c93f7a1db406eb2b2b8ce17d41c82R7-L21

sideninja: Let's use previewnet until emulator is not an option.

PR #1529, opened 2024-04-19
Use GitHub reports for `is-validated` command

### LINKED ISSUE #1508: Add command to pull staged contract status
### Issue

Currently, `flow-c1 migrate is-validated` pulls the contract status from the on-chain repository.  However, this is not currently being populated and there are no existing plans to do so, so we need to determine a workaround

### Suggested solution

Add cli command to allow user to pull current status from the published GitHub report on https://github.com/onflow/cadence repository, as this is where the reports are currently being published

jribbink:
Closes #1508 

## Description

Fetches the report from GitHub instead of looking at the on chain data.

Failure result:

Screenshot 2024-04-24 at 2 34 39 PM

Success result:

Screenshot 2024-04-24 at 2 38 16 PM

______

For contributor use:


internal/migrate/is_validated.go
```diff
@@ -42,45 +65,239 @@ var IsValidatedCommand = &command.Command{
...
+		}
+
+		return validationResult{
+			Timestamp: *timestamp,
+			Status:    *status,
+		}, nil
+	}
+}
+
+func getContractValidationStatus(network config.Network, address string, contractName string, state *flowkit.State, repoService GitHubRepositoriesService, logger output.Logger) (*contractUpdateStatus, *time.Time, error) {
```
chasefleming, line 129: Is it necessary to return a pointer here for `contractUpdateStatus`?

internal/migrate/is_validated.go
```diff
@@ -42,45 +65,239 @@ var IsValidatedCommand = &command.Command{
...
 
-func isValidated(
+const (
+	repoOwner = "onflow"
+	repoName  = "cadence"
+	repoPath  = "migrations_data/raw"
+	repoRef   = "master"
+)
+
+func isValidated(repoService GitHubRepositoriesService) func(
```
chasefleming, line 78: Nit, but does is it more go idiomatic to do something like this:

```go
type Validator struct {
    repoService GitHubRepositoriesService
}

func NewValidator(repoService GitHubRepositoriesService) *Validator {
    return &Validator{
        repoService: repoService,
    }
}

func (v *Validator) isValidated() (command.Result, error) {
}
```

internal/migrate/is_validated.go
jribbink: Not necessary.  I can use an empty struct in place of everywhere I used nil, and it just felt more natural when doing the search inside of the for-loop (i.e. nil for object that may not exist), I don't really know what convention is here

chasefleming (approved):

PR #1547, opened 2024-04-26
add contract migration check on all calls

### LINKED ISSUE #1543: Notify devs of no longer validated contracts
After some migrations, depending upon dependencies and migrations to Crescendo, some staged contracts may be no longer valid. We should notify in the terminal during certain commands, like deploy, if there are any no longer valid contracts

This should occur on the master and feature branches.

ianthpun:
Closes #1543 

## Description

confirmed working here with a manual test

image
______

For contributor use:


internal/migrate/is_validated.go
```diff
@@ -111,6 +112,85 @@ func newValidator(repoService GitHubRepositoriesService, network config.Network,
...
 }
 
+func (v *validator) getContractUpdateStatuses(contractNames ...string) ([]contractUpdateStatus, *time.Time, error) {
+	var contractUpdateStatuses []contractUpdateStatus
+	err := checkNetwork(v.network)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	v.logger.StartProgress("Checking if contracts has been validated")
```
chasefleming, line 122: Nit: would probably add an ellipses on this at the end.

internal/command/command.go
```diff
@@ -126,6 +126,8 @@ func (c Command) AddToParent(parent *cobra.Command) {
 			checkVersion(logger)
 		}
 
+		// command to validate contracts are sstill valid for migration
```
chasefleming, line 129: Spelling error

internal/migrate/is_validated.go
```diff
@@ -111,6 +112,85 @@ func newValidator(repoService GitHubRepositoriesService, network config.Network,
...
+			contractUpdateStatuses = append(contractUpdateStatuses, s)
+			foundAddresses = append(foundAddresses, s.AccountAddress)
+		}
+	}
+
+	for addr, contractName := range addressToContractName {
+		var missingContractErr error
+		if !slices.Contains(foundAddresses, addr) {
+			builder := strings.Builder{}
+			builder.WriteString("some contracts do not appear to have been a part of any emulated migrations yet, please ensure that it has been staged & wait for the next emulated migration (last migration report was at ")
```
chasefleming, line 159: Capitalize "Some"

chasefleming (commented):
A couple minor things

cmd/flow/main.go
```diff
@@ -57,34 +53,6 @@ func main() {
 	var cmd = &cobra.Command{
 		Use:              "flow",
 		TraverseChildren: true,
-		// Messaging for Cadence 1.0 upgrade
```
ianthpun, line 60: deleting this since this feature branch is literally 1.0

internal/migrate/get_staged_code.go
```diff
@@ -53,13 +54,13 @@ func getStagedCode(
 	flow flowkit.Services,
 	state *flowkit.State,
 ) (command.Result, error) {
-	err := checkNetwork(flow.Network())
+	err := util.CheckNetwork(flow.Network())
```
ianthpun, line 57: moved `CheckNetwork` and `GetAddressByContractName` to `util` package

internal/migrate/is_validated.go
```diff
@@ -19,49 +19,25 @@
...
 	"github.com/onflow/flowkit/v2/output"
 	"github.com/spf13/cobra"
 
 	"github.com/onflow/flow-cli/internal/command"
+	"github.com/onflow/flow-cli/internal/migrate/validator"
 	"github.com/onflow/flow-cli/internal/util"
 )
 
-//go:generate mockery --name GitHubRepositoriesService --output ./mocks --case underscore
-type GitHubRepositoriesService interface {
```
ianthpun, line 43: moved the whole `validator` bits into `internal/migrate/validator`

internal/command/command.go
```diff
@@ -424,14 +433,40 @@ func UsageMetrics(command *cobra.Command, wg *sync.WaitGroup) {
...
+	Format                 string
+	Save                   string
+	Host                   string
+	HostNetworkKey         string
+	Log                    string
+	Network                string
+	Yes                    bool
+	ConfigPaths            []string
+	SkipVersionCheck       bool
+	ContractMigrationCheck bool
```
ianthpun, line 446: change this to `SkipContractMigrationCheck`

internal/migrate/validator/validator.go
```diff
@@ -0,0 +1,356 @@
...
+		builder.WriteString(contract.ContractName)
+		builder.WriteString("\n - Network: ")
+		builder.WriteString(contract.Network)
+		builder.WriteString("\n\n")
+	}
+
+	return builder.String()
+}
+
+type brokenContractError struct {
```
ianthpun, line 84: delete this i think

internal/command/command.go
```diff
@@ -424,14 +431,40 @@ func UsageMetrics(command *cobra.Command, wg *sync.WaitGroup) {
...
+			logger.Error(fmt.Sprintf(
+				"Contract %s has failed the last emulated migration\n"+
+					" - Account: %s\n"+
+					" - Contract: %s\n"+
+					" - Error: %s\n",
+				contract.ContractName,
+				contract.AccountAddress,
+				contract.ContractName,
+				contract.Error,
+			))
```
jribbink, line 465: Can we print this to stderr (the default logger here is an `StdoutLogger`).  This is diagnostic information intended for the user and not necessarily comand output.  Using stdout could break external tooling that relies on the output from the Flow CLI (e.g. Language Server, VSCode, Flow JS Testing).

jribbink, line 465: Also - imo we should keep this message minimal, e.g. Just list all of the contracts that failed, if any, and link to the migration report.  It takes up to 1 week (and many CLI commands) to resolve these errors, so we should try not to pollute the user's terminal too much as this could become annoying.

chasefleming:
@ianthpun has this been manually tested? The screenshot shows the code error output? That seems large and unnecessary. Can we keep everything but the error output and maybe say something like "run blah command" to see the errors?

chasefleming:
Looks like @jribbink already left a similar comment as I just did so plus one.

internal/command/command.go
```diff
@@ -424,14 +431,40 @@ func UsageMetrics(command *cobra.Command, wg *sync.WaitGroup) {
...
+	contractStatuses, err := validator.NewValidator(github.NewClient(nil).Repositories, flow.Network(), state, logger).GetContractStatuses()
+	if err != nil {
+		logger.Error(fmt.Sprintf("failed to validate contracts: %s", err))
+		return
+	}
+
+	for _, contract := range contractStatuses {
+		if contract.IsFailure() {
+			logger.Error(fmt.Sprintf(
+				"Contract %s has failed the last emulated migration\n"+
```
chasefleming, line 457: I feel like this needs more context to people. Maybe they are running a project and didn't do the staging. They may have no idea what this means. Seems like you'd want to say something along the lines of "You had contracts staged for migrating to Cadence 1.0. Since successfully staged an error occurred and you need to restage/migrate (whatever the right term is) your contracts. To learn more about the migration process, check out..." Not worded perfectly, but you get the idea.

internal/command/command.go
```diff
@@ -424,14 +431,40 @@ func UsageMetrics(command *cobra.Command, wg *sync.WaitGroup) {
...
+					" - Account: %s\n"+
+					" - Contract: %s\n"+
+					" - Error: %s\n",
+				contract.ContractName,
```
chasefleming, line 461: Also, this is a lot to show. Is it better just to say like "3 contracts with errors found since last migration," and then a way to see more?

ianthpun: ya i agree i think it's quite a bit. Maybe we can list the contracts and then just like to the migration data?

ianthpun: updated now

image


I/ll update the original PR screenshot as well

chasefleming: What does it look like if there are multiple that fail?

internal/command/command.go
```diff
@@ -126,6 +128,11 @@ func (c Command) AddToParent(parent *cobra.Command) {
 			checkVersion(logger)
 		}
 
+		// check contract migrations if flag is set
+		if !Flags.SkipContractMigrationCheck {
+			checkContractMigrations(state, logger, flow)
```
jribbink, line 133: I think you may encounter a nil pointer dereference here.

Notice how the error may be swallowed when initializing the flowkit state https://github.com/onflow/flow-cli/pull/1547/files#diff-69d3575d337e6861d23e877e04bbdb99f437a82661f706f055a50e82b1e8282aR111-R113.  If someone runs this comand & doesn't have a flow.json present, I'd hazard to guess that this ends up in a panic state.

internal/command/command.go
```diff
@@ -424,14 +431,48 @@ func UsageMetrics(command *cobra.Command, wg *sync.WaitGroup) {
...
+	SkipContractMigrationCheck bool
+}
+
+
+const migrationDataURL = "https://github.com/onflow/cadence/tree/master/migrations_data"
+
+func checkContractMigrations(state *flowkit.State, logger output.Logger, flow flowkit.Services) {
+	contractStatuses, err := validator.NewValidator(github.NewClient(nil).Repositories, flow.Network(), state, logger).GetContractStatuses()
+	if err != nil {
+		logger.Error(fmt.Sprintf("failed to validate contracts: %s", err))
```
jribbink, line 453: This should just be ignored.  If we can't get the migration report, we shouldn't log anything to the user.  This is something that happens silently in the background unrelated to their actions in the Flow CLI, and only notifies if a failed migration is found, so I think logging this causes more harm than good

internal/command/command.go
ianthpun: image

i'll update the initial PR  to show that as well if anyones curious

internal/command/command.go
ianthpun: pushed changes so that we check to see if `state` is not nilil first. If it is nil, we just return, so no panic

internal/command/command.go
ianthpun: updated, thanks!

chasefleming (approved):

PR #1550, opened 2024-04-29
Default `flow test` behaviour to test `**/*_test.cdc`

### LINKED ISSUE #1378: `flow test` configuration and default behaviour
### Issue To Be Solved

Currently we require to execute a  command line command with many parameters each time we want to run cadence tests with flow cli, e.g. `flow test --cover --covercode="contracts" cadence/test/*`

To increase developer efficiency while working with `flow test` I suggest to be able to run `flow test` with sensible defaults without parameters, and provide with an option to define a test configuration file, where defaults can be overridden. 

The expected developer experience is to by running `flow test` tests are executed with sensible behaviour as long as they are placed in the directory defined by conventions the community agrees on. 
Developers should also be able to configure in a test configuration file any custom location and other configurations if their project has different requirements.

Sensible conventions should be determined for each parameter, e.g. in the above example flow test could execute tests in the cadence/tests/* directory by default.

### (Optional): Context

Testing frameworks in other language ecosystems (e.g. jest) adopted philosophy to make the tool work great by default, but give a way to configure if developers prefer to use it in a different way. This can positively effect developer experience.

jribbink:
Closes #1378 

## Description

Removes mandatory requirement for specifying test file names and defaults to `**/*_test.cdc` (not a breaking change).

______

For contributor use:

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -74,10 +74,10 @@ var testFlags = flagsTests{}
 
 var TestCommand = &command.Command{
 	Cmd: &cobra.Command{
-		Use:     "test <filename>",
+		Use:     "test [filenames]",
 		Short:   "Run Cadence tests",
-		Example: `flow test script.cdc`,
-		Args:    cobra.MinimumNArgs(1),
+		Example: `flow test`,
+		Args:    cobra.ArbitraryArgs,
 		GroupID: "tools",
 	},
 	Flags: &testFlags,
@@ -101,8 +101,22 @@ func run(
 		)
 	}
 
+	var filenames []string
+	if len(args) == 0 {
+		filenames, err := filepath.Glob("**/*_test.cdc")
+		if err != nil {
+			return nil, fmt.Errorf("error loading script files: %w", err)
+		}
+
+		if len(filenames) == 0 {
+			return nil, fmt.Errorf("no test files found matching the pattern *_test.cdc")
+		}
+	} else {
+		filenames = args
+	}
+
 	testFiles := make(map[string][]byte, 0)
-	for _, filename := range args {
+	for _, filename := range filenames {
 		code, err := state.ReadFile(filename)
 		if err != nil {
 			return nil, fmt.Errorf("error loading script file: %w", err)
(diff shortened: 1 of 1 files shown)


internal/test/test.go
```diff
@@ -101,8 +101,22 @@ func run(
 		)
 	}
 
+	var filenames []string
+	if len(args) == 0 {
+		filenames, err := filepath.Glob("**/*_test.cdc")
```
chasefleming, line 106: `_test.cdc` should probably be in a constant

chasefleming (changes requested):

internal/test/test.go
```diff
@@ -101,8 +101,22 @@ func run(
...
+		if err != nil {
+			return nil, fmt.Errorf("error loading script files: %w", err)
+		}
+
+		if len(filenames) == 0 {
+			return nil, fmt.Errorf("no test files found matching the pattern *_test.cdc")
```
chasefleming, line 112: Is this an error? Technically not wrong to find no matching patterns right?

jribbink: Yeah, on the fence about this one.

I see value in an error status because it may be informative to the user (i.e. if somebody misconfigures their CI and no test files run, this prevents the situation where they mistakenly assume that their tests are passing).

The flip side is that like you said, it's not necessarly an error state, strictly speaking... an empty result is still a valid result when performing a search.

Personally, I like the error better because it's fairly harmless but exit code 1 ensures that it won't pass CI.  What do you think?

chasefleming: Well if no test files were found, would it not say something like "0 tests ran" somewhere else? If not, we should add that in.

jribbink: Yeah, I can just swap it over if you think it's better to exit successfully when no tests are found.  Previously, it was impossible to encounter such a state (command required valid filename arguments).

jribbink: Yeah, it already displays that message.  I've just removed the error.

chasefleming (approved):

PR #1555, opened 2024-05-01
Fix Windows tests

### LINKED ISSUE #1556: Windows CI not working as expected
### Problem

The Windows CI is not failing even though there are test failures.  Also, we should update our tests to work on Windows as well.

### Steps to Reproduce

Look at CI logs for Windows.

### Acceptance Criteria

Tests pass,and CI fails if they aren't

### Context

n/a

jribbink:
Closes #1556 

## Description

Fixes our long-standing issue with Windows tests doing nothing :P.  The CGO fixes we made a while back sovled half the issue, this solves the rest.

Example failing CI here (we used to have the issue that failing Windows tests didn't cause the to fail, this shows that it is fixed):
https://github.com/onflow/flow-cli/actions/runs/8915898654/job/24486335666

______

For contributor use:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,8 @@ binary: $(BINARY)
 
 .PHONY: install-tools
 install-tools:
-	cd ${GOPATH}; \
-	mkdir -p ${GOPATH}; \
+	cd '${GOPATH}'; \
+	mkdir -p '${GOPATH}'; \
 	GO111MODULE=on go install github.com/axw/gocov/gocov@latest; \
 	GO111MODULE=on go install github.com/matm/gocov-html/cmd/gocov-html@latest; \
 	GO111MODULE=on go install github.com/sanderhahn/gozip/cmd/gozip@latest; \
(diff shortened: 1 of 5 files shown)


jribbink:
This _works_ & is a significant improvement over what existed before but there's two issues I'm going to investigate (timeboxed, so I may or may not solve them in this PR).

1. `actions/setup-go` is slow, like REALLY slow.  It seemed like there was some semi-recent work done on this front by the maintainers, and I tried updating the action, but it didn't seem to solve the problem.  Maybe I solve it, maybe I don't.
2. There's an issue with the Makefile, seemingly related to Windows backslashed paths.  It's coming from a very old workaround in the Makefile to attempt to create $GOPATH if it doesn't exist, so it doesn't actually cause any issues becuase this workaround is not really necessary.  I'll see if either we can fix this or just ditch the workaround.

Makefile
```diff
@@ -86,10 +86,6 @@ publish:
 clean:
 	rm ./cmd/flow/flow*
 
-.PHONY: install-linter
-install-linter:
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ${GOPATH}/bin v1.47.2
-
 .PHONY: lint
```
jribbink, line 89: We don't use this at all anymore, it's part of `install-tools`

jribbink:
Only open issue is getting a repo secret setup for `CODECOV_TOKEN`, kind of unrelated to this PR, but thought it was worth fixing since it doesn't seem to work currently.  I'm seeing if @bthaile is able to do this for me.

chasefleming (approved):

ianthpun (approved):

PR #1566, opened 2024-05-03
Refactor generate logic, add disable logs, and disable from setup

### LINKED ISSUE #1554: Disable logs in generate for setup
Logs appear inside of the `generate` logic saying where they were created. There should be a way to disable logs in generate since the interactive setup uses its own logging.

chasefleming:
Closes #1554 

______

For contributor use:

### DIFF
--- a/internal/super/generate.go
+++ b/internal/super/generate.go
@@ -96,6 +96,8 @@ func init() {
 	GenerateScriptCommand.AddToParent(GenerateCommand)
 }
 
+const DefaultCadenceDirectory = "cadence"
+
 func generateContract(
 	args []string,
 	_ command.GlobalFlags,
@@ -104,7 +106,7 @@ func generateContract(
 	state *flowkit.State,
 ) (result command.Result, err error) {
 	options := GeneratorOptions{
-		Directory: "cadence",
+		Directory: DefaultCadenceDirectory,
 		State:     state,
 		Logger:    logger,
 	}
@@ -121,7 +123,7 @@ func generateTransaction(
 	state *flowkit.State,
 ) (result command.Result, err error) {
 	options := GeneratorOptions{
-		Directory: "cadence",
+		Directory: DefaultCadenceDirectory,
 		State:     state,
 		Logger:    logger,
 	}
@@ -138,7 +140,7 @@ func generateScript(
 	state *flowkit.State,
 ) (result command.Result, err error) {
 	options := GeneratorOptions{
-		Directory: "cadence",
+		Directory: DefaultCadenceDirectory,
 		State:     state,
 		Logger:    logger,
 	}
@@ -184,7 +186,7 @@ func (g *Generator) generate(templateType, name string) error {
 
 	var fileToWrite string
 	var testFileToWrite string
-	var rootDir = "cadence"
+	var rootDir = DefaultCadenceDirectory
 	var basePath string
 	var testsBasePath = "tests"
 	var err error
--- a/internal/super/generate.go
+++ b/internal/super/generate.go
@@ -22,10 +22,11 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
-	flowsdk "github.com/onflow/flow-go-sdk"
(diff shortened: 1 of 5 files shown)


internal/super/generate.go
```diff
@@ -102,7 +102,18 @@ func generateContract(
 	_ flowkit.Services,
 	state *flowkit.State,
 ) (result command.Result, err error) {
-	return generateNew(args, "contract", generateFlags.Directory, logger, state)
+	if len(args) < 1 {
```
ianthpun, line 105: shouldn't line 86 cover that already?

https://github.com/onflow/flow-cli/pull/1566/files#diff-1a34fb00f194ec131896919bfad4b3fdaf639c76fd91304228c346eb73dc89b9L86

chasefleming: Yes good catch! Removed

internal/super/generate.go
```diff
@@ -122,64 +136,60 @@ func generateScript(
...
-	return strings.TrimSuffix(name, filepath.Ext(name))
+// TemplateMap defines a map of template types to their specific names
+type TemplateMap map[string]string
+
+type GeneratorOptions struct {
+	Directory   string
+	State       *flowkit.State
+	Logger      output.Logger
+	DisableLogs bool
 }
```
jribbink, line 163: Imo this is unnecessary & belongs in the `Generator` struct itself, and these can be set in the factory method.  Defining options like this isn't necessarily any safer/different than:

```
g := &Generator{
	Directory: xxx,
	State: xxx,
	Logger: xxx,
	DisableLogs: xxx,
}
```

And `GeneratorOptions` probably just adds an unnecessary layer of complexity.  The factory method can be used to enforce that a `&Generator` object is created with valid configuration.

e.g. `NewGenerator(directory, state, logger)`, `NewGeneratorWithoutLogs(directory, state)` -- although I _think_ that NOOP loggers is the idiomatic way to handle this.

internal/super/generate.go
```diff
@@ -122,64 +136,60 @@ func generateScript(
...
+	return &Generator{
+		Options: options,
 	}
+}
 
-	var executedTemplate bytes.Buffer
-	// Execute the template with the provided data or nil if no data is needed
-	if err = tmpl.Execute(&executedTemplate, data); err != nil {
-		return "", fmt.Errorf("failed to execute template: %w", err)
+func (g *Generator) Create(typeNames TemplateMap) error {
```
jribbink, line 175: Is a map keyed by type the best data structure for this?  What happens if I want to generate two scripts?

internal/super/generate.go
```diff
@@ -122,64 +136,60 @@ func generateScript(
...
-func stripCDCExtension(name string) string {
-	return strings.TrimSuffix(name, filepath.Ext(name))
+// TemplateMap defines a map of template types to their specific names
+type TemplateMap map[string]string
+
+type GeneratorOptions struct {
+	Directory   string
+	State       *flowkit.State
+	Logger      output.Logger
+	DisableLogs bool
```
jribbink, line 162: Could this be handled by a NOOP/nil logger?

internal/super/generate.go
chasefleming: You could do it but we could use an array instead. Updated.

internal/super/generate.go
chasefleming: Will that this in when we have support

internal/super/generate.go
```diff
@@ -122,157 +133,185 @@ func generateScript(
...
-	}
-	return fmt.Sprintf("%s.cdc", name)
-}
+// TemplateMap defines a map of template types to their specific names
+type TemplateMap map[string][]string
 
-func stripCDCExtension(name string) string {
-	return strings.TrimSuffix(name, filepath.Ext(name))
+type Generator struct {
+	Directory   string
```
ianthpun, line 145: i believe we can make these private now since you got the constructor params workign :)

ianthpun (approved):

jribbink (approved):

PR #1772, opened 2024-09-24
Refactor Generator

### LINKED ISSUE #1774: Generalize & refactor project generator utility
### Issue To Be Solved

Currently the Generator used by supercommands init/generate is strongly coupled to individual template file types (i.e. there is. branching logic for scripts/transactions/contracts with a lot of duplicate logic).

The `TemplateItem` interface should be adjusted so that we can remove this branching logic and make the generator more extensible.  This helps unblock future needs (i.e. generate test files, other project files like README.md, etc.)

jribbink:
Closes #1774 

## Description

Most of the work is just reoganizing logic

 - Removed branching logic based on template type in `Generator.generate` function.  This logic was all moved to the TemplateType structs.  This helps us generalize a bit better without needing special handling for each new type (i.e. README.md, tests, and anything else in the future.
 - Added a "TestTemplate" for generating tests instead of embedding this logic inside of Contract templates

I feel like this could be potentially abstracted further by breaking the template files themselves out of the generator, but probably out of scope for this PR/unnecessary right now and I think what's been changed in this PR is already an improvement.

______

For contributor use:

### DIFF
--- a/internal/super/generator/generator.go
+++ b/internal/super/generator/generator.go
@@ -113,8 +113,7 @@ func (g *Generator) generate(item TemplateItem) error {
 
 	outputContent, err := g.processTemplate(templatePath, fileData)
 	if err != nil {
-		// TODO, better error based on template type
-		return fmt.Errorf("error generating template: %w", err)
+		return fmt.Errorf("error generating %s template: %w", item.GetType(), err)
 	}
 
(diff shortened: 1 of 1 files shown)


chasefleming:
@jribbink You tested this manually with `init`, `generate`, and with the custom directory flag in `generate`?

PR #1424, opened 2024-02-21
Update flix networks

bthaile:
## Description

Update flixkit-go version to support passing in targeted networks that support Cadence v1.0
Fix bug with wrong address format for contract alias.
______

For contributor use:

- [x] Targeted PR against `master` branch
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [ ] Code follows the [standards mentioned here](https://github.com/onflow/flow-cli/blob/master/CONTRIBUTING.md#styleguides)
- [ ] Updated relevant documentation
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Added appropriate labels

PR #1852, opened 2024-12-03
update to use latest version of flixkit-go, pass in core contracts as…

bthaile:
… dependencies

Closes #1424 

## Description

Update to use latest version of flixkit-go that does not have a dependency on flowkit.
Update to include core contracts allow with user defined contracts when generating a flix template.

______

For contributor use:


nvdtf (approved):

internal/super/flix.go
```diff
@@ -57,13 +60,15 @@ type flixResult struct {
...
-	Use:              "flix",
-	Short:            "execute, generate, package",
-	TraverseChildren: true,
-	GroupID:          "tools",
-}
+var (
+	flags   = flixFlags{}
+	FlixCmd = &cobra.Command{
+		Use:              "flix",
+		Short:            "execute, generate, package",
```
chasefleming, line 67: Should be a sentence description

chasefleming (changes requested):

internal/super/flix.go
```diff
@@ -283,6 +294,49 @@ func (fr *flixResult) Oneliner() string {
...
+	coreContracts := make(flixkit.ContractInfos)
+
+	excludeMainnet := slices.Contains(excludeNetworks, config.MainnetNetwork.Name)
+	excludeTestnet := slices.Contains(excludeNetworks, config.TestnetNetwork.Name)
+	excludeEmulator := slices.Contains(excludeNetworks, config.EmulatorNetwork.Name)
+
+	if !excludeMainnet {
+		// Add mainnet contracts
+		sc := systemcontracts.SystemContractsForChain(flowGo.Mainnet)
+		for _, c := range sc.All() {
```
chasefleming, line 307: This should be abstracted to avoid repetition

chasefleming (changes requested):

internal/super/flix.go
```diff
@@ -256,6 +261,13 @@ func generateFlixCmd(
 
 	ctx := context.Background()
 
+	// merge deployed contracts and core contracts
+	for name, addresses := range coreContracts {
+		if _, exists := depContracts[name]; !exists {
```
chasefleming, line 266: What if they are different contracts but the same name? This won't work. You could use the source string as identifiers. There is a function for that somewhere.

chasefleming (changes requested):

internal/super/flix.go
```diff
@@ -57,13 +60,15 @@ type flixResult struct {
...
-var flags = flixFlags{}
-var FlixCmd = &cobra.Command{
-	Use:              "flix",
-	Short:            "execute, generate, package",
-	TraverseChildren: true,
-	GroupID:          "tools",
-}
+var (
+	flags   = flixFlags{}
+	FlixCmd = &cobra.Command{
```
chasefleming, line 65: Can you add examples?

chasefleming (changes requested):

internal/super/flix.go
bthaile: How do string imports work for contracts that import multiple contracts with same name? Also can you point out an example of core contracts that have the same name?

bthaile: Or are you saying there would be an issue if the user has deployed a contract that uses the same name as a core contract? in that case if the user already has a contract name "FungibleToken" then that contract would be used instead of the core contract. Which I think is the correct behavior.

bthaile: Actually since FLIX only deals with transactions and scripts. Is it possible for them to use String imports and import a contract same name and different addresses?

chasefleming: > How do string imports work for contracts that import multiple contracts with same name? Also can you point out an example of core contracts that have the same name?

You can't import two contracts with the same name. They have to be named different in the `flow.json`. Since you're using the name to filter it's possible you could have a name in your `flow.json` matching the name from core contracts, which is why using an identifier like `testnet://0x123.Hi` might be a better ID to filter on. 

> Or are you saying there would be an issue if the user has deployed a contract that uses the same name as a core contract? in that case if the user already has a contract name "FungibleToken" then that contract would be used instead of the core contract. Which I think is the correct behavior.

There is no guarantee that because someone names it `FungibleToken` in their `flow.json` its actually the `FungibleToken` contract. 

Question: Do you even need to pull the core contracts, because wouldn't someone making a FLIX with imports have the core contract listed in their `flow.json`?

> Actually since FLIX only deals with transactions and scripts. Is it possible for them to use String imports and import a contract same name and different addresses?

Can't import two with the same name.

bthaile: ok, I think all is good then. I wanted to import core contracts to make it easier to generate flix. If there is an issue with multiple contracts with same name I'll look into using an identifier. I don't think the case will come up. I rather keep flix simple at this stage.

internal/super/flix.go
```diff
@@ -218,6 +225,10 @@ func generateFlixCmd(
 ) (result command.Result, err error) {
 	cadenceFile := args[0]
 	depContracts := getDeployedContracts(state)
```
chasefleming, line 227: I think you need to be pulling all contracts from state and not just deployed contracts. For example, you could be using a third party contract, have it in your `flow.json`, and not have it deployed. It would then miss grabbing it in this code.

chasefleming (changes requested):

internal/super/flix.go
chasefleming: We don't really have this pattern in the CLI where contracts are handled behind the scenes without any visibility to the developer. It seems like if you want to have the core contract handled for flix, it should also exist in your `flow.json`. What if on running flix it recognized missing core contracts and prompted if you want to add each to your project with a yes/no prompt then called Dependency Manager behind the scenes to automatically add them to a project that required them for flix?

internal/super/flix.go
bthaile: Good call. I added dependencies from flow.json using alias, that looks to be the format that contract dependencies puts in the flow.json. Updated the PR. thanks

internal/super/flix.go
```diff
@@ -328,17 +377,37 @@ func getDeployedContracts(state *flowkit.State) flixkit.ContractInfos {
...
 			}
+			if slices.Contains(excludeNetworks, network) {
+				continue
+			}
 			allContracts[name][network] = addr
 		}
 	}
 
+	// add contracts that have not been deployed
+	for _, c := range *state.Dependencies() {
```
chasefleming, line 388: This should be `.Contracts` not `.Dependencies`. `Contracts` should include `Dependencies`, but not vice versa.

chasefleming (changes requested):

internal/super/flix.go
chasefleming: An error instead of the interactive (although nice) would be simpler and could tell them to install missing dependencies with Dependency Manager so they are in the `flow.json`

internal/super/flix.go
bthaile: flow.json

```
         "contracts": {
		"HelloWorld": "./cadence/contracts/HelloWorld.cdc"
	},
	"dependencies": {
		"Burner": {
			"source": "mainnet://f233dcee88fe0abe.Burner",
			"hash": "71af18e227984cd434a3ad00bb2f3618b76482842bae920ee55662c37c8bf331",
			"aliases": {
				"emulator": "f8d6e0586b0a20c7",
				"mainnet": "f233dcee88fe0abe",
				"testnet": "9a0766d93b6608b7"
			}
		},
...
```

Does .Contracts have addresses? Is Dependencies property exist in both .Contracts and .Dependencies?

chasefleming: Dependencies are ingested and treated as contracts, so if you use `.Contracts` you get contracts and dependencies, but not the other way around.

bthaile:
closing, these changes get picked up on another PR

chasefleming (approved):

PR #1884, opened 2025-02-04
Update to Cadence v1.3.1, Go v1.23

turbolent:
## Description

- Update to Cadence v1.3.1
  - Since https://github.com/onflow/cadence/releases/tag/v1.1.0 the package structure of Cadence changed, so the imports need to be adjusted 
  - Since https://github.com/onflow/cadence/releases/tag/v1.2.0 the `Crypto` contract is no longer part of Cadence, but is now deployed on-chain / imported from the chain.

    Remove support for direct imports in the linter for now. Adding support for this deprecated form of import requires quite a bit of work, which is out-of-scope for this update PR (resolving the Crypto contract from chain, based on given chain).

- While at it, fix the order of `Equal()` assertions (correct order is expected, actual)

______

For contributor use:


turbolent:
@m-Peter @onflow/flow-cadence-execution  Could you please take a look at `internal/evm/gateway.go` and make the necessary adjustments?
- `config.Config`'s `CreateCOAResource` field is gone. Maybe the flag just needs to be removed?
- The arguments to `bootstrap.Run` seems to have changed

PR #2114, opened 2025-10-06
Add alias configuration command

T4Bu:
Adds `flow config add alias` command to add contract aliases to configuration. Supports both interactive and flag-based modes.

### DIFF
--- a/internal/config/add-alias.go
+++ b/internal/config/add-alias.go
@@ -1,0 +1,118 @@
+/*
+ * Flow CLI
+ *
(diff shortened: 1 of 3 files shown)


T4Bu:
Hello @jribbink no problem, and of course.

T4Bu:
I'm assuming it's intentional that addresses in memory have the 0x prefix, while in json they don't? Just mentioning because tests that only check in memory values will pass but will fail when reading explicit json.

PR #2246, opened 2026-01-06
Add pins to dependency height

### LINKED ISSUE #2243: Pin `flow.json` dependencies to block heights for historical contract availability
## Problem

Currently, `flow.json` dependencies are always effectively pinned to the most recent version, enforced by a hash check/on-chain query.

However, when local dependencies drift from those on-chain, there is currently no way for the dependency manager to query these historical versions, so we must return an error if they choose not to update their outdated dependencies.

## Suggested Solution

Introduce a JSON schema similar to the following:

```
"dependencies": {
    "source": "",
    "aliases": {},
    "block_height": "1234",
    "hash": "4567"
}
```

Where instead of returning errors in cases where the user has a valid, but outdated, dependency and chooses not to update (i.e. https://github.com/onflow/flow-cli/blob/master/internal/dependencymanager/dependencyinstaller.go#L687), we should rather install the dependency from the included block height pin (if it exists).

For legacy projects, the first install after updating the CLI should pin the dependendency to the latest block height.  We should also validate that we don't have backward-compatibility issues with older CLI versions, or if these exist identify where these are/what mitigation strategies can leverage.

## Drawbacks

By no longer forcing developers to update dependencies to the latest version, we risk developers letting their local dependencies drift from real, on-chain runtime dependencies too far.  Before making this change we should validate that we have considered all the potential risks in this area.

jribbink:
Closes #2243 

## Description

Adds `block_height` pins to DM dependencies.

- Behavioiur remains backward-compatible, however, depdedencies can now be historically backfilled on new machines/CI without requiring a manual update step.
- Migration/`block_height` pins will be added after the first install after updating. 
- Post-migration `flow.json` is still backward-compatible with old CLI version, and will just ignore the unknown field

______

For contributor use:


internal/dependencymanager/dependencyinstaller.go
```diff
@@ -533,9 +579,61 @@ func (di *DependencyInstaller) fetchDependenciesWithDepth(dependency config.Depe
 		return fmt.Errorf("error adding dependency: %w", err)
 	}
 
-	accountContracts, err := di.getContracts(networkName, address)
+	// Prevent duplicate dependencies: check if this contract is already managed via a different network
+	// Example: Foo stored as mainnet://0xabc, but discovered transitively as testnet://0xdef (aliased)
+	shouldContinue, err := di.checkForCrossNetworkDuplicate(dependency.Name, networkName, address.String())
```
chasefleming: Do we need to be aware of anything in imports with checkForCrossNetworkDuplicate?

PR #2325, opened 2026-04-15
Fall back to project root when file resolver cannot find file

holyfuchs:
The `fileResolver` always resolved file paths relative to the test file that triggered the import. This breaks when a helper script is shared across test files in different directories.

For example, given this layout: 
``` 
tests/some_test.cdc -> imports "test_helpers" -> reads some_contract.cdc -> resolves to tests/some_contract.cdc
tests/mocks/some_test.cdc -> imports "../test_helpers" -> reads some_contract.cdc -> resolves to tests/mocks/some_contract.cdc
``` 

This essentially forces having all test files in a single directory.

The fix adds a fallback: if a file isn't found at the path relative to the test file, it retries the path relative to the project root (the working directory where `flow.json` lives).
The original error is returned only if both lookups fail, so existing behavior is fully preserved. 
Any test that already works continues to work unchanged.

### DIFF
--- a/internal/test/test_test.go
+++ b/internal/test/test_test.go
@@ -255,6 +255,29 @@ func TestExecutingTests(t *testing.T) {
 		assert.NoError(t, result.Results[script.Filename][0].Error)
 	})
 
+	t.Run("with file read from project root when not found relative to test file", func(t *testing.T) {
+		_, state, rw := util.TestMocks(t)
+		_ = rw.WriteFile(
+			tests.SomeFile.Filename,
+			tests.SomeFile.Source,
+			os.ModeTemporary,
+		)
+		t.Parallel()
+
+		// Place the test script in a subdirectory. SomeFile only exists at the
+		// project root, so the first lookup (relative to the script) will fail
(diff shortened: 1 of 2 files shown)


internal/test/test_test.go
```diff
@@ -255,6 +255,29 @@ func TestExecutingTests(t *testing.T) {
...
 	})
 
+	t.Run("with file read from project root when not found relative to test file", func(t *testing.T) {
+		_, state, rw := util.TestMocks(t)
+		_ = rw.WriteFile(
+			tests.SomeFile.Filename,
+			tests.SomeFile.Source,
+			os.ModeTemporary,
+		)
+		t.Parallel()
```
janezpodhostnik, line 265: can you move this to the top of the function?

janezpodhostnik (approved):

internal/test/test_test.go
holyfuchs: Not sure but all of the tests have same setup, I just copied it, would rather leave this consistent.

PR #2328, opened 2026-04-16
Fix unused import detection in Cadence linter

holyfuchs:
## Problem

`flow cadence lint` never reported `unused-import-hint` diagnostics despite the `UnusedImportAnalyzer` being registered and running.

The Cadence sema checker populates `ImportDeclarationResolvedLocations` in its elaboration during type checking. Without a `LocationHandler`, implicit imports (`import "Foo"` or `import "./Foo.cdc"`) get an empty `Identifiers` slice in the resolved location. The `UnusedImportAnalyzer` iterates those identifiers to know what was imported — finding nothing, it never flags anything.

## Fix

Added `resolveLocation` as a `LocationHandler` in the sema checker config. For implicit imports it resolves the imported file and returns its exported declaration names as the resolved identifiers, giving the analyzer the data it needs.

- A `currentLocation` field on the linter tracks the file being checked, saved/restored around sub-checker calls in `handleImport` so nested imports resolve correctly.
- An `exportedIdentifiers` cache avoids re-parsing: `handleImport` pre-populates it from the already-parsed program before sub-checking.

## Known limitation

See: https://github.com/onflow/cadence-tools/pull/628

Imports used *only* in conformances (`resource Foo: Bar.Interface`) or entitlement access specifiers (`access(Bar.Entitlement)`) are false-positively flagged. Root cause: `UnusedImportAnalyzer` tracks usage via `ast.Inspector`, which relies on `Walk` — and `CompositeDeclaration.Walk` does not traverse `Conformances` or `Access`. Fix belongs in `onflow/cadence-tools` (`lint/unused_import_analyzer.go:98`). Tests for correct behavior are included and currently failing.

### DIFF
--- a/internal/cadence/linter.go
+++ b/internal/cadence/linter.go
@@ -42,12 +42,12 @@ import (
 )
 
 type linter struct {
-	checkers                map[string]*sema.Checker
-	exportedIdentifiers     map[string][]ast.Identifier
-	state                   *flowkit.State
-	checkerStandardConfig   *sema.Config
-	checkerScriptConfig     *sema.Config
-	currentLocation         common.Location
+	checkers              map[string]*sema.Checker
+	exportedIdentifiers   map[string][]ast.Identifier
+	state                 *flowkit.State
+	checkerStandardConfig *sema.Config
+	checkerScriptConfig   *sema.Config
+	currentLocation       common.Location
 }
 
 type positionedError interface {
--- a/internal/cadence/linter.go
+++ b/internal/cadence/linter.go
@@ -42,10 +42,12 @@ import (
 )
 
 type linter struct {
-	checkers              map[string]*sema.Checker
-	state                 *flowkit.State
-	checkerStandardConfig *sema.Config
-	checkerScriptConfig   *sema.Config
+	checkers                map[string]*sema.Checker
+	exportedIdentifiers     map[string][]ast.Identifier
+	state                   *flowkit.State
+	checkerStandardConfig   *sema.Config
+	checkerScriptConfig     *sema.Config
+	currentLocation         common.Location
 }
 
 type positionedError interface {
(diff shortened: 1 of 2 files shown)


internal/cadence/linter.go
```diff
@@ -474,6 +486,82 @@ func (l *linter) resolveImportFilepath(
...
+
+func exportedIdentifiersFromProgram(program *ast.Program) []ast.Identifier {
+	var identifiers []ast.Identifier
+	for _, decl := range program.CompositeDeclarations() {
+		identifiers = append(identifiers, decl.Identifier)
+	}
+	for _, decl := range program.InterfaceDeclarations() {
+		identifiers = append(identifiers, decl.Identifier)
+	}
+	return identifiers
```
turbolent, line 566: There might also be other declarations, like global variables, functions, etc.

holyfuchs: This seems to be ok?
Checked it with production repo and couldn't find issues.
```cadence
for _, decl := range program.Declarations() {
        identifier := decl.DeclarationIdentifier()
        if identifier == nil {
                continue
        }
        if decl.DeclarationAccess() != ast.AccessAll {
                continue
        }
        identifiers = append(identifiers, *identifier)
}
```

holyfuchs: importing this from a _helper.cdc:
`access(all) let x = 1`
and using 
`access(all) let y = x`
correctly does not flag as unused import

turbolent: Right, this can just be generalized to the `DeclarationIdentifier()` method defined for all declarations. Access might also be e.g. `access(account)` (one contract from an account imports another from the same), but looks good other than that 👍

Kay-Zee (approved):

PR #2330, opened 2026-04-17
Add //lint:ignore <category> directive to Cadence linter

holyfuchs:
## Description

Adds inline suppression directives to the Cadence linter. Place `//lint:ignore <category>` on the same line or the line immediately before a diagnostic to suppress it. A new `--show-ignored` flag reveals what is being suppressed.

## How it works

```cadence
// Suppress on the line before:
//lint:ignore semantic-error
someUndeclaredVar

// Suppress inline on the same line:
someUndeclaredVar //lint:ignore semantic-error
```

The category must match the diagnostic category exactly (e.g. `semantic-error`, `removal-hint`, `replacement-hint`, `security`).

Ignored diagnostics are counted in the summary but never affect the exit code:

```
$ flow cadence lint contract.cdc
1 problem (0 errors, 0 warnings, 1 ignored)

$ flow cadence lint --show-ignored contract.cdc
contract.cdc:3:2: removal-hint: unnecessary force operator (ignored)
1 problem (0 errors, 0 warnings, 1 ignored)
```

### DIFF
--- a/internal/cadence/lint.go
+++ b/internal/cadence/lint.go
@@ -42,16 +42,19 @@ import (
 type lintFlagsCollection struct {
 	WarningsAsErrors bool   `default:"false" flag:"warnings-as-errors" info:"Treat warnings as errors"`
 	BaseDir          string `default:"" flag:"base-dir" info:"Directory to search for .cdc files (defaults to current directory)"`
+	ShowIgnored      bool   `default:"false" flag:"show-ignored" info:"Show diagnostics suppressed by //lint:ignore <category> directives"`
 }
 
 type fileResult struct {
-	FilePath    string
-	Diagnostics []analysis.Diagnostic
+	FilePath           string
+	Diagnostics        []analysis.Diagnostic
+	IgnoredDiagnostics []analysis.Diagnostic
 }
 
 type lintResult struct {
-	Results  []fileResult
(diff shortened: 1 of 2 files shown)


jribbink:
Great idea @holyfuchs !  Couple questions:

Would it be possible to make this a PR to the https://github.com/onflow/cadence-tools repo instead?  Ideally we would preserve this abstraction layer, and not require the CLI to independently parse the source code.  I.e. the linter is used internally by the Cadence Language Server & Flow MCP Server as well, so ideally this sort of feature lands in all the places it would apply to.

Second, I'm not really sure, but curious what you think - unlike other general purpose languages, the source code of a smart contract language is intended to be readable by users.  Are there any concerns that an inline flag could encourage users to pollute their production contracts/transactions with development artifacts?  I'm assuming that a config file/CLI flags would not solve your use case?

turbolent:
Such a feature was already added to the linter in https://github.com/onflow/cadence-tools/pull/619.

I'm not sure if there's additional work needed in the CLI to make it work, but we shouldn't have two different approaches/directives. 

Let's discuss feature additions before we start on them, e.g. on Slack, so we don't duplicate work.

holyfuchs:
yeah it works
`// lint-disable-next`!

Issue #2334, opened 2026-04-20
Parallel test execution does not work when coverage is enabled


turbolent:
### Problem

#2299 made test execution concurrent.

However, it does not work when running with coverage enabled.

For example, see https://github.com/onflow/flow-core-contracts/actions/runs/24682268286/job/72182352680?pr=605#step:7:425

```sh
$ flow test --cover --covercode="contracts" tests/*_test.cdc

fatal error: concurrent map read and map write

goroutine 46 [running]:
internal/runtime/maps.fatal({0x320a91b?, 0xc00298e3b0?})
	/root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.1.linux-amd64/src/runtime/panic.go:1046 +0x18
github.com/onflow/cadence/runtime.(*CoverageReport).IsLocationExcluded(...)
	/root/go/pkg/mod/github.com/onflow/cadence@v1.10.2/runtime/coverage.go:151
github.com/onflow/cadence/runtime.(*CoverageReport).AddLineHit(0xc000df2ae0, {0x3ba82a0, 0xc00298e3b0}, 0x1ff)
	/root/go/pkg/mod/github.com/onflow/cadence@v1.10.2/runtime/coverage.go:160 +0x45
```

### Steps to Reproduce

- Run tests in parallel, with coverage enabled

### Acceptance Criteria

Coverage can be collected when running tests.

### Additional Info

The coverage report in Cadence is currently not thread-safe. We need to either disable concurrent test execution when coverage is enabled, or add ideally make the coverage report thread-safe.

turbolent:
Opened #2335 to temporarily fix the crasher, and https://github.com/onflow/cadence/pull/4476 in Cadence. Once we've merged the Cadence PR and updated the CLI, we should be able to re-enable concurrent test execution when coverage is enabled.

PR #2335, opened 2026-04-20
Temporarily disable concurrent test execution when collecting coverag…

turbolent:
Work towards #2334 

## Description

Temporarily disable concurrent test execution when collecting coverage, until the coverage report in Cadence is thread-safe 

______

For contributor use:

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -273,8 +273,12 @@ func testCode(
 	}
 
 	// Limit concurrency to flags.Jobs, defaulting to number of CPU cores.
+	// When coverage is enabled, force sequential execution because the
+	// coverage report is shared across all test file goroutines.
 	jobs := flags.Jobs
-	if jobs <= 0 {
+	if flags.Cover {
+		jobs = 1
+	} else if jobs <= 0 {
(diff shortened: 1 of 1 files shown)


internal/test/test.go
```diff
@@ -273,8 +273,12 @@ func testCode(
 	}
 
 	// Limit concurrency to flags.Jobs, defaulting to number of CPU cores.
+	// When coverage is enabled, force sequential execution because the
+	// coverage report is shared across all test file goroutines.
 	jobs := flags.Jobs
-	if jobs <= 0 {
+	if flags.Cover {
```
janezpodhostnik, line 285: Do you think it would be ok to print something here, so that the user is aware?

janezpodhostnik (approved):

internal/test/test.go
turbolent: Good idea! Added in [75b5655](https://github.com/onflow/flow-cli/pull/2335/commits/75b5655fb8b8e0364bb0c01d4c18b5878a246139) and also refactored the existing warning to use the logger instead

